PR #113, opened 2016-08-23
Add autocomplete values for PLoT-ME

mchlltt:
Rearranged and added values extensively.


jthrilly:
Awesome work Mich. Have you got the development environment set up such that you can run the grunt tasks, and create main.min.js?

mchlltt:
Hey Josh, I don't, actually -- I've just been modifying things in Sublime Text. I imagine this would be helpful!

jthrilly:
So does that mean that you are manually editing main.min.js in your pull request? If so, let me know and I'll show you how I automatically generate it...!

mchlltt:
That would be lovely!

PR #137, opened 2017-03-23
Jsdoc demo

meshulam:
**Do not merge**

Demo of generating API docs from JSDoc annotations. This uses the `jsdoc` tool to render a default HTML template into `docs-build/`. Other tools can use the same annotations with minimal modification, but `jsdoc` is among the most widely used.

At this point, the important thing is less the actual rendered output, but to get in the habit of adding annotations, especially to functions and classes that will be user-facing.

Feel free to close this PR, I'll keep this branch around for reference.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /build
 /release-builds
 /www
+/docs-build
 
 npm-debug.log*
 yarn-debug.log*
--- a/jsdoc.conf.json
+++ b/jsdoc.conf.json
@@ -1,0 +1,5 @@
+{
+  "source": {
+    "includePattern": ".+\\.js$"
+  }
+}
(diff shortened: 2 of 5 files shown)


jthrilly:
Good stuff Matt. Now we have the babelrc stuff sorted in master, do you want to do this as a feature branch off of this repo, which we can merge when we are happy with it?

jthrilly:
Also, feel free to integrate this with our npm tasks!

meshulam:
I wound up not relying on the project's .babelrc and just specifying the couple transforms that jsdoc couldn't understand automatically. So there probably won't be any changes to this branch but I'll test it out and let you know.

jthrilly:
From what I saw, this should be mergeable as-is. I suppose the only question is the workflow once the docs have been generated.

PR #143, opened 2017-04-18
Select stages

wwqrd:
Just the functional part.

Doesn't yet load the appropriate Interface component, but does affect other components that rely on this data (like form/prompt) when using compatible stage config.


src/ducks/modules/stage.js
```diff
@@ -1,13 +1,18 @@
+import { actionTypes as protocolActionTypes } from '../../ducks/modules/protocol';
+
 const NEXT_PROMPT = 'NEXT_PROMPT';
 const PREVIOUS_PROMPT = 'PREVIOUS_PROMPT';
+const NEXT_STAGE = 'NEXT_STAGE';
+const PREVIOUS_STAGE = 'PREVIOUS_STAGE';
 
 const initialState = {
   promptIndex: 0,
```
jthrilly: Not sure that we need to track the prompt index, and almost definitely don't want to do it here. If we track it, can it be put in the redux store for the interface as part of its state?

wwqrd: Yeah you're probably right. It's here because it's reset by a couple of actions, but then it only needs to be reset because it's here... Could just store it using setState on the component. That said, it doesn't hurt it being stored in the redux store, and means that the rehydrate step remembers which prompt you are on.

I'm not totally happy with the name of this module either, any thoughts? `particpantProgress`?

wwqrd: Okay, I've tracked down why this is in the state: The interface needs to know which prompt is active when creating nodes so it can append the appropriate attributes.

It's great that you brought this up though, it's made me reconsider how the current stage is propagated though nested components. They probably should get it as a prop from the interface and not direct from the state.

Still think it the module should have a better name though.

wwqrd: Okay, so based on an attempt at refactoring that, we need to track prompt states independently. I'll rename the module to `participantProgress` and add the additional state data, so you can take a look.

jthrilly: I think this would be best as something like `sessionProgress`, or `sessionManager`. This seems to approximate what was in `session.js` in the old code. 

Great catch on why it needs to be where it is. Hopefully we can come up with something graceful. Let me know when you are finalized on the naming, and we can merge.

wwqrd: How about `session` then? `sessionManager`/`sessionProgress` seem a bit tautological?

wwqrd:
Thought of a way to implement your comment RE removing prompt meta from the app state. The result seems less verbose, we should probably factor out the prompt handler for the next interface.

Overall I think it's an improvement, and I think will make the prompt component more flexible.

PR #145, opened 2017-04-24
Feature/name gen panels [wip]

wwqrd:


src/containers/Elements/NodeProvider.js
```diff
@@ -8,61 +8,58 @@ import { actionCreators as networkActions } from '../../ducks/modules/network';
...
-    this.setState({ nodes: nodes });
-  }
+    console.log('updateNode', source, node, updatedNode);
 
-  handleStop = (node, index, event, draggableData) => {
-    // const nodes = this.state.nodes;
-    // nodes[index].position = { x: draggableData.x, y: draggableData.y };
-    // this.setState({ nodes: nodes });
-    alert("you moved it, for now we're ignoring that");
+    switch(source) {
```
jamieshark: why are we using a switch statement here for only one case?

src/containers/Elements/NodeProvider.js
```diff
@@ -72,11 +69,13 @@ class NodeProvider extends Component {
...
         case 'selectable':
-          return nodes.map(this.draggableNode);
-        case 'draggable':
           return nodes.map(this.selectableNode);
+        case 'draggable':
+          return nodes.map(this.draggableNode);
         default:
-          return nodes.map(this.regularNode);
+          return nodes.map((node, index) => {
+            <Node key={ index } {...node } />
```
jamieshark: nit: missing a space here before `...node`

src/containers/Interfaces/NameGenerator.js
```diff
@@ -58,19 +40,34 @@ class NameGeneratorInterface extends Component {
     });
   }
 
-  handleFormSubmit(node, _, form) {
+  handleAddNode = (node, _, form) => {
```
jamieshark: what is the `_` argument?

wwqrd: It's to indicate an unused parameter provided by SemanticForm. How would you like to see this changed?

src/containers/Elements/NodeProvider.js
wwqrd: No need to preface with nit, it's a valid point ;-)

src/containers/Elements/NodeProvider.js
wwqrd: This is going to have two different behaviours, but I agree that this should be removed until that case actually exists.

wwqrd:
@jthrilly @jamieshark What I'd really like feedback on here is whether or not this is a sensible way to handle panels? Should NodeProvider have direct access to redux, or should it be passed from the interface? Does passing a filter down through components make sense? etc.

PR #146, opened 2017-04-24
Feature/test examples [wip]

wwqrd:


src/components/Elements/Pips.js
```diff
@@ -8,17 +8,26 @@ class Pips extends Component {
       currentIndex,
     } = this.props;
 
+    let pips = [];
+
+    for(let index = 0; index < count; index++) {
```
jamieshark: nit: missing a space here

src/components/__tests__/Elements/Modal-test.js
```diff
@@ -0,0 +1,32 @@
...
+        <Modal show={ true }>
+          <span>foo</span>
+        </Modal>
+      );
+
+      expect(component.contains(<span>foo</span>)).toBe(true);
+    });
+
+  });
+});
```
jamieshark: Perhaps also adding a test for `simulate('click')` and making sure that changes the show/hide?

src/components/Elements/Pips.js
wwqrd: Shall we add this to the lint settings? Wondering if we should extend from stricter preset too like airbnb?

PR #149, opened 2017-04-28
Feature/droppable [wip]

wwqrd:
Roughly working demo of drag and drop

- tracks a drop zone (node list), with `<Droppable accepts={ ['node'] } />`
- triggers an event on `<Draggable>`, which can be used to capture drops
- currently console.logs any draggables dropped into drop zones.

next steps:
- test on devices
- make drop zone more accurate?

enhancements:
- could specify types for draggable and droppable, e.g.
  ```jsx
  <Droppable accepts={ ['node'] } />
  <Draggable type='node' />
  ```
- updates zones on window.resize, suggest also just running a slow setInterval for edge cases

### DIFF
--- a/src/components/Elements/InteractiveNodeList.js
+++ b/src/components/Elements/InteractiveNodeList.js
@@ -1,15 +1,38 @@
 import React, { Component } from 'react';
+import { bindActionCreators } from 'redux';
+import { connect } from 'react-redux';
 import Draggable from 'react-draggable';
 import Touch from 'react-hammerjs';
-import _ from 'lodash';
+import _, { filter } from 'lodash';
 
 import { Node } from '../../components/Elements';
 
+import { actionCreators as droppableActions } from '../../ducks/modules/droppable';
+
 class InteractiveNodeList extends Component {
+
+  dropped = (event) => {
+    const wrap = document.getElementById('page-wrap');
+    const y = event.y + wrap.scrollTop;
+    const x = event.x;
+    this.props.updateZone({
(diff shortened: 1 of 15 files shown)


wwqrd:
Agree on the accuracy part. It's a non trivial thing to get right so I think it's important to verify that this approach works on devices first. Would you be able to do that?

jthrilly:
I've fixed the touch responsiveness issues on iOS which should make testing easier. 

From the look of things, the accuracy is still an issue. That said, I'm not sure _exactly_ what behavior I'm supposed to see. Specifically, where should the green rectangle appear when the node is dropped?

wwqrd:
Just to be clear, I stopped working on the accuracy when I had concerns that the drag functionality wasn't going to work. So it may not be accurate yet!

But essentially it's going by the top left corner (which includes some invisible padding on the node - see web inspector), and positions the green square by top left corner too.

I think tracking by the object center would be a big perceptual improvement, so I'll add that now.

Scrollable divs are probably going to make things more complicated so it'll be worth experimenting with that soon.

wwqrd:
Node center worked a treat.

Making the draggable areas scrollable outright breaks draggable: draggable repositions using offsets, which become hidden as soon as you introduce scrolling. I think we're going to need to take another approach (hammer + el.cloneNode() perhaps).

jthrilly:
That's great. thanks.

Yes, you're going to need to clone the node and do this based on window offset. It is a royal pain.

PR #154, opened 2017-05-04
Feature/droppable clean

wwqrd:
Roughly working demo of drag and drop

- tracks a drop zone (node list), with `<Droppable />`
- triggers an event on `<Draggable>`, which can be used to capture drops
- currently console.logs any draggables dropped into drop zones.

enhancements:
- could specify types for draggable and droppable, e.g.
  ```jsx
  <Droppable accepts={ ['node'] } />
  <Draggable type='node' />
  ```
- updates zones on window.resize, suggest also just running a slow setInterval for edge cases

### DIFF
--- a/src/components/Elements/DraggableNodeList.js
+++ b/src/components/Elements/DraggableNodeList.js
@@ -1,19 +1,19 @@
 import React, { Component } from 'react';
-import Draggable from 'react-draggable';
+import Draggable from '../../containers/Elements/Draggable';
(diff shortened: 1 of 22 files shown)


wwqrd:
Okay this is ready to go now, it's basically `feature/droppable`, but squished and rebased, and then some updates to selectors and network utils to get the correct behaviour in the panels (and with *much* clearer naming I think?).

PR #160, opened 2017-05-08
Feature/name gen styles

### LINKED ISSUE #159: Style the Name Generator interface
Functionality is complete.

Add styles to match designs so we can consider this interface "complete".

wwqrd:
This is a work in progress but I'd appreciate a few bits of feedback/direction!

- There are some SVGs in the project, I'm assuming this is just an asset dump, and these need to go into public to be used?
- @jamieshark How does this fit into what you're doing with the styles project? Do I need to port these styles over as mixins once they are done? Is the style project linked up with this one? Where should assets live?
- @jthrilly I've got the two reference designs (4:3, and 16:9), I'm assuming we care about % widths more or less, with those two as fixed reference points?

Thanks,
Steve!

resolves #159


wwqrd:
Thanks for the feedback!

> The node dragging seems to be messed up. The node 'grows' to the full node size when it is picked up from the side panel, but it is offset to the top left away from the drag point (the cursor). Probably not an issue to be fixed in this branch.

Hmm, I think this is a result of all the new absolute positionings. I think hooking into the absolute "mouse" position, rather than what it's using now would make this problem vanish throughout.

> There's no form styling, although I assume this is because it is so complicated and probably needs to be done by @jamieshark in the base styles.

Not because it's too complicated, I simply hadn't got that far with it. I'm a fan of smaller PRs though so it might make sense to do that separately.

> We need to think about how we handle scrollbars, because they are hideous. The old app just hid them all, which is the approach I still want to take, but only if we are using a touch capable device. Is this possible?

These are just the built in scrollbars at present. I wouldn't recommend trying to implement your own scroll functionality with javascript, because you're essentially re-inventing the wheel. Probably more so with touch. As a general principle it's a good idea to use familiar widgets.

> On that note, I also had a kind of 'fade out' effect at the top and bottom of every panel (including the main node list). It gave a visual clue that there was more data than could be shown in the container. Could we do something similar? I can explain implementation if you like.

Interesting, perhaps we could implement a kind of 'scrollable' component to show this state. I take the implementation to only show the top fade when the window is scrolled already?

> There are a couple of functionality issues I've just noticed. (1) Dragging a node into the interview, and then changing the prompt or interface doesn't cause those nodes to show in the "people you already mentioned" panel. (2) I realized that I missed off the delete functionality entirely from the spec. Basically, while any node is being dragged, a bin icon becomes visible on the bottom left of the screen. Dropping the node on this icon deletes it. This functionality applies to the main node panel, too.

1) Should do so for changing the stage?? That it doesn't for the prompt was intentional - I understand you to be saying now that a node can belong to multiple prompts in the same stage?

2) Believe this functionality should happen in a separate PR, but can roll in adding this to the dom and styles in this one.

wwqrd:
Implemented a scrollable component, makes the markup much cleaner too. Needs to be updated to take account of local colours, I'll think about the best way to do that.

jthrilly:
> There's no form styling, although I assume this is because it is so complicated and probably needs to be done by @jamieshark in the base styles.
> > Not because it's too complicated, I simply hadn't got that far with it. I'm a fan of smaller PRs though so it might make sense to do that separately.
> 

Okay, so you want to do a separate PR for that? It is quite a complicated styling job, what with all the form components and the two modes (take a look at the differences between 4x3 and 196x9 for the new node form in the mockups).

> We need to think about how we handle scrollbars, because they are hideous. The old app just hid them all, which is the approach I still want to take, but only if we are using a touch capable device. Is this possible?
> > These are just the built in scrollbars at present. I wouldn't recommend trying to implement your own scroll functionality with javascript, because you're essentially re-inventing the wheel. Probably more so with touch. As a general principle it's a good idea to use familiar (i.e. built-in) widgets.

I've not made myself clear enough here. I'm not talking about the scrolling functionality - native functionality is fine on both desktop and mobile. I would never want to try and implement scrolling behavior in javascript. I'm talking about the **visibility of the actual scrollbar UI**. Like here:

screen shot 2017-05-09 at 5 20 44 am

It can be hidden, and it can be styled with CSS. Right now, the scrollbars are stock OS X look, which ruins the app UI. I was saying that I'd like to hide them completely on touch enabled devices (since we won't have a discoverability issue) and style them as a fallback.

> On that note, I also had a kind of 'fade out' effect at the top and bottom of every panel (including the main node list). It gave a visual clue that there was more data than could be shown in the container. Could we do something similar? I can explain implementation if you like.
> > Interesting, perhaps we could implement a kind of 'scrollable' component to show this state. I understand the implementation to be that we only show the top fade when the window is scrolled already?

So my solution was less sophisticated. Essentially I used a background gradient for the header and footer of the panels, and a negative margin to make the part in between underlap. I used `::before` and `::after` to avoid more markup. The background gradient blended the background color (or panel color in the case of the panels) at full opacity, with the same at zero capacity. You can probably think of a much smarter way of doing it. Here's how it looked:

screen shot 2017-05-09 at 5 09 36 am

screen shot 2017-05-09 at 5 12 17 am

I see you've implemented this on the bottom, but not the top. And also the side panel version blends with the overall background color, and not the panel bg color as it should. Just a couple of small changes there.

> There are a couple of functionality issues I've just noticed. (1) Dragging a node into the interview, and then changing the prompt or interface doesn't cause those nodes to show in the "people you already mentioned" panel. (2) I realised that I missed off the delete functionality entirely from the spec. Basically, while any node is being dragged, a bin icon becomes visible on the bottom left of the screen. Dropping the node on this icon deletes it. This functionality applies to the main node panel, too.
> > Should do so for changing the stage?? That it doesn't for the prompt was intentional - I understand you to be saying now that a node can belong to multiple prompts in the same stage?

If you recall our conversation about how the panel/node-list logic works, the "matching" that determines if a node is in the main panel is: the node was created on _this_ prompt, and on _this_ interface.

The logic for the "People you already mentioned" panel is the opposite of this: nodes that were _not_ created on this prompt, or on _this_ interface.

So when switching prompts on the same interface, nodes elicited in the previous prompt should show in "People you already mentioned". 

Hope that is clearer.

> Believe this functionality should happen in a separate PR, but can roll in adding this to the dom and styles in this one.

Whatever you think works best :)

wwqrd:
Hey will do a proper response later, but the scrollable component totally does have a top fade, it only activates when it detects scrolling. And I know about the colour thing, it's what I meant by local colours.

wwqrd:
I think I'll need to call you about the side panel.

wwqrd:
Okay, so I've updated the scrollable component to use -webkit-mask, rather than the before, after elements I used before. The advantage is that we don't need to concern ourselves with the background colour of the surrounding component, this disadvantage is that the top is *always* faded, rather than just when the div is scrolled.

I've also hidden the scrollbars.

PR #161, opened 2017-05-10
Feature/node bin

### LINKED ISSUE #121: Implement name generator interface

wwqrd:
This adds a new droppable region for deleting nodes.

As part of introducing this feature I've changed some of the previous components into `behaviours` which can decorate components (i.e. HOCs).

I'm still uncertain about the directory organisation, but it's starting to make more sense now that we have the behaviours extracted into their own place. Otherwise I'm thinking that maybe we keep the `Interfaces` and maybe drop `Elements` until it's obvious we need it?

The draggables and droppables behaviors have a level of complexity which makes me think that they should have something like in `ducks` to store all the different `draggableType`s?

Feedback and suggestions welcome.

Final feature needed to resolve #121


wwqrd:
Yes it's supposed to work 😕 I'll look into it.

> The node bin shouldn't show when you drag nodes from the previous visit panel - these nodes are in an immutable state.

It's hard to see right now if you can't delete nodes, but the bin won't actually delete immutable nodes. But I agree that would be better!

wwqrd:
@jthrilly Okay, draggable is fixed (another touchevent learning curve). The bin only shows up now for a specific draggableType, and also only once the draggable preview starts to show up.

src/behaviors/draggable.js
```diff
@@ -0,0 +1,142 @@
...
+      this.setState({
+        start: {
+          x: draggableData.x,
+          y: draggableData.y,
+        }
+      });
+    }
+
+    onDrag = (event, draggableData) => {
+      if (!this.state.preview) {  // TODO: better way to keep track of drag start?
```
rebeccamadsen: Is this a TODO we're going to address in this merge? If not, should it become an issue in github to address later?

src/components/Elements/NodeList.js
```diff
@@ -1,26 +1,37 @@
...
-        <div className='node-list'>
-          { nodes.map((node, index) => {
-            const label = `${node.nickname}`;
-            return <Node key={ index } label={ label } />;
-          }) }
-        </div>
-      </Droppable>
+      <div className='node-list'>
+        { network.nodes.map((node, index) => {
+          return <EnhancedNode key={ index } label={ label(node) } isActive={ isActive(node) } activeNodeAttributes={ activeNodeAttributes } onSelected={ () => handleSelectNode(node) } onDropped={ (hits) => handleDropNode(hits, node) } draggableType={ draggableType } { ...node } />;
```
rebeccamadsen: Do we have length of line being linted? This is a little hard to read all in one line.

src/containers/Elements/NodeProvider.js
```diff
@@ -29,13 +36,14 @@ class NodeProvider extends Component {
...
 
+    const label = (node) => `${node.nickname}`;
+    const isActive = (node) => isMatch(node, activePromptAttributes);
+
     switch (interaction) {
       case 'selectable':
-        return <SelectableNodeList network={ network } activeNodeAttributes={ activePromptAttributes } handleSelectNode={ this.handleSelectNode } />;
-      case 'draggable':
-        return <DraggableNodeList network={ network } activeNodeAttributes={ activePromptAttributes } handleDropNode={ this.handleDropNode } />;
+        return <NodeList network={ network } label={ label } draggableType='EXISTING_NODE' handleDropNode={ this.handleDropNode } handleSelectNode={ this.handleSelectNode } isActive={ isActive } />;
```
rebeccamadsen: this line is also kind of long

src/styles/components/_name-generator.scss
```diff
@@ -39,11 +39,24 @@
   &__add-person {
     @include svg-button('/images/add-a-person.svg');
     position: absolute;
-    bottom: 20px;
-    right: 0;
-    width: 150px;
+    bottom: 35px;
```
rebeccamadsen: Do some of these numbers make sense to extract as a variable? e.g. 35px seems to be used multiple times; if those usages are connected a variable would make sense.
As a side note, are we linting the scss code?

src/components/Elements/DropZone.js
```diff
@@ -0,0 +1,13 @@
+import React, { Component } from 'react';
+
+import droppable from '../../behaviors/droppable';
+
+class DropZone extends Component {
```
rebeccamadsen: Are we linting for react? If so, are we linting to prefer pure functions over Components? This would be a candidate for function with no state and no methods other than render.

src/components/Elements/NodeList.js
```diff
@@ -1,26 +1,37 @@
 import React, { Component } from 'react';
 import { Node } from '../Elements';
-import Droppable from '../../containers/Elements/Droppable';
+import { scrollable, droppable, draggable, selectable } from '../../behaviors';
+
+const EnhancedNode = draggable(selectable(Node));
 
 class NodeList extends Component {
```
rebeccamadsen: another candidate for function vs Component?

wwqrd:
Thanks @rebeccamadsen, these all seem like reasonable suggestions. I'll have a look at linting options to enforce this kind of thing!

jthrilly:
@rebeccamadsen's suggestion is actually a good one. In the old app drop zones could react when a valid/invalid droppable was hovering over them and react accordingly. Do we have anything like this in your droppable component?

wwqrd:
When you drag nodes from a previous visit the bin doesn't show, this was the requested functionality added in #b63bc94.

That said, we should probably add some on hover animations all around.

jthrilly:
Yes, I _know_ I requested that functionality (ha!) - I was asking about if the droppable can react to valid or invalid hovering draggables. This is useful for the main node list, too.

src/styles/components/_name-generator.scss
wwqrd: Good point, this kind of thing will probably end up in the style guide, which is going to also serve as a library of styles to use in the app.

wwqrd:
Agreed, but not in this PR?

jthrilly:
Agreed.

src/behaviors/draggable.js
wwqrd: I think it would have been better to tackle this before the merge, but there are fixes in this PR that stop the build from breaking, so I hurriedly merged it. It's up there in my todos ;-)

Issue #162, opened 2017-05-11
Enforce some stricter linting (JS/SASS)


wwqrd:
For javascript we're already using eslint, but without many rules, perhaps we could start from the airbnb ones, and also add the react rules: https://github.com/airbnb/javascript/tree/master/react

For sass we could use sass-lint. I'm not sure what rules we want from this except to enforce hyphenatedbem

jthrilly:
Great suggestion, and happy to go ahead with it. I've heard that Airbnb is extremely prescriptive, and somewhat controversial for some of its decisions. We might want to soften it back from that. 

I'm not sure how to handle this since there's probably some existing code that will break, but you could do it as a fix branch against master that people can pull in to any active PRs and manage themselves, maybe? Or we could just put it straight into master and deal with the pain when we open PRs.

jthrilly:
I see @wwqrd went ahead and did a PR and mentioned the Airbnb linting. Steve: if you think that's the best way to deal, let's go for it.

wwqrd:
Pain is the right word for it, this breaks everything!

PR #163, opened 2017-05-11
Feature/linting

wwqrd:
This introduces some linting for scss, and bases our eslinting of the much stricter airbnb.

Airbnb is *seriously* strict. Any that people object to that we should soften?


rebeccamadsen:
Are there any specific rules you are disagreeing with? I know some of the linting rules are preferences, and just helping us maintain similar looking code. Those kinds of rules I'm okay with shifting.

wwqrd:
We're in agreement, I'm just running it by people 😉 

I think maybe the best thing to do would be to set all the errors to warnings and we can tackle them as we come across them?

wwqrd:
Wow, that was a lot of errors 😅 
`eslint ./src/ | grep error | awk '{ print $NF }' | sort | uniq`

jthrilly:
If you are working today, sorting these warnings/errors out seems like a nice "Friday task" to wind down for the weekend. I realize it won't take you all day or anything, but I think you've put a lot of work in the last few days and I want to make sure you don't get too knackered!

wwqrd:
Okay then. It passes. It still works.

Some caveats. For files I know are in the midst of change `Menu`, `Form`, I just disabled the linter for those specific files with a comment. When you merge them, please make sure they lint!

In a couple of places I disabled specific rules for those files, for example `index.js`, where we export lots of different things without a default. If anyone can thing of a fix that will still pass the lint (that is still sane), let me know.

I disabled a few rules globally for now, that we could re-introduce:

- no-array-index-key: we have to use a key on maps, and somethings don't have id's
- forbid-prop-types: the linting forces you to explicitly define proptypes: done. Some aren't allowed?: Kind of tedious, especially since we pass around objects a lot in this app. Perhaps we could slowly trim down our props to be more specific to what we need at a later date?
- jsx-filename-extension: instantly broke some things which looked more like webpack things than code things, decided that this wasn't a great use of time. Maybe we revisit this later.

rebeccamadsen:
I'll check this over first thing Monday. I'm impressed how quickly you got things linting!

rebeccamadsen:
I'm okay with turning off forbid-prop-types and jsx-filename-extension, although we could try:

`"react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }]`

instead of just turning it off. But I do have questions about the [no-array-index-key](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md). Specifically that react uses that key to know if things need to be re-rendered. Which arrays do not have an id; we would need to be confident they won't be reordered to continue using the index as the key.

When I run the linter locally I get the following warning:

> The react/jsx-space-before-closing rule is deprecated. Please use the react/jsx-tag-spacing rule with the "beforeSelfClosing" option instead.

Is that a versioning thing? Because it doesn't look like a rule we have changed in this branch; I'm surprised the library hasn't been updated if this is deprecated.

Also, I am surprised this is the first time my windows environment has come up with a difference, but I get a ton of these:

`1:42   error  Expected linebreaks to be 'LF' but found 'CRLF'  linebreak-style`

So that's fun. I currently have git configured to change line endings to windows style on checkout, and back to unix on commit (core.autocrlf true). I can change that if we want to keep this rule.

wwqrd:
Resolves #162

PR #164, opened 2017-05-11
Feature/style forms and modals

wwqrd:
Have a look see, I've parked this for now based on what seems to be going on in the interactive style guide. We can pick up some more styles later. For now, should we merge this?

Some of the main font sizes will bork, thats because I've taken font sizes out of the style guide and made them global, so probably some fonts need respecifying throughout.


jthrilly:
Depending on which way round you are working with regards to the linting PR, you can merge this in.

PR #165, opened 2017-05-16
Feature/menu stages

### LINKED ISSUE #151: Menu System
# Menu System Spec

In the app we have a menu system. It serves two purposes:

- Navigation between the stages of the interview (the "Stages" menu)
- Access to settings and actions (closing the app, downloading data, etc.) (the "Settings" menu)

The settings menu will be defined within the network canvas app at a later date. This spec will focus on the stages menu.

## Purpose

The stages menu provides a navigation option to quickly move to a specific point in the interview.

## Functionality

- The menu lists the stages of the interview.
- The current stage shows in a selected state.
- Menu items are searchable.
- The list of menu items is in a scrollable container. The menu header (including the filter box) remains fixed to the top of the screen.
- The stages of the interview are defined by the interview protocol. At present, this is located in `src/utils/ProtocolService.js`, and thereafter available in the redux store.
- Clicking an item closes the menu and navigates to that point in the interview.

## Screenshots/Mockups
08network_canvas_stages_16x9
*Mockup of how this will look ultimately*

screen shot 2017-05-03 at 5 43 54 pm
*Example of existing implementation*

screen shot 2017-05-03 at 5 52 15 pm
*Closeup of filter box*

rebeccamadsen:
This fixes #151, adding the Stages menu. 

I added `StageMenu.js` to the containers, since it handles the state. I also refactored `Menu `to extend a `Component `and added a `MenuItem `component. One of the biggest changes here is in using the stage id instead of an index number to set stages; this is in large part because ordering of the array of stages changes when using the "filter" textbox in the menu, so index becomes arbitrary. As a (small) bonus, the address in the url is more user friendly by using the id; of course when running as an app you wouldn't really notice that.

Formatting is not feature complete, as it was my understanding we are waiting for icons and such. I'm happy to adjust this as needed. You can stub adding an image to the menu item if you either add a file in `public/images` with the name of `namegenerator.svg` or change the name of a stage's type in the `ProtocolService `to match one of the svg's in that directory already.

Note that there are at least two more tests I need to add -- `Menu `and `StageMenu `-- and I will update here as soon as I get those in. Also this still needs the linting branch merged in so I can fix whatever won't lint. I just wanted to give you guys a chance to look over the code in the meantime, knowing you will likely be online before me tomorrow. If there is someone else who wants to review, they are welcome, of course.

I may add tests for `App`, `Protocol`, `Stage `and `routes.js `if there is time; no tests are pre-existing for them, but I changed those files in this branch a little bit so it would be great to take time to add tests.


src/containers/Stage.js
```diff
@@ -1,71 +1,75 @@
...
 }
 
-function mapStateToProps(state) {
-  // TODO: http://redux.js.org/docs/recipes/ComputingDerivedData.html
-  const stage = state.protocol.protocolConfig.stages[state.session.stage.index];
+const rotateIndex = (max, next) => {
+  return (next + max) % max;
+}
+
+function mapStateToProps(state, ownProps) {
```
wwqrd: Lets make use of selectors for this, and can we use full words?:

```js
return {
  previousLink: previousLink(state),
  ...,
};

File: src/containers/StageMenu.js [hunk 2]
```diff
@@ -0,0 +1,77 @@
...
+          id: stage.id,
+          title: stage.title,
+          imageType: stage.type,
+          onClick: () => this.props.onStageClick(this.props.stages, stage.id)
+        };
+      }
+    );
+
+    const search = (
+      <div className='bm-search'>
```
wwqrd: I think this might be a remnent of existing code, but can we migrate this to use hyphenated bem styles, with a file per component, see some of the other components for examples, but basically:
 
```scss
// general format:

.component__sub-component--variants {
   ...
}
```

Which can be written in scss as:

```scss
// in menu.scss
.menu {
  ...
  &__menu-item {
     ...
     &--active {
       ...
     }
  }
}

```

src/selectors/session.js
```diff
@@ -16,6 +16,11 @@ const prompt = createSelector(
   (promptIndex, stage) =>  stage.params.prompts[promptIndex]
 )
 
+export const getStages = createSelector(
```
wwqrd: can we use the same naming as the rest of the methods here: i.e `stages`, and refactor `stage` to use this method.

src/styles/components/_menu.scss
```diff
@@ -28,6 +28,7 @@
   background: map-get($roles, menu);
   height:100%;
   color: $charcoal;
+  overflow-y: auto;
```
wwqrd: Might be able to use the HOC 'scrollable' on Menu?

rebeccamadsen:
Thanks for the feedback, Steve! These are good comments. 

It sounds like the big overlying issue is the router vs. state issue. I agree that the router complicates things. My choice to use the router was based on instructions from Josh that the menu should just have routes to all individual stages and that the address history needs to change when you click an address item, so I'd love to hear his opinion on this.

I think react-router-redux looks interesting and would solve the issue of duplicating router vs. state information. I guess the down side is another dependency, but it may be preferable if we need to maintain the router.

src/containers/Protocol.js
```diff
@@ -12,22 +12,23 @@ class Protocol extends Component {
 
   componentWillMount() {
     if (!this.props.protocol.protocolLoaded) {
-      this.props.loadProtocol()
+      this.props.loadProtocol(this.props.stageId);
```
rebeccamadsen: This is how you get the info from the router initially. But if we move away from using the router this is not necessary either.

src/containers/Protocol.js
```diff
@@ -12,22 +12,23 @@ class Protocol extends Component {
...
     );
   }
 }
 
-function mapStateToProps(state) {
+function mapStateToProps(state, ownProps) {
   return {
-    protocol: state.protocol
+    protocol: state.protocol,
+    stageId: ownProps.params.id
```
rebeccamadsen: This is how you get the info from the router initially. But if we move away from using the router this is not necessary either.

src/containers/Stage.js
```diff
@@ -1,71 +1,75 @@
...
+            to={'/protocol/' + this.props.prevLink}
+            className="stage__control-button stage__control-button--back"
+            onClick={() => this.props.onStageClick(this.props.stages, this.props.prevLink)}
+          >
+            Back
+          </Link>
         </div>
         <div className="stage__interface">
-          <CurrentInterface config={ stage }/>
+          <CurrentInterface config={ this.props.stage }/>
```
rebeccamadsen: Maybe this should be a new PR then? It affects NameGenerator which is not yet altered in this branch.

src/components/Menu.js
```diff
@@ -1,22 +1,23 @@
...
-    const newState = { isOpen: !this.state.isOpen };
+    this.state = {
+      isOpen: initialIsOpenProp ? this.props.isOpen : false
+    };
 
-    console.log(newState);
-
-    this.setState(newState);
-  },
+    this.toggleMenu = this.toggleMenu.bind(this);
```
rebeccamadsen: Nice! Doing this.

src/containers/Stage.js
rebeccamadsen: Agreed.

src/containers/StageMenu.js
rebeccamadsen: Yes to the remnant, but I can update it here.

src/selectors/session.js
rebeccamadsen: Will do.

src/styles/components/_menu.scss
rebeccamadsen: I considered that when I saw scrollable, but the specs don't indicate fading. If we decide we want fading, I love the idea of reuse.

rebeccamadsen:
I updated the code based on some of these comments. However, I did not yet pull back from using the router. That is a little more extensive than what I have time for today, but I am happy to do that tomorrow.

src/styles/components/_menu.scss
jthrilly: Lets do this, if you have time. We need a visual affordance to indicate something is scrollable, and I would like it to be consistent.

.eslintrc
```diff
@@ -22,5 +22,6 @@
       "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
       "react/forbid-prop-types": "off",
       "react/no-array-index-key": "off",
+      "linebreak-style": "off",
```
wwqrd: Could you use this .gitattributes specific rule instead?:

`*.js text eol=lf`

(source: http://eslint.org/docs/rules/linebreak-style)

src/behaviors/scrollable.js
```diff
@@ -13,11 +13,15 @@ export default function scrollable(WrappedComponent) {
     }
 
     componentDidMount() {
-      this.node.addEventListener('scroll', this.updateScrollState);
+      if (this.node) {
```
wwqrd: Doesn't componentDidMount only run once? If `this.node` isn't being found then we have some bigger problems!

That said, this scroll tracking feature isn't required anymore, so perhaps we ought to remove it altogether?

src/components/Menu.js
```diff
@@ -1,22 +1,42 @@
...
+  constructor(props) {
+    super(props);
+    const initialIsOpenProp = props && typeof props.isOpen !== 'undefined';
+
+    this.state = {
+      isOpen: initialIsOpenProp ? this.props.isOpen : false
+    };
+  }
 
   // Sets or unsets styles on DOM elements outside the menu component.
```
wwqrd: If this is the required functionality then the open closed state should probably live in app, and be passed down to Menu as a prop. Or alternatively moved into redux.

wwqrd: ...I see this was the previous functionality, perhaps we should refactor it in this PR?

.eslintrc
rebeccamadsen: Oh yes, thanks for the reminder. I did try it once, briefly, but I think it didn't work before because I didn't commit it? Or maybe there is something else I need to do to make git really refresh the line endings on my end. I guess worst case, I commit and then delete my local copy of the repo.

src/behaviors/scrollable.js
rebeccamadsen: You know, I don't run into a problem running the code. But once I added "scrollable" to Menu, I get an error with the Menu test: 
```
● Menu component › renders menu with list

  TypeError: Cannot read property 'addEventListener' of null

    at Scrollable.componentDidMount (src\behaviors\scrollable.js:33:16)
```

If this is not really needed, I can take it out. But I wasn't familiar enough with what it is doing to make that call.

src/components/Menu.js
rebeccamadsen: Yes, this was previous functionality. What are you wanting to refactor? The component's use of state to track whether it is open? Or that an optional prop feeds the state's initial value?

jthrilly: This was linked up to redux before, but I took it out while we were developing. The menu state should go back in redux, as Steve says.

.eslintrc
rebeccamadsen: Got this to work. There was[ a little git magic ](https://help.github.com/articles/dealing-with-line-endings/)to make my local version update.

src/components/Menu.js
rebeccamadsen: Okay, I can put menu open state back in redux. I assume that negates the need for initialIsOpenProp.

wwqrd: I imagine it'll also allow Menu to become a much simpler component perhaps even a stateless function?

src/behaviors/scrollable.js
wwqrd: Originally this component used ::before ::after classes that were animated in depending on the scrolling state from the dom. Now the fade is a constant, so we don't need to keep track of the scrolling state.

rebeccamadsen: So, how much simpler are we talking then? I can get rid of the updateScrollState entirely? Or also the isScrolling state/class?

src/components/Elements/MenuItem.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+/**
+  * Renders a menu item using router links. Image is optional.
+  */
+function MenuItem(props) {
```
jamieshark: nit: we can pull out the various properties here for better readability in the jsx syntax. i.e.
`const {onClick, imageType, title} = this.props;`

and then you can just use `onClick` `imageType` and `title` in the jsx.

src/components/Menu.js
```diff
@@ -1,113 +1,113 @@
...
+import { scrollable } from '../behaviors';
+import { MenuItem } from './Elements';
+
+/**
+  * Renders the internal content of a menu
+  * @param props {object}
+  * props: searchField, items, toggleMenu
+  * @return div
+  */
+function MenuInternal(props) {
```
jamieshark: `MenuInternal` props aren't defined anywhere.

src/containers/App.js
```diff
@@ -1,10 +1,18 @@
...
-import { Link } from 'react-router';
-import { Menu } from '../components';
+import { connect } from 'react-redux';
+
+import { StageMenu } from '.';
 
 require('../styles/main.scss');
 
+
+const menuIsOpen = state => state.menu.menuIsOpen;
```
jamieshark: do we need this declaration? seems like the same thing is being brought in via props and appState down below.

src/containers/StageMenu.js
```diff
@@ -0,0 +1,106 @@
...
+  * @extends Component
+  */
+class StageMenu extends Component {
+  /**
+    * adds search term and list of matching stages to local state
+    */
+  constructor(props) {
+    super(props);
+
+    this.state = {
```
jamieshark: I think that this internal state should go into a reducer?

wwqrd: If it was me I'd put `matchingStages` into a selector too.

src/containers/App.js
wwqrd: This follows the selector pattern, maybe it should live in `selectors` one day, but this looks fine to me.

src/containers/Stage.js
```diff
@@ -4,75 +4,74 @@ import { bindActionCreators } from 'redux';
...
-    stage,
+    currentStage,
   };
 }
 
 function mapDispatchToProps(dispatch) {
   return {
     next: bindActionCreators(stageActions.next, dispatch),
     previous: bindActionCreators(stageActions.previous, dispatch),
+    onStageClick: bindActionCreators(stageActions.setStage, dispatch),
```
wwqrd: @rebeccamadsen @jamieshark I'm happy to use `on` instead of `handle`, but we should decide which and be consistent.

src/components/Menu.js
wwqrd: `MenuInternal` should probably live in `components/`

wwqrd: and then perhaps exported with the scrollable HOC?

src/behaviors/scrollable.js
```diff
@@ -1,42 +1,14 @@
-import React, { Component } from 'react';
-import classNames from 'classnames';
-import { throttle } from 'lodash';
+import React from 'react';
 
 export default function scrollable(WrappedComponent) {
```
wwqrd: 👍

src/containers/Stage.js
wwqrd: I'm kinda leaning towards `on`?

rebeccamadsen: I can use either, just tend to prefer "on" because it parallels onclick and is shorter to type. But we searching src for " on" versus "handle" is not as helpful as one might like!

src/components/Elements/MenuItem.js
rebeccamadsen: Done.

src/containers/StageMenu.js
rebeccamadsen: I had thought of the search term as something more temporary, but I can move it to redux if that is a better place for it.

PR #167, opened 2017-05-24
Load protocol from external file

wwqrd:
- Working custom interface
- Load protocol from github repo
- Setup screen

Main event is: `src/utils/runProtocol.js`, everything else is more or less supporting code


public/advanced.protocol
```diff
@@ -0,0 +1,216 @@
...
+  return {
+    addNode: bindActionCreators(api.actions.network.addNode, dispatch),
+    closeModal: bindActionCreators(api.actions.modal.closeModal, dispatch),
+    openModal: bindActionCreators(api.actions.modal.openModal, dispatch),
+  };
+}
+
+MyAmazingInterface = environment.connect(mapStateToProps, mapDispatchToProps)(MyAmazingInterface);
+
+console.log('I was run 5');
```
rebeccamadsen: Do we need this console.log statement?

public/advanced.protocol
```diff
@@ -0,0 +1,216 @@
...
+            name: "bluefruit",
+          }
+        ]
+      }
+    },
+    "stages": [
+      {
+        "id": "namegen1",
+        "type": {
+          "custom": MyAmazingInterface,
```
rebeccamadsen: There's a warning in the console because type is an object instead of a string:
``Warning: Failed prop type: Invalid prop `imageType` of type `object` supplied to `MenuItem`, expected `string`.
    in MenuItem (at Menu.js:44)``
MenuItem was using type to try and figure out which image to show, but this will need to be changed if type can be a string or an object.

src/components/Elements/Form.js
```diff
@@ -4,6 +4,10 @@ import React, { Component } from 'react';
 import { reduxForm } from 'redux-form';
 import TextInput from '../../components/Form/TextInput';
 
+/**
+  * Renders a redux form that contains fields according to a `fields` config.
+  * @extends Component
+  */
 class Form extends Component {
```
rebeccamadsen: This may not apply to your branch, but is Form simple enough to be a function instead of extending Component?

src/components/Elements/Modal.js
```diff
@@ -2,6 +2,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+/**
+  * Renders a modal window.
+  * @extends Component
```
rebeccamadsen: This doesn't actually extend component as the comment suggests.

src/components/Form/index.js
```diff
@@ -1,2 +1,7 @@
+<<<<<<< HEAD
 /* eslint-disable */
+=======
+/* eslint-disable import/prefer-default-export */
+
+>>>>>>> 22575c1d4... Load basic config over http using rx
```
rebeccamadsen: Unresolved merge conflict?

src/utils/runProtocol.js
```diff
@@ -0,0 +1,51 @@
...
+  connect,
+};
+
+const api = {
+  actions,
+  selectors,
+  elements,
+};
+
+// TODO: What if protocol isn't a valid protocol?
```
rebeccamadsen: New issue in github instead of a TODO, so it can be tracked/commented on?

public/advanced.protocol
jthrilly: Can the interface specify it's icon? That way a custom interface can specify a custom icon. It seems simpler.

rebeccamadsen: Or have an icon that is specifically for custom interfaces. Either way. Or something else entirely.

src/components/Elements/Form.js
wwqrd: I can do that here

public/advanced.protocol
wwqrd: I thought the style guide had a specific icon for a custom interface in it.

Probably the most flexibility we want to give the user is to choose from one of the presets?

wwqrd: @rebeccamadsen What do you think the protocol should look like for this? We could make it so that type was always an object, i.e: `type: { preset: 'name-generator' }`, or we could keep this short hand. Either way, I'm happy to implement the fix to menu.

src/utils/runProtocol.js
wwqrd: Agreed!

public/advanced.protocol
rebeccamadsen: Either way, just so long as we are consistent!

wwqrd: I've kept it as is, and implemented the fix, did a bit of refactoring of styles too.

Issue #152, opened 2017-05-03
App auto updating


jthrilly:
# Electron app update process for Alpha

## Purpose
We need to be able to push updates to our test sites, and ultimately our users, seamlessly and without effort. Initially we will tackle this issue for the electron app, and then move to address this process within android/cordova later.

## Functionality
- The app should track the state of the GitHub repository.
- Either using tags, a special branch, or something else, the app should be aware when a new version is available.
- It should be able to download, extract, perform any post- tasks, and then restart the app (if needed).
- This is taking place in the electron environment, so node is available. The specifics of how this works will need to be researched. If there are multiple solutions, let's talk about them!

racpa:
## Electron
- As I have limited experience with Electron the immediate results bring up [Electron's autoUpdater](https://github.com/electron/electron/blob/master/docs/api/auto-updater.md) and the likely solution being [Nuts](https://github.com/GitbookIO/nuts)
## Cordova
- [CodePush](https://github.com/Microsoft/code-push) is what I have utilised for react-native/cordova.

jthrilly:
Thanks @racpa.

I've taken a quick look, and it seems you are correct that these are the best pre-made options. Do you have any thoughts on if it would make sense to build our own functionality instead? What are the pros and cons?

One of the things I dislike about Squirrel and Nuts is that they seem to require a release server. Yes, they can be minimal stateless apps (like, something we could host for free somewhere) but it still adds maintainability issues to our project. We need to remember that we are only funded for another 4 years, and the project infrastructure must be robust enough to keep working after that.

With that said, we have access to a decent VPS. If we need a release server, we can certainly do it.

Either way, next steps seem to be to explore this functionality in a feature branch. What do you need from me to make that happen? i.e would you like VPS access to explore that, or are you happy experimenting with heroku?

jthrilly:
Also interested in the opinions of @codaco/network-canvas-team more broadly on this.

racpa:
## Release-Server
- Heroku could be used, but on the free tier it will eventually have downtime.  Will the VPS eventually be unavailable in the future? 

## Alternatives
The problem that may eventually arise with a custom solution are the differences installing onto different platforms.  Luckily we could more than likely bootstrap other open-source alternatives for a custom-fit to track github-releases.  

### New Builder
[...]

PR #171, opened 2017-05-25
Feature/app auto updater

racpa:
Issue #152 

I added electron-builder.  And changed a few of the build-scripts.

The shared production folder is now /app which can be run by `npm run build-app`

Travis/AppVeyor send artifacts to be published by GH.  Tests currently fail for the test script in my branch.  Which I think prevents appveyor from continuing on in the build-process.

The variables are set in Travis with CSC_LINK being the certificate needed for Mac. CSC_LINK_PASSWORD is blank.  The GH_TOKEN is the token provided by the bot account.

AppAutoUpdater is provided by electron-updater which syncs with electron-builder and based on version numbers.  You can also use [tags](https://github.com/electron-userland/electron-builder/wiki/Publishing-Artifacts#cli-flags)  

Side-note, sorry haven't squashed any commits.


jthrilly:
Great to see this!

Only issue I see is with switching to `/app` from `/build`. Does anyone forsee any problems related to this? I know build was just an artifact from create-react-app, so can we do any better and just compile into `/www`, since that is the only directory we cannot change?

@racpa - now the tests are passing, can you confirm that the appveyor build is working? Please link up with me on slack about the windows signing certificate, as I now have it ready to go.

public/electron-starter.js
```diff
@@ -7,6 +7,30 @@ const BrowserWindow = electron.BrowserWindow;
...
+//   checkUpdateOnStart: true,
+//   autoDownload: true
+// });
+const log = require('electron-log');
+const {autoUpdater} = require("electron-updater");
+
+autoUpdater.logger = log;
+autoUpdater.logger.transports.file.level = 'info';
+log.info('App starting...');
+console.log("LOG");
```
rebeccamadsen: Do we need a console.log here or was this for debugging?

public/index.html
```diff
@@ -8,7 +8,7 @@
     <title>Network Canvas</title>
   </head>
   <body>
-    <script type="text/javascript" src="cordova.js"></script>
+    <!-- <script type="text/javascript" src="cordova.js"></script> -->
```
rebeccamadsen: Is this script not needed then? Can this line be deleted instead of commented?

jthrilly:
Rebecca: See the Travis build log (I've asked about appveyor) to see it building. Also, see the releases on GitHub to see a draft release with accompanying binaries (OS X only until appveyor is fixed). 

As for seeing if the update works, its hard to test, but this works off the released version in the master branch and compares it with the package.json version of the local version of the app. See the docs for electron-builder for more.

public/index.html
jthrilly: This is added automatically by the Cordova build process. It can be removed from outside of `/www`.

public/electron-starter.js
racpa: console.log isn't necessary

racpa:
@jthrilly 

electron-builder needs to use `/build` to provide the icons, but there is a possibility to jump into the docs to possibly redirect but there were issues regarding this, so I decided to just change the path in `scripts.js` to `/app`

If you wanted to just make the production build go towards `/www` only, that would work as well

wwqrd:
Can we speed up the build at all?

I'm guessing Travis does some caching, so perhaps follow up builds will be quicker? Could we test that?

Some other thoughts on speeding it up:

- Remove any large files / commits from the branch (like release-builds, maybe a squash would fix this?)
- Parallelise the build process (Though I'm not sure this is the bottleneck):

```YAML
env:
  - TARGET=app
  - TARGET=mac
  - TARGET=linux

script: "npm run build-$TARGET"
```

Incidentally, do we need linux? And I'm assuming 'app' is Windows?

jthrilly:
@wwqrd, Linux is a supported platform, yes. Or are you asking if we should build differently for Linux?

Ronald, any response to these optimizations? We definitely need release-builds and node_modules outside of the git tree.

wwqrd:
@jthrilly RE Linux, I didn't know that, awesome! 🤵

racpa:
It's a two package.json structure so the node modules being installed are the bare minimum to get it to work in production. (updater etc.)

I am not sure how to handle the bottleneck, as it could be due to the environment downloading the files to be able to build the packages.

It's possible to give yarn a try for the builds.

Parallel targeting is a great idea.

wwqrd:
Does the fact that there are two package.json mean that `node_modules` and `release-builds` need to be committed? I don't follow?

I'm pretty sure the npm packages are cached by Travis, I don't know about the git checkout? Yarn is certainly faster though, if we aren't using that already.

racpa:
Ahh... I was mainly talking about the Travis build.  

The checkout will probably be gone with a squash, as the branch isn't supposed to be tracking it anymore...?

jthrilly:
Okay, I really want to get this merged in ASAP. @racpa, would it be easier to arrange a meeting to discus these outstanding issues? I am with Steve in not being clear about why any node modules need to be part of the git tree, and I think if release-builds is a result of the build process it should be ignored too so that it is silently deleted after each build completes. 

If you address these issues and squash, I think we can merge.

PR #176, opened 2017-06-08
Fix/menu ui repo

jamieshark:
Dependent on https://github.com/codaco/Network-Canvas-UI/pull/4 being merged in.

Before merging this PR, merge in the NetCanUI PR and then change the dependency in `package.json` to point to head.

### DIFF
--- a/src/styles/components/_input.scss
+++ b/src/styles/components/_input.scss
@@ -5,7 +5,7 @@ $component-name: 'input';
   margin: 20px 0 0;
 
   &__field {
-    @include type('title-2');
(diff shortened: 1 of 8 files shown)


src/styles/components/_input.scss
```diff
@@ -5,7 +5,7 @@ $component-name: 'input';
   margin: 20px 0 0;
 
   &__field {
-    @include text(h2);
+    @include type('title-2');
```
wwqrd: type as in type or type as in typography? If the latter can we remove the abbreviation?

PR #177, opened 2017-06-19
Feature/session menu

### LINKED ISSUE #170: "Session" (Settings) Menu
# Purpose
A menu that provides several in-app utility functions, such as resetting the data, downloading the data, and quitting the app.

# Implementation
- Another instance of the menu component, but with a different icon (see style guide)
- The existing implementation was that a reference to a function was passed to the menu constructor as it was initialized, and then executed on click with a short delay to allow the menu animation to finish. I'm not sure how this would work in react. I'm afraid I can't offer any advice on this. 
- Example functions I would want to implement to start off would be:
  - Quitting the app (on electron, consult the docs on how this is done. for Cordova, I think this just returns false unless there is something equivalent).
  - Resetting the app data. Just clearing the redux store.
  - (bonus if possible) Downloading the data. This will eventually require its own logic as we will want to support various types of export functionality. For now,

rebeccamadsen:
Resolves #170. This adds a second menu for "settings" or "session" items. So far, Quit and Reset session are the only two items.

This also makes use of the ui for icons in the menu.

Note there are some console bugs coming from the ui Icons that Jamie and I are still ironing out. 

Also, I have updated a protocol to explicitly specify the icon name; I am sure there are other protocol files I need to update in parallel -- Steve might want to weigh in on that?

This branch also closes the menu when you click outside it.

I'll check on tests until I hear feedback.


src/styles/main.scss
```diff
@@ -3,12 +3,28 @@
 
 #root {
     @import "containers/home";
+    background: $light-sea-green;
+}
+
+#outer-container {
```
jthrilly: BEM yes. Somewhere else? Maybe. These are menu specific styles, but the elements are higher level. It was somewhat tricky, so I bottled the decision and just put the menu specific styles here, and anything not menu specific in the relevant higher level stylesheet.

wwqrd:
@jthrilly advanced.protocol shouldn't be in there but having a built in protocol speeds up development and allows offline development. In fact this is what I've been doing myself.

My feeling is that we should use the repo as the canonical representation of the protocol, and if we want to use local copies in our branch development we can manage that ourselves (using symlinks for example).

jthrilly:
> @jthrilly advanced.protocol shouldn't be in there but having a built in protocol speeds up development and allows offline development. In fact this is what I've been doing myself.
> My feeling is that we should use the repo as the canonical representation of the protocol, and if we want to use local copies in our branch development we can manage that ourselves (using symlinks for example).

Okay, I take your point. What I would actually like to do is to have the "default" (read: fallback) protocol embedded within the app. This can be what we work on for testing purposes. If you think this is that protocol, can we please rename it to default, fallback, demonstration, or something else sensible, and keep this committed and updated with whatever feature we are working on for development?

rebeccamadsen:
> This looks a little funky to me, with big text buttons, and no icons.
This sounds like a previous iteration; perhaps there's a refresh issue in your view of the branch or your browser?

> Could you implement more of the style guide look for the menu? 
Yes, I will check the style guide. The double menu icon was on my todo this morning anyhow; Jamie added the icon last night for me.

> protocols had been moved outside of the app
I thought that as well, but it seems like the app is defaulting to the one here in /public, which makes sense per Steve's comments. There is also the ProtocolService.js file still in /src/utils, which makes at least three locations. I am happy to update things in the proper place, but it seems redundant to maintain three (or more?) files we need to update when we have changes. Are there any other locations, or just these three?

src/styles/main.scss
rebeccamadsen: I didn't actually realize this file was update in this PR. I'll give it a look over myself. ;)

wwqrd:
@rebeccamadsen ProtocolService isn't used anywhere, you can delete it.

wwqrd:
> This sounds like a previous iteration; perhaps there's a refresh issue in your view of the branch or your browser?

Yes that fixed it for me more or less, thanks!

public/example.protocol.js
```diff
@@ -78,7 +78,8 @@ return {
       {
         "id": "namegen1",
         "type": "NameGenerator",
-        "title": "Name Generator Title",
+        "image": "menu-ng",
```
wwqrd: I think the spec requires us to be able to specify it, so this seems an okay place to put it. Perhaps we should call it `icon` though?

src/styles/main.scss
wwqrd: As it stands I don't think this will even pass the linting (regardless of being BEM or not, because we're styling by ID).

If this is a top level bunch of styles, my preference would be to create a new BEM component for it called something like: "app", which would exist in it's own file.

jthrilly: There is no reason for `#outer-container` over anything else. We just need to ensure we have the perspective property set on the parent of the menu and the page-wrapper. Feel free to refactor however you see fit @rebeccamadsen.

public/example.protocol.js
rebeccamadsen: `icon` works for me!

src/styles/main.scss
rebeccamadsen: Alright, I'll take care of refactoring and double check the linting!

src/components/Menu.js
```diff
@@ -53,37 +82,41 @@ class MenuFactory extends Component {
...
-          visible={isOpen}
-          onClose={toggleMenu}
-        />
-        {!isOpen && <div className="menu__burger">
-          <button onClick={toggleMenu} className="ui large button">
-            <i className="download icon" />
-            Burger
-          </button>
+      <div className="menu" ref={(node) => { this.domNode = node; }}>
+        <div className={isOpen ? 'menu__wrap menu__wrap--isOpen' : 'menu__wrap'}>
```
jamieshark: Nit: I would prefer a class like `menu__wrap--open` rather than the camelcased.

src/styles/containers/_app.scss
```diff
@@ -0,0 +1,30 @@
+// container for menus and content
+.app {
+  height: 100vh;
+  perspective: 1500px;
+  width: 100vw;
+
+  // page wrap for content
+  &__content {
+    background: $navy-taupe;
```
jamieshark: I want to try and deprecate these color vars in favor of a map set reference. so instead of `background: $navy-taupe;` let's use `background: palette('background')` instead.

src/styles/main.scss
```diff
@@ -1,21 +1,18 @@
-// sass-lint:disable clearn-import-paths
+// sass-lint:disable clean-import-paths
 @import '~network-canvas-ui/src/styles/all.scss';
+@import 'containers/_app.scss';
 @import 'components/_all.scss';
 
 // sass-lint:disable no-ids
 #root {
   @import 'containers/home';
+  background: $light-sea-green;
```
jamieshark: use `palette('primary')` or `color('sea-green')`

src/styles/components/_menu.scss
```diff
@@ -1,10 +1,31 @@
...
+  &__wrap {
+    // General sidebar styles
+    position: fixed;
+    left: 0;
+    z-index: 2;
+    transform: none;
+    transition: all $animation-standard-duration $animation-default-easing;
+    height:100%;
+    transform: translate3d(-100%, 0, 0);
+    &--isOpen {
```
jamieshark: I would prefer `--open` rather than `--isOpen`

src/styles/components/_menu.scss
```diff
@@ -1,10 +1,31 @@
-$interface-types: (
```
jamieshark: This file has a lot of color things going on. let's see if we can change all of them to either use the `color` or `palette` mixin.

src/styles/components/_menu.scss
```diff
@@ -62,30 +83,30 @@ $interface-types: (
...
-    font-size: 8;
+    top: 20px;
+    left: 20px;
     z-index: 1;
+    animation-name: menu-button-move;
+    animation-delay: 0s;
+    animation-duration: 1.5s;
+  }
+
+  ~ .menu {
```
jamieshark: I'm not sure if this is doing anything.

src/styles/components/_menu.scss
```diff
@@ -62,30 +83,30 @@ $interface-types: (
...
-    @each $interface-type in $interface-types {
-      &--#{$interface-type} {
-        background-image: url('../images/menu/#{$interface-type}.svg');
-
-        &.menu__menuitem--active,
-        &:hover {
-          background-image: url('../images/menu/#{$interface-type}--active.svg');
-        }
-      }
+    svg {
```
jamieshark: I would maybe prefer the class selector `.icon` here.

src/styles/components/_menu.scss
jthrilly: We will want to use palette, and we will want to define a role color for the menu. @wwqrd can you explain how to do this since I think you set it up?

wwqrd: @rebeccamadsen There are two helper functions in UI, `color()/color-set()` and `palette()/palette-set()`, color defines named colors, and palette attaches those colors to named properties. e.g.

```scss
// in UI
@include color-set('green', #00FF00);
@include palette-set('node', color('green'));
```
```scss
// in Network Canvas
.node {
  color: palette('node');  // #00FF00
}
```

Ideally in Network Canvas we should use the `palette()` method, and define them with `@include palette-set()` in UI.

You can see a list of the presets here, please any that you need:
https://github.com/codaco/Network-Canvas-UI/blob/master/src/styles/global/_default.scss#L29,L47

rebeccamadsen:
I think I've addressed everything, except some of the tests are still failing at getIcons. Any other feedback?

src/styles/components/_menu.scss
```diff
@@ -1,91 +1,147 @@
...
-    padding: 15px 25px 15px 80px;
+    padding: 15px 25px 15px 20px;
     text-decoration: none;
     text-transform: uppercase;
-    background-size: 35px;
-    background-repeat: no-repeat;
-    background-position: 25px 50%;
-    cursor: pointer;
+
+    &--settings:not(:hover):not(.menu__menuitem--active) .icon {
```
jamieshark: Do we need this extra selectors? If the "non-active" state is this colour, can we not just have `&--settings .icon`?

rebeccamadsen: Basically we are toggling white(/gray?) with a color of some sort. We want white-ish when the item is being hovered, or when it is active (selected). We want color when it is not hovered AND not selected. So I think it is either three rules: one to set the color by default, and two to change back to white-ish on hover and also on active selection, or leave white-ish the default and have one rule with the extra selectors.

The --settings class is also used to dictate the background-color when an item is hovered, so it is there whether an item is actively selected or hovered or neither.

jamieshark: Ah, I see. Okay. Sounds good then!

PR #180, opened 2017-06-23
Feature/animation squashed

### LINKED ISSUE #175: Animation
Right now, everything is static. Animation is _vital_ to our user experience. Elsewhere (https://github.com/codaco/Network-Canvas-UI/blob/merging-styles-branch/styles/global/core/_animation.scss) we specify the building blocks of our animation system. The following will need to be implemented both in our in-app CSS, and our base styles repo (for example, component state animation [button highlighting, etc.] is probably universal enough to go into the UI repo, whereas 'between stage transitions' will be implemented in-app using the `_animation.scss` parameters).

The following is a general list of items that need animation.

- Screens should transition as the interview advances forwards or backwards. I am open to suggestions about how this should look, but the existing approach was to fade to black with a loader, and then fade back in.
- Node state changes should animate. For example, selecting/deselecting, and changes in size when dragging should transition nicely.

### Sub-tasks
- [x] Fix node color (use new palette())
- [x] Make stage fade to rich-black
- [x] Modal animation
- [x] Node state animations (selected toggle)
- [x] Animate away empty panels, if possible?
- [x] Stagger node animate in, if possible?
- [ ] Fix animations for menu

wwqrd:
Summary:
- [x] Make stage fade to rich-black
- [x] Modal animation
- [x] Node state animations (selected toggle)
- [x] Animate away empty panels
- [x] Stagger node animate in

Closes #175


wwqrd:
Some of the tests are failing on UI's getIcon, planning to look into fixing that today, the rest should be fine :ok_hand:

wwqrd:
Okay, I think I'm done on my tweaks.

Please note, for the placeholder node on drag hover to display properly you must have the latest version of UI installed (the node will then appear semi-opaque).

src/behaviours/draggable.js
```diff
@@ -1,11 +1,15 @@
+/* eslint-disable */
```
wwqrd: I need to remove this

package.json
```diff
@@ -32,6 +32,7 @@
     "babel-eslint": "7.1.1",
     "babel-jest": "18.0.0",
     "babel-loader": "6.2.10",
+    "babel-plugin-add-module-exports": "^0.2.1",
```
wwqrd: TODO: check what this is for

src/behaviours/draggable.js
```diff
@@ -26,6 +30,16 @@ function moveDistance(start, draggableData) {
   return Math.sqrt((draggableData.x - start.x) ** 2, (draggableData.y - start.y) ** 2);
 }
 
+function* uidGenerator() {
```
wwqrd: Move to utils?

wwqrd:
@jthrilly Just to note our convo on slack: "People from your previous visit list" Shows all nodes initially without animation before removing nodes attached to main view. To repeat:

1. Clear storage
2. Drag node into main list
3. Refresh page

src/styles/components/_interface.scss
```diff
@@ -0,0 +1,10 @@
+.interface {
+  // position: absolute;
```
rebeccamadsen: If this is all commented, do we not need it?

wwqrd:
**Previous node bug**

Nodes in the previous panel are ALL shown, before being abruptly updated with the correct nodes.

To reproduce:
1. Clear application cache
2. Drag a node from the previous nodes panel into the main panel
3. Refresh the app

---

Okay so half of this bug is due to the fact that we are loading the protocol automatically - i.e. the nodes change before the initial animation has finished (which it waits for before updating the stage), and the other half is that it isn't animated at all. The first half isn't easily resolvable and will disappear in real world usage. The second I'm working on.

wwqrd:
Okay, that's all implemented! I'm gonna merge!

wwqrd:
One thing to mention, my fix for the staggered previous node list is a bit of a hack. Basically the list is rendered twice with the same data (which I haven't gotten to the bottom of). My fix here is to stagger the animation for the case that the view is rendered with exactly the same nodes.

The other thing is that, this fix might not look implemented because we are auto loading the protocol *before* the app has a chance to rehydrate - this confusing rendering all round, and since the transitions here need to wait to be rendered before the changes can be displayed causes the strange behaviour on load. It can be fixed by not auto loading the protocol (in the real world app, any auto loading of the protocol would be handled by rehydrate anyway).

Issue #192, opened 2017-08-04
Local interview protocol doesn't load in built version of the app (6)


jthrilly:
The built version of the app looks for the local interview protocol at `http://localhost:3000/`, which obviously doesn't work outside of our dev server setup. 

We need a way of loading the local protocol across all our platforms. I have temporarily pointed the app to our remote protocol hosted on GitHub, but we need this to work in offline settings.

# Tasks

- [ ] Update `ducks/modules/protocol` to allow support local file access (desktop) (4)
- [ ] Update local file access to support cordova (2)

jthrilly:
## Total estimate: 4

wwqrd:
## Estimate (4)

- Update ducks/modules/protocol to allow support local file access (desktop) (2)
- Update local file access to support cordova (2)

jamieshark:
## Estimate (4-6)
- Update ducks/modules/protocol to allow support local file access (desktop) (2-4)
- Update local file access to support cordova (2)

rebeccamadsen:
Estimate (6)
--

- Update ducks/modules/protocol to allow support local file access (desktop) (4)
- Update local file access to support cordova (2)

PR #196, opened 2017-08-09
Feature/demo protocol loading

wwqrd:
Load demo protocol for all platforms #192
- Add LOAD_DEMO_PROTOCOL action
- Modularise code for loading protocols
- Partial implementation of local file system protocol loading for electron (subject to change, if protocols become multi-file, but developed under original scope of issue and seems worth keeping for the time being)


public/electron-starter.js
```diff
@@ -34,7 +34,7 @@ function createWindow() {
   }));
 
   // Open the DevTools.
-  mainWindow.webContents.openDevTools({ mode: 'detach' });
+  mainWindow.webContents.openDevTools(); //{ mode: 'detach' });
```
jthrilly: What's the thinking with this? I think ultimately we want to have a menu button or a key combination (or both) during alpha that opens the dev tools. Perhaps get rid of this, and I'll put up an issue for that?

wwqrd: I just found switching between windows slowed my testing down. But actually we can remove it, there's already a shortcut in our app for dev tools `Cmd+Option+i`

PR #197, opened 2017-08-10
Component/form builder

jamieshark:


wwqrd:
In theory we can point the code in https://github.com/codaco/Network-Canvas/blob/master/src/utils/fieldComponents.js at the network-canvas-ui components when they're ready.

Then we can delete the ones in https://github.com/codaco/Network-Canvas/tree/master/src/ui/Fields

jthrilly:
hey @jamieshark - I heard you might have an idea for a different approach to this. Is that right? Or else is this ready to review?

jamieshark:
@jthrilly - this is about as clean as I wanted to get it so even though it's not quite "plug and play" directly from the UI, I didn't want to make the UI components exactly specific to Redux Form. So the `fieldComponents.js` file has some helper functions so that we have the values properly working with redux form as well as some containers in case we need to finagle any styles.

wwqrd:
I'm not convinced. Having the redux form compatible "smart" components in UI doesn't preclude also having the "view" components. It'll also mean we don't need to copy this code between the NC/Server/Architect apps.

jamieshark:
@wwqrd - okay, i'll take a closer look at it.

Issue #126, opened 2017-03-08
Sociogram Interface (59)


jthrilly:
## General overview:

This interface has three purposes: (1) positioning the nodes amongst the concentric circles, (2) linking the nodes together (creating edges), and (3) selecting nodes that have a given attribute. Each instance of this interface can have one or more tasks, which are switched between using the _prompt swiper_. 

The definition of each prompt also contains the name of the variables where the result of the interaction should be stored. For example, a layout task would include the node variable to store the coordinates on. An edge creation task would define the edge type, and any other attributes that the edge should have. A node nomination task would define the node variable and its value. 

Also important is that the prompt should contain a definition of which visual variables should be shown on the layout. For example, on a given prompt, the definition should define: which nodes are highlighted, which edges are shown (could be edges of multiple types, and therefore different colors), and which layout variable is used for the node coordinates. 

# Tasks

## Research (10)

- [x] Research UI technology for interface (e.g. Canvas for edges, vectors for nodes...) (4)
- [x] Create sociogram interface (2)
- [x] Build prompt swiper/slider. Perhaps re-use the component from name generator. (4)

### Container and background component (5)

- [x] Create container (1) 
- [x] Create background component (1)
- [x] Background: Any number of rings between 0 and 7. Should also have a mode that biases size towards inner rings (skew mode) - default. I used layered opacity to achieve the variable color effect.
- [x] BONUS: Meeting today suggested that people also want to be able to provide an arbitrary image as the background for this screen. (4)

### Add node from top left (20)

- [x] Create node bucket component: Show 1 node at a time (until it's positioned), configure order they are shown (say by node attribute), once node is dropped next node appears (8)
- [x] Create Node layout component (1)
- [x] Position nodes in Node layout component (4)
- [x] When dropped, save x/y position against specified node attributes for coordinate (defined on the current prompt) (8) 

### Link nodes (20)

- [x] Create link component (1)
- [x] Build link component (canvas view showing edges). (16)
- [x] Click one node and then another to add an edge between them to the network (4)

### Select a node (4)
- [x] When a node is clicked, add the related attribute from the prompt to the node (4)

jthrilly:
Meeting with Bernie suggested that we explore alternative UI for node placement to explore biases associated with a fixed placement ("stack") of nodes in one position (i.e top right for RADAR. middle left for DPhil).

[...]

PR #213, opened 2017-08-18
Feature/sociogram interface

wwqrd:
Implementation for issue #126 

To test:

1. The first time you run this (with a clean state), you'll need to add some nodes first, this can be done from the dev console with `populateNodes(n)`, e.g. `populateNodes(5)`.
1. The sociogram inteface contains the following prompts:
    1. Position nodes on a radar view from the node bucket
    1. Add edges between nodes (friends), but not position them
    1. Select nodes (to add attributes), but not position them.
    1. Position nodes on another layout, and add a different edge between nodes (family).
    1. Position nodes on a background image from the node bucket, whist showing 'family' edges, but not allow the edges to be altered.


wwqrd:
Okay, those changes are up with the following caveats:

- I didn't add drop shadows to the edges. I think this may be possible, using filters (this is the approach I didn't have success with). Alternatively we could fully render the edge layer with an offset for a pseudo shadow effect.
- We are separating out the filters into another issue.
- The reset breaks some of the tests
- Still need to add that dialog style to the prompts (note to self)

wwqrd:
Okay, fixed the following:

- ~The reset breaks some of the tests~
- ~Still need to add that dialog style to the prompts (note to self)~

Ready to merge if you're otherwise happy with it.

pfsj:
Moving around using the menu seems to break the coordinates of the edges relative to the nodes. See the before or after below. This happens both in the browser and built versions. 

At the edge creation screen, additional edges will still be created but they will appear at the coordinates of the original nodes, not where they appear on screen - despite having to click on the nodes where they appear on the screen to create those edges.

screen shot 2017-08-23 at 10 29 28 am

screen shot 2017-08-23 at 10 29 42 am

pfsj:
Not sure if this is an issue because I'm uncertain of the permanence of the "Reset Layout" feature. But, "Reset Layout" keeps the edges that were placed on the sociogram once two or more nodes that were connected are placed back on the sociogram.

jthrilly:
The requestAnimationFrame fix has caused the following error when switching between stages:

```
getAbsoluteBoundingRect.js:42 Uncaught TypeError: Cannot read property 'scrollLeft' of null
    at getAbsoluteBoundingRect (getAbsoluteBoundingRect.js:42)
    at Droppable._this.updateZone (droppable.js:43)
    at invokeFunc (lodash.js:10350)
    at trailingEdge (lodash.js:10397)
    at timerExpired (lodash.js:10385)
```

I think this refers to a droppable component on the _previous_ interface, and is happening because the DOM object has been destroyed by the time `getAbsoluteBoundingRect()` is called.

PR #216, opened 2017-08-24
Feature/data export

### LINKED ISSUE #179: Initial data export functionality (9)
As discussed in #170, we need a basic data export function. This will live in the settings menu. More details to follow.

rebeccamadsen:
Resolves #179. This adds a menu item to export data, implements formatting for network nodes and edges into a graphml structure, and then saves the file. It uses a system "save as" dialog in electron. In cordova, it saves a temporary file that can be "shared" as the user decides (using `cordova-plugin-file` to create the temporary file and `cordova-plugin-x-socialsharing` to share). In the browser we use the default "download" function to open/save.

This uses the edge's index in the array as an id, and when you import the graphml to gephi, parallel edges will be merged (increasing the weight of the edge). Locations are assumed to have this structure:
`geographicLayout: { type: ‘layout’, x: 234, y: 345 }`. Otherwise attributes of type object are added as strings.

There is a bug with the menu, and you may have to click the menu item to export data twice to get it to do anything (I'm looking at you, Chrome). Electron doesn't seem to have this problem. I am looking into this bug, but I think there are some other issues with opening and closing the menu using the menu items, so this is probably a separate issue.


src/containers/SessionMenu.js
```diff
@@ -68,6 +76,7 @@ SessionMenu.defaultProps = {
 function mapStateToProps(state) {
   return {
     isOpen: sessionMenuIsOpen(state),
+    currentNetwork: network(state),
```
wwqrd: Why the preference for the selector here? They are generally for derived properties, but in this case we could just map directly to state: `state.network`'.

src/utils/ExportData.js
```diff
@@ -0,0 +1,193 @@
...
+    'xsi:schemaLocation="http://graphml.graphdrawing.org/xmlns\n' +
+    'http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd">\n' +
+    '  <graph edgedefault="undirected">\n' +
+    ' </graph>\n' +
+    ' </graphml>\n';
+
+  let xml = '';
+  if (window.DOMParser) { // Standard
+    xml = (new DOMParser()).parseFromString(xmlDoc, 'text/xml');
+  } else if (window.ActiveXObject) { // old IE
```
wwqrd: Do we need IE support?

src/utils/ExportData.js
```diff
@@ -0,0 +1,193 @@
...
+    dialog.showSaveDialog({ filters: [{ name: 'graphml', extensions: ['graphml'] }] }, (filename) => {
+      if (filename === undefined) return;
+      fs.writeFile(filename, data, (err) => {
+        if (err) {
+          dialog.showErrorBox('Error Saving File.', err.message);
+        }
+      });
+    });
+  } else if (isCordova()) { // cordova save to downloads
+    window.requestFileSystem(window.LocalFileSystem.TEMPORARY, 5 * 1024 * 1024, (fileSystem) => {
```
wwqrd: Might be worth using a promise library to make these callbacks/error handling easier to read. We've already got bluebird installed so we could use that:

```js
import Promise from 'bluebird';

const getFileSystem = () => new Promise((resolve, reject) => {
  window.requestFileSystem(..., (fileSystem) => { resolve(fileSystem); }, (err) => reject(err) );
});

...

getFileSystem
  .then(getFile)
  .then(createWriter)
  .then(() => {
    // blob
  }).
  .error((err) => { // handle any errors })

File: src/utils/ExportData.js [hunk 4]
```diff
@@ -0,0 +1,193 @@
...
+      element.style.display = 'none';
+      document.getElementById('root').appendChild(element);
+      element.click();
+      document.getElementById('root').removeChild(element);
+      window.URL.revokeObjectURL(blob);
+    }
+  }
+};
+
+const createGraphML = (networkData) => {
```
wwqrd: This looks really nice 👌

src/utils/ExportData.js
jthrilly: Nope! This can be removed.

src/utils/ExportData.js
jthrilly: Elsewhere I have used Promises without bluebird. Is there a reason to use it? Isn't it supported in our target platforms?

wwqrd: Well these days I suppose it's more of a polyfill. If we have native implementations on all our platforms then lets use that. I think this comes with some helper methods for "promisifying" things, but we aren't using those afaik.

jthrilly: Good to know. I will remove this dep after I've tested that native promises are working correctly where I've used them on my branch.

jthrilly:
Build is currently failing because of the use of alert. I can see why you want it, but could you use the new Dialog component instead for this?

wwqrd:
@jthrilly RE the build fail. Interesting that it fails on Travis and not locally?

jthrilly:
@wwqrd - yeah its because Travis treats warnings as fails. Its a stricter environment in that sense, which I think is a good thing. We shouldn't allow build warnings in the master branch.

PR #221, opened 2017-08-28
Auto scale node text size

wwqrd:
Request for review not merge.

This shrinks the text inside a node until it fits. It starts at the initial stylesheet size, and works down from there.

Known limitations: Is run every time props change on a component.

Any thoughts or feedback?

Things left to do:
- Extract scaleTextToFit into own util
- Move scaleTextToFit and Node back into UI

### DIFF
--- a/src/components/Elements/Node.js
+++ b/src/components/Elements/Node.js
@@ -2,29 +2,34 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-const childOutOfBounds = (parentBounds, childBounds) =>
-  childBounds.height > parentBounds.height || childBounds.width > parentBounds.width;
+const textOutOfBounds = (parent, child) => {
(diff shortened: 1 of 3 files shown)


jamieshark:
Hmm. @wwqrd @jthrilly - do we know if there will be any character limit at all on what will be displayed in the nodes?

jthrilly:
@jamieshark the design specification called for 8 characters, since we identified this as covering 95% of data we have encountered so far. This text sizing issue is also about scaling up text for fewer characters. One use case we can test is something like a single character emoji. 

At and beyond 8 characters, injecting line-breaks in combination with reducing text size seems to be the way to go.

jthrilly:
SO based on my previous comment, I think we would like to see scaling _up_ as much as scaling down. @wwqrd can you work this in?

jthrilly:
This is ready to merge (tested and working).

PR #226, opened 2017-08-31
Feature/protocol error dialog

jthrilly:
This PR adds "check for update" to the session menu, along with a flow through the update process using dialogs, that (hopefully) catches and deals with errors along the way.

*note:* this will only work from within electron. The update will only complete from within a fully built version of the app (`yarn run dist:mac` etc.). Checking for updates will work doing `yarn run build; electron ./www/`.


src/containers/SessionMenu.js
```diff
@@ -21,12 +23,29 @@ function addMockNodes() {
   * @extends Component
   */
 class SessionMenu extends Component {
-  componentWillMount() {
-    this.props.registerModal('EXPORT_DATA');
```
rebeccamadsen: So you don't need to register a modal to use it? Huh.

src/containers/SessionMenu.js
```diff
@@ -47,17 +66,145 @@ class SessionMenu extends Component {
...
+            title: 'An Error Occured',
+            type: 'error',
+            content: 'An error has occured during the update process. More information to help diagnose the issue can be found below.',
+            additionalInformation: error,
+            onConfirm: () => {},
+            confirmLabel: 'Okay',
+            hasCancelButton: false,
+          },
+        });
+        // This shouldn't be here. But setState is async. How to handle better?
```
rebeccamadsen: An `await` or something?

src/containers/App.js
```diff
@@ -69,11 +48,18 @@ App.defaultProps = {
   isSessionMenu: false,
 };
 
+function mapDispatchToProps(dispatch) {
+  return {
+    closeModal: bindActionCreators(modalActions.closeModal, dispatch),
```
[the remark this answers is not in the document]
jthrilly: Correct. This is left over. I'll remove it.

src/containers/SessionMenu.js
jthrilly: Not anymore!  It's a recent change.

src/containers/SessionMenu.js
jthrilly: It's weird. I tried using the setState callback but that didn't work. Not sure about await, but I'll try it.

src/components/Elements/Dialog.js
```diff
@@ -34,7 +36,24 @@ const Dialog = (props) => {
     'error': 'neon-coral',
   }
 
+  function createMarkup() {
```
wwqrd: Like I said before, I'd rather we used a bundle release notes for this to avoid the possibility of malicious code. But at least there are some precautions in place.

src/containers/SessionMenu.js
wwqrd: It's been factored out into a HOC: https://github.com/codaco/Network-Canvas/blob/66a143abc915472ab158d5cd4c32ba0e314a1afc/src/behaviours/modal.js#L27

src/containers/SessionMenu.js
```diff
@@ -47,17 +66,145 @@ class SessionMenu extends Component {
...
   };
 
+  checkForUpdates = () => {
+    // Initialise the update checker. Returns a promise
+    updater.checkForUpdate().then(
+      (response) => {
+        // Update available.
+        const version = response.version.toString();
+        const releaseNotes = response.releaseNotes.toString();
+        this.setState(...this.state, {
```
wwqrd: `setState()` is partially applied, so the `...this.state` shouldn't be needed.

jamieshark:
I also like @wwqrd suggestion of factoring out the dialog content / state to redux. Are we also sure that we're always only ever going to have one dialog open at a time? Maybe making a `dialogStack` sort of array might be of interest?

PR #239, opened 2017-09-11
Feature/simplify menu

### LINKED ISSUE #232: Simplify menu syntax
# Purpose

Adding new elements to the menus, like utility functions requires additional code inside the respective menu components, that would probably make more sense inside App. Using a more react-like syntax might clarify usage and allow us to better place responsibilities.

# Specification

Use more React-like syntax, where concerns are nested less deeply. And callbacks can be handled elsewhere.

e.g.

```
// In App.js
<Menu>
  <MenuItem onClick={doThing()} />Do thing</MenuItem>
  <MenuItem onClick={doOtherThing()}>Do other thing</MenuItem>
</Menu>
```

rebeccamadsen:
Fixes #232. Per conversation in slack, this factors out MenuContent and creates a scrollable component. 

As a small added bonus, this cleans up the dom structure a little to get rid of a few unnecessary divs.

It is possible we want to revisit the organization of StageMenu and/or SessionMenu at a future date, having a MenuItem component closer to the creation of the containers, but it isn't clear yet.


wwqrd:
Looks good to me.

PR #243, opened 2017-09-20
Fix/edit node bug

### LINKED ISSUE #233: Edit node form closing instantly after opening
# Specification

Our edit form modal is receiving a click event from somewhere after it opens on the close button. This is automatically closing the form.

We have a work around in place (which adds a delay before the button is clickable), but it'd be great to find the root cause.

Likely candidate is the Form component.

rebeccamadsen:
Fixes #233.

Ultimately this was a "ghost click" issue. See the "tip" on [“After a tap, also a click is being triggered, I don’t want that!”](http://hammerjs.github.io/tips/). It looks like hammer-time [may be a fix](https://github.com/hammerjs/hammer.js/issues/808#issuecomment-125267066), but it is not a react library, so I focused on the[ PreventGhostClick solution ](https://gist.github.com/jtangelder/361052976f044200ea17)from the "tip". This lets us simplify Modal.js again.


src/behaviours/draggable.js
```diff
@@ -75,6 +76,7 @@ export default function draggable(WrappedComponent) {
     }
 
     componentDidMount() {
+      PreventGhostClick(findDOMNode(this.node));
```
wwqrd: I have a feeling this should live in `selectable`, since this class doesn't trigger the click/tap callback?

For example if we did:
`const EnhancedNode = selectable(node);`

Would we get the same double click event again?

rebeccamadsen: Good catch! Will move the call.

wwqrd:
Looks great! :shipit:

jthrilly:
@rebeccamadsen ready to merge as far as I can see :)

Issue #130, opened 2017-03-08
"Variable registry" - a component that contains a list of variables, how they are defined, validation functions, etc. Provides an api for other modules/interfaces and the interview protocol to register variables


jthrilly:

jthrilly:
I'm reopening this with more details to follow. 

Basically I think it would be advantageous to have some kind of registry for variables used in the interview. Other components/views could reference this registry to enhance their functionality, or simply to better understand the nature of a variable.

Examples:

- Coordinates can be stored on any variable, but the exporter functionality needs to know which variables are intended to store these values because some file formats require special syntax for these variables. 
- Variable constraints could be defined in the registry and attached to variables, meaning that we don't need to duplicate validation code. This could be redux-form compatible. 
- The registry could also contain human readable descriptions of variables, which could be used by Architect and Server.

More to come.

wwqrd:
Had a little think about this.

We have three points of variable creation at the moment:

- `nodeAttributes`
- layouts
- form fields

I think it probably does make sense to have a collated variable registry for these, which would include data types and other meta data. Perhaps we can leverage `PropTypes` for the data types.

Where these variables are defined currently would be replaced with a reference to the variable. Perhaps by name, or perhaps a literal reference.

For one it'd mean we can simplify the export methods.

I'm not convinced it makes sense to put validation data in the registry: what do we do with data that doesn't match the validation requirements? When would data be validated? I think these questions don't make sense because the validation is specific to the forms.

PR #247, opened 2017-09-28
Feature/variable registry

wwqrd:
Feedback on this please!

It's mostly a working implementation of a variable registry.

Relates to: #130.

- The form config has been updated to refer to pre existing variables.
- The form component has been updated to rehydrate variables based on the registry. You can also specify them directly using objects -- any thought on overloading it like this?
- I refactored some of the Form/NodeForm props to be more explicit (specific fields rather than just a `form` object), and combined form/modal names - removing them from the config in fact - because they seem redundant.
- The other changes relate to linting/test updates


jthrilly:
Following our phone call, here are some changes:

- Perhaps look at tidying up the structure of the protocol file itself, so that variables, forms, and even stages are defined in their own const and then simply linked into the config object that is actually being exported. This should make things more readable and also easier to generate within Architect. We definitely want to do this for at least the forms, so that they can reused more easily across interfaces.
- Variables need to have a property indicating if they are intended for a node or an edge. 
- Rename 'rules' property to 'validation' and show an example of a custom function for validation.
- Form definitions need to be more complex than a list of pointers to the variable registry, since we also want to be able to specify which UI component should be connected to this variable. I will provide a comprehensive list of variable types and mappings to input components (both ones we have already implemented and ones we still need to make). Until then, we can still do this with (for example) text or numerical inputs, which map directly to the straight "<Input>" component. We eventually want to support providing a custom component within the protocol.

wwqrd:
Okay those changes are in and the tests should be passing now.

PR #249, opened 2017-10-12
Feature/mobile form ui

### LINKED ISSUE #234: Mobile form UI for the add new node form
# Purpose

The 4:3 aspect ratio experience for the new node modal is currently missing.  The key difference from the existing modal is that it aims to avoid obscuring the inputs by only displaying one at a time, with next/previous functionality to navigate through the form.

There is an example of a wizard style form in the Redux Form docs (http://redux-form.com/6.7.0/examples/wizard/), but we will probably want to make this work with CSS only if possible.

05network_canvas_addperson_contexts_4x3

# Specification

- [x] Update Form component to automatically generate a wizard style form from the protocol.

rebeccamadsen:
Resolves #234.

This implements the mobile version of the add/edit node form. As discussed, this uses `min-device-aspect-ratio`, because the aspect-ratio changes in android once the keyboard opens. Also, using css to show/hide parts of the form keeps validation and auto-populate from functioning correctly (it wants to validate everything at once!), so this uses javascript to manage which field of the form to render.

This change can be viewed cordova or in the chrome developer tools by toggling the device toolbar. Clicking on a node or adding a new node shows the mobile version of the form.

I still debate the merits of pulling out the wizard-y aspects from NodeForm into a separate component, and whether that makes sense now, or when/if we need it in the future.


src/components/Elements/Modal.js
```diff
@@ -23,7 +24,7 @@ function Modal(props) {
       transitionLeaveTimeout={animation.duration.standard}
     >
       { show &&
-        <div key="modal" className="modal" onClick={() => close()}>
+        <div key="modal" className={`modal ${className}`} onClick={() => close()}>
```
wwqrd: We could use the classnames lib for this for consistency?

`cx('modal', ...className)`

src/containers/Elements/Form.js
```diff
@@ -119,7 +135,9 @@ export default compose(
   connect(makeMapStateToProps),
   autoInitialisedForm,
   reduxForm({
-    destroyOnUnmount: true,
+    destroyOnUnmount: false, // need this false to make it validate across wizard
```
wwqrd: We should probably also give the form a name in this case so redux can distinguish between two forms that have similar field names?

src/containers/Elements/NodeForm.js
```diff
@@ -51,26 +74,51 @@ class NodeForm extends Component {
...
+      <div>
+        {this.state.fieldIndex !== 0 && <Icon name="form-arrow-left" onClick={this.previousField} />}
+      </div>);
+    const lastField = this.state.fieldIndex === fields.length - 1;
+    const nextElement = lastField || large ? null :
+      (<Icon name="form-arrow-right" />);
+
     return (
-      <Modal name={name} title={title}>
+      <Modal name={name} title={title} close={this.close} className={large ? '' : 'modal--mobile'}>
```
wwqrd: Could be: `cx({ 'modal--mobile': !large })`

src/containers/Elements/NodeForm.js
```diff
@@ -51,26 +74,51 @@ class NodeForm extends Component {
...
     const {
       title,
       fields,
+      autoPopulate,
       name,
       addAnother,
       initialValues,
     } = this.props;
 
+    const large = window.matchMedia('screen and (min-device-aspect-ratio: 16/9)').matches;
```
wwqrd: I'd be inclined to make this a class method with the name `isLarge()`

src/containers/Elements/NodeForm.js
```diff
@@ -51,26 +74,51 @@ class NodeForm extends Component {
...
     return (
-      <Modal name={name} title={title}>
+      <Modal name={name} title={title} close={this.close} className={large ? '' : 'modal--mobile'}>
+        { !large && <div className="modal__pips">
+          <Pips count={fields.length} currentIndex={this.state.fieldIndex} />
+        </div>}
         <Form
-          fields={fields}
-          initialValues={!addAnother ? initialValues : null}
+          fields={large ? fields : [fields[this.state.fieldIndex]]}
```
wwqrd: This could be a class level getter `getFields()`

src/containers/Elements/NodeForm.js
```diff
@@ -51,26 +74,51 @@ class NodeForm extends Component {
...
+          autoPopulate={autoPopulate}
+          initialValues={initialValues}
           autoFocus
           form={name.toString()}
-          onSubmit={this.onSubmit}
+          onSubmit={lastField || large ? this.onSubmit : this.nextField}
```
wwqrd: What does this do? Can we make it easier to read with named functions / variables?

jthrilly: It decides if we need to submit, or if we need to go to the next field (depending on if we are large). Seems pretty readable to me.  I guess it could be broken out into a function. 

How do we feel about inline comments for things like this? Should the code be self-explanatory?

wwqrd: I read it as: "string | probablyBoolean ? eventHandler() : propertyGetterProbablyString()"

So if it's like you say I'd change it to:

`onSubmit`

rebeccamadsen: Makes sense. I can make these changes!

jthrilly:
I know it is somewhat out of spec, but could you also include this plugin in this branch: https://github.com/mesmotronic/cordova-plugin-fullscreen

Specifically, we want to activate "immersive mode" on android so we don't get nav bars at the top and bottom.

rebeccamadsen:
> I know it is somewhat out of spec, but could you also include this plugin in this branch: https://github.com/mesmotronic/cordova-plugin-fullscreen

I can look into this and see if it is straightforward. Though it *is* out of spec. ;)

jthrilly:
Looks like we just need to add `<preference name="Fullscreen" value="true" />` to config.xml, actually.

src/containers/Elements/Form.js
rebeccamadsen: NodeForm passes in `form` as a prop which becomes the name of the redux-form. So as of now we have three forms that show up in redux: "setup", "Symbol(EDIT_NODE)", and "Symbol(ADD_NODE)". I can more explicitly require `form`.

jthrilly:
> Yes, that works, however there is an open issue that affects us too. So full screen works until the soft keyboard opens. Given that it is an open issue, I will guess switching back to full screen after the keyboard closes is not an easy fix?

Annoying! There are some workarounds, but I think its a big enough issue to do separately. Leave it out for now.

rebeccamadsen:
> I think the previous/next buttons should probably be vertically centred?

Maybe so; I considered that. But vertically centered with respect to what? The exact center of the page? The center of the input field? I played briefly with that, but having two buttons on the right for new nodes complicates positioning, and having the keyboard possibly shift things up/down on open/close also did. I'm open on tweaks though.

jthrilly:
> Maybe so; I considered that. But vertically centered with respect to what? The exact center of the page? The center of the input field? I played briefly with that, but having two buttons on the right for new nodes complicates positioning, and having the keyboard possibly shift things up/down on open/close also did. I'm open on tweaks though.

The buttons should be vertically centered relative to the viewport, so that as the keyboard opens, they shift up, and are vertically centered within the remaining space. I think this should be possible with viewport units or flex box.

jthrilly:
Following a discussion, we have decided to move wizard functionality from NodeForm to Form, and have a prop (wizard?) that tells the form to be in wizard mode. The form will then handle progress through the form.

Modal will have a 'fullscreen' prop (or similar), and Form will have a 'wizard' prop (or similar).

Other changes (if deemed within scope):
- be able to pass a submit button component to Form, allowing custom submit button text and icon.
- remove continuousSubmit functionality (to be reimplemented)

rebeccamadsen:
I've moved the wizard logic to a new component, outside of NodeForm. I did not put the wizard logic inside Form; I thought it would be simpler as its own component. As with anything there are tradeoffs --
 I'm open to discussing this.

I also teased out "fullscreen" scss from "wizard" scss. So theoretically you could full-screen a modal without it being a wizard, or wizardize a form without it being fullscreen, by toggling two css classes.

The submit/add another/next buttons now are part of a "controls" prop, so Form just renders whatever controls it is given, with the default control being a "Submit" button. This meant moving continuous submit out of Form. I put it in NodeForm for now, but if we want to refactor it further we could.

src/containers/Elements/Form.js
```diff
@@ -73,32 +71,29 @@ class Form extends Component {
...
+  controls: PropTypes.array,
 };
 
 Form.defaultProps = {
   autoFocus: false,
-  addAnother: false,
   autoPopulate: null,
-  continuousSubmit: null,
-  normalSubmit: null,
+  controls: [<Button key="submit" aria-label="Submit">Submit</Button>],
```
wwqrd: 👌

src/containers/Elements/FormWizard.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  render() {
+    const { fields, onSubmit, ...rest } = this.props;
+
+    const extraProps = {};
+    extraProps.fields = [fields[this.state.fieldIndex]];
+    extraProps.onSubmit = this.onSubmit;
+    if (this.shouldShowNextField()) {
+      extraProps.controls = [(
```
wwqrd: Just a thought, but this seems all seems little opaque?

What about something more like:
```javascript
currentField() {
  return [this.props.fields[this.state.fieldIndex]];
}

render() {
  const nextButton = (
    <button key="next" className="form__next-button" aria-label="Submit">
      <Icon name="form-arrow-right" />
    </button>
  );

  const formProps = {
    ...this.props,
    fields: this.currentField(),
    onSubmit: this.onSubmit,
  }

  // ...

  <Form
     {...formProps}
     controls={this.shouldShowNextButton() ? [nextButton] : undefined}
  />
}
```

src/containers/Elements/NodeForm.js
```diff
@@ -51,50 +61,70 @@ class NodeForm extends Component {
...
+      autoPopulate,
+      fields,
       initialValues,
+      name,
+      title,
     } = this.props;
 
+    const modalClassNames = cx({ 'modal--fullscreen': !this.isLarge() });
+
+    const props = {};
```
wwqrd: Same as `extraProps` in wizard from.

Would probably take away the need to specify `autoPopulate` in propTypes.

wwqrd:
👍

jthrilly:
This is brilliant. I think it is structured far more sensibly than before, and we can merge the form components again in the future should we need to. For now we just need to ensure that any changes to the form API are replicated in both. 

I think this should be merged.

PR #251, opened 2017-10-13
Feature/flatter protocol

jthrilly:
Removes params from interface API in the protocol.

Any other flattening/tidying up people think we can do here?


src/components/Elements/MenuItem.js
```diff
@@ -4,10 +4,17 @@ import classNames from 'classnames';
...
 
 /**
-  * Renders a menu item. Image is based on interfaceType.
+  * Renders a menu item. Image is based on icon.
   */
 function MenuItem(props) {
-  const { onClick, interfaceType, isActive, menuType, title } = props;
+  const { onClick, interfaceType, icon, isActive, menuType, title } = props;
+
+  const iconMap = {
```
wwqrd: I'd be inclined to set an assumed format so that we don't need to keep adding to this list?

We already have change-case on the project so could do:

```
const icon = `menu-${changeCase.paramCase(interfaceType)}`;
```

src/ducks/modules/prompt.js
```diff
@@ -17,14 +17,14 @@ export default function reducer(state = initialState, action = {}, stageState) {
     case protocolActionTypes.SET_PROTOCOL:
       return {
         ...initialState,
-        counts: action.protocol.config.stages.map(stage => stage.params.prompts.length),
+        counts: action.protocol.config.stages.map(stage => stage.prompts.length),
```
wwqrd: Is this part of the state still in use? We may be able to remove this.

src/components/Elements/MenuItem.js
```diff
@@ -4,10 +4,17 @@ import classNames from 'classnames';
...
 /**
-  * Renders a menu item. Image is based on interfaceType.
+  * Renders a menu item. Image is based on icon.
   */
 function MenuItem(props) {
-  const { onClick, interfaceType, isActive, menuType, title } = props;
+  const { onClick, interfaceType, icon, isActive, menuType, label } = props;
+
+  const iconMap = {
+    NameGenerator: 'menu-name-generator',
```
wwqrd: I'd be inclined to set an assumed format so that we don't need to keep adding to this list?

We already have change-case on the project so could do:

```
const icon = `menu-${changeCase.paramCase(interfaceType)}`;
```

jthrilly: Yeah that would be better, but we have got ourselves into a bit of a naming pickle with interfaces and icons which makes it annoying. We have `NameGenerator` as the interface, but `menu-name-generator` as the icon name.

I was being lazy here with the mapping, but what really needs to happen is that the icons should be renamed to match the interface names.

wwqrd:
These changes are also in the branch which updates the components to work with the new protocol so I'm going to close this PR.

PR #256, opened 2017-10-31
Feature/flatter protocol upgrade components

wwqrd:
I've updated the components to work with the new protocol format defined in `feature/flatter-protocol`.

Depending on the age of that branch, we may find this introduces some more breaking changes. @rebeccamadsen @jthrilly, anything come to mind?

Additionally this contains code that loads the demo protocol from a pure JSON file contained in a folder - this is how we decided to move forward with the protocol format when @jthrilly and I collaborated last week.

One thing I noticed was some subtle differences in how the definitions for the interfaces for NameGenerator and Sociogram. `nodeAttributes`, and `nodeType`, currently, perhaps more in future? Because of that I've started to separate out the selectors, into interface specific ones, but perhaps this would be component specific in future? - for example the `NodeBin` doesn't strictly belong to the `Sociogram`. Any pointers or suggestions for better organisation, welcome!


wwqrd:
@jthrilly Okay I refactored the selectors based on our conversion about `subject: { entity: 'node', type: 'person' }`, being generic for stages/prompts, and automatically finding it.

wwqrd:
> I'm pretty good with what you have. Although the last commit I am getting an error:
Line 6: Using exported name 'LayoutNode' as identifier for default export import/no-named-as-default
> It could just be a problem with my build not updating or something?

On this one I think a change in the lint rules is in order.

What I'm doing is exporting each component twice:

Once as a pure class: `export { MyComponent }:`
And then after all the higher order components are applied: `export default connect(mapStateToProps)(MyComponent);`

I'm led to believe this is a pretty common practice with react/redux to enable testing without needing to mock out the whole store, and is in the official docs:

https://github.com/reactjs/redux/blob/master/docs/recipes/WritingTests.md
https://github.com/benmosher/eslint-plugin-import/issues/468
https://github.com/benmosher/eslint-plugin-import/issues/544

> Just a minor one left: the "hidden" date created field in the add node form is not actually hidden, it just isn't labeled. That may be a UI thing though?

We don't have this kind of component yet, even though it's in the new protocol spec. I think maybe implement it separately?

PR #257, opened 2017-10-31
Feature/ordinal bin interface

SunjayDK:
This adds the ordinal bin interface!

I refactored the NodeBucket component slightly to allow it to be used in this context. Specifically, the onDropNode/handleDropNode function was moved from the NodeBucket out into the interface.

This does not include the tests yet; unfortunately I didn't have the chance to write them yet. It seems to function correctly currently though so they could be pretty easily built out from this.


rebeccamadsen:
I did not try altering the protocol yet or look closely at the code itself, just focused in this first look at the functionality.
* general layout needs to scale better
* opacity of the last column seems really dark? is this what specs call for?
* size of the drop area should fill the column
* if you change prompts, the placements of nodes are the same on every prompt. I imagine different prompts ought to allow different node placements.
* there is a hover effect on bins? Is this right? If so, should there be a hover effect on the farthest left column too?
* similarly the color of the drop area for the farthest left doesn't really show
* the mouse changes to a hand when you hover over nodes, even if you cannot select the node (do we have an "inactive" class or something?)
* menu item text probably needs a space between Ordinal and Bin, and the icon is just the default right now; I think the specs have a different icon
* bins should be scrollable. if you fill a bin up the overflow is funky
* download data shows labels like "ORDINALBIN-3" instead of the user friendly text

PR #259, opened 2017-11-17
Feature/change drag drop

### LINKED ISSUE #246: Changes to drag and drop behavior in name generator
After user testing we have decided to change the drag and drop behaviour on the name generator interface.

The changes are as follows:

- When a node drag starts, the droppable area _immediately_ highlights. Current behaviour is that the droppable area highlights when the node is over it.
- The behaviour for the "people you already mentioned" panel is changing from "tap to select", to "drag into the node list". That is, instead of the user tapping a node to nominate it when it is in this panel (causing it to be come highlighted), the user will instead drag nodes from this panel into the main node list, as is done with the "people from your previous visit" panel.

Other changes:
- The panel title should be a configurable property
- The color of nodes in the side panels should be a configurable property.
- The node should not immediately be removed from its container when it is being dragged. There should be a blank space where the node was, which disappears once the node is dropped.

rebeccamadsen:
Resolves #246. 

I did not commit my protocol changes, but this is an example of what I was using with the panels in my branch:

> "panels": [
        {
          "title": "Can we change these?",
          "dataSource": "existing",
          "filter": "network => network",
          "nodeColor": "node-alt-5",
          "droppable": true,
          "highlight": "edge-alt-3"
        },
        {
          "title": "Yes, we can!",
          "dataSource": "previousInterview",
          "filter": "network => network",
          "nodeColor": "node-alt-9",
          "highlight": "edge-alt-4"
        }
      ],

Titles already were configurable.

Everything but node colors will work in this branch with what is in the UI master right now. But to experiment with different node colors, you will need [this PR](https://github.com/codaco/Network-Canvas-UI/pull/26) and to add a `nodeColor` in the protocol for panels. In the code, we just need a `color` prop for a `Node`. Example of colors in use:
screenshot

Dragging a node with the area "highlighted":
screenshot2

Droppability indicated with the `+` node appearance, and space is left in the panel the node it dragging from until the node it dropped:
screenshot3

Side panel is also droppable to remove an added node (but not a node created on the current prompt):
screenshot4

This additionally makes the trash bin effect change on hover by toggling a classname in DropZone.

### DIFF
--- a/src/behaviours/draggable.js
+++ b/src/behaviours/draggable.js
@@ -254,7 +254,7 @@ export default function draggable(WrappedComponent) {
     }
 
     styles() {
-      return this.state.dragStart ? { opacity: 0, width: 0, height: 0 } : { opacity: 1 };
+      return this.state.dragStart ? { visibility: 'hidden' } : { visibility: 'initial' };
     }
 
     render() {
--- a/src/behaviours/draggable.js
+++ b/src/behaviours/draggable.js
@@ -178,7 +178,7 @@ export default function draggable(WrappedComponent) {
     detectDragStart = (movement) => {
       if (this.state.type === 'DRAG' && !this.preview && movement.distance > 4) {
         this.setState({ dragStart: true }, () => {
-          this.props.dragStart(this.props.draggableType, this.props.originIds);
+          this.props.dragStart(this.props.draggableType, this.props.meta);
           this.createPreview();
         });
       }
@@ -288,7 +288,7 @@ export default function draggable(WrappedComponent) {
     onDropped: PropTypes.func,
     onMove: PropTypes.func,
     updateActiveZones: PropTypes.func.isRequired,
-    originIds: PropTypes.object,
+    meta: PropTypes.object,
     canDrag: PropTypes.bool,
   };
 
@@ -297,13 +297,13 @@ export default function draggable(WrappedComponent) {
     animate: false,
     onDropped: () => {},
     onMove: () => {},
-    originIds: {},
+    meta: {},
   };
 
   function mapStateToProps(state, props) {
     return {
       zones: state.droppable.zones,
-      originIds: { promptId: props.promptId, stageId: props.stageId },
+      meta: { promptId: props.promptId, stageId: props.stageId },
     };
   }
 
--- a/src/behaviours/draggable.js
+++ b/src/behaviours/draggable.js
@@ -178,7 +178,7 @@ export default function draggable(WrappedComponent) {
     detectDragStart = (movement) => {
       if (this.state.type === 'DRAG' && !this.preview && movement.distance > 4) {
         this.setState({ dragStart: true }, () => {
-          this.props.dragStart(this.props.draggableType);
+          this.props.dragStart(this.props.draggableType, this.props.originIds);
           this.createPreview();
         });
       }
@@ -288,6 +288,7 @@ export default function draggable(WrappedComponent) {
     onDropped: PropTypes.func,
     onMove: PropTypes.func,
     updateActiveZones: PropTypes.func.isRequired,
+    originIds: PropTypes.object,
     canDrag: PropTypes.bool,
   };
 
@@ -296,11 +297,13 @@ export default function draggable(WrappedComponent) {
     animate: false,
     onDropped: () => {},
     onMove: () => {},
+    originIds: {},
   };
 
-  function mapStateToProps(state) {
+  function mapStateToProps(state, props) {
     return {
       zones: state.droppable.zones,
+      originIds: { promptId: props.promptId, stageId: props.stageId },
     };
   }
 
--- a/src/behaviours/droppable.js
+++ b/src/behaviours/droppable.js
@@ -35,12 +35,6 @@ export default function droppable(WrappedComponent) {
       this.updateZone();
     }
 
-    componentDidUpdate(prevProps) {
-      if (prevProps.acceptsDraggableType !== this.props.acceptsDraggableType) {
-        this.props.updateAcceptedType(this.props.droppableName, this.props.acceptsDraggableType);
-      }
-    }
-
     componentWillUnmount() {
       window.removeEventListener('resize', this.onResize);
       this.onResize.cancel();
@@ -85,7 +79,6 @@ export default function droppable(WrappedComponent) {
 
   Droppable.propTypes = {
     updateZone: PropTypes.func.isRequired,
-    updateAcceptedType: PropTypes.func.isRequired,
     droppableName: PropTypes.string,
     acceptsDraggableType: PropTypes.string,
     hover: PropTypes.bool,
@@ -116,7 +109,6 @@ export default function droppable(WrappedComponent) {
   function mapDispatchToProps(dispatch) {
     return {
       updateZone: bindActionCreators(droppableActions.updateZone, dispatch),
-      updateAcceptedType: bindActionCreators(droppableActions.updateAcceptedType, dispatch),
     };
   }
 
--- a/src/behaviours/droppable.js
+++ b/src/behaviours/droppable.js
@@ -35,6 +35,12 @@ export default function droppable(WrappedComponent) {
       this.updateZone();
     }
 
+    componentDidUpdate(prevProps) {
+      if (prevProps.acceptsDraggableType !== this.props.acceptsDraggableType) {
+        this.props.updateAcceptedType(this.props.droppableName, this.props.acceptsDraggableType);
+      }
+    }
+
     componentWillUnmount() {
       window.removeEventListener('resize', this.onResize);
       this.onResize.cancel();
@@ -79,6 +85,7 @@ export default function droppable(WrappedComponent) {
 
   Droppable.propTypes = {
     updateZone: PropTypes.func.isRequired,
+    updateAcceptedType: PropTypes.func.isRequired,
     droppableName: PropTypes.string,
     acceptsDraggableType: PropTypes.string,
     hover: PropTypes.bool,
@@ -109,6 +116,7 @@ export default function droppable(WrappedComponent) {
   function mapDispatchToProps(dispatch) {
     return {
       updateZone: bindActionCreators(droppableActions.updateZone, dispatch),
+      updateAcceptedType: bindActionCreators(droppableActions.updateAcceptedType, dispatch),
     };
   }
 
--- a/src/components/Elements/DropZone.js
+++ b/src/components/Elements/DropZone.js
@@ -1,28 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import cx from 'classnames';
 
 import { droppable } from '../../behaviours';
 
 /**
   * Renders a container onto which a `draggable` can be dropped.
   */
-const DropZone = ({ children, hover }) => (
+const DropZone = ({ children }) => (
   <div className="drop-zone">
-    {React.Children.map(
-      children,
-      (child => React.cloneElement(child, { className: cx(child.props.className, { hover }) })))}
+    {children}
   </div>
 );
 
 DropZone.propTypes = {
   children: PropTypes.any,
-  hover: PropTypes.bool,
 };
 
 DropZone.defaultProps = {
   children: null,
-  hover: false,
 };
 
 export default droppable(DropZone);
--- a/src/components/Elements/DropZone.js
+++ b/src/components/Elements/DropZone.js
@@ -1,23 +1,28 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import cx from 'classnames';
 
 import { droppable } from '../../behaviours';
 
 /**
   * Renders a container onto which a `draggable` can be dropped.
   */
-const DropZone = ({ children }) => (
+const DropZone = ({ children, hover }) => (
   <div className="drop-zone">
-    { children }
+    {React.Children.map(
+      children,
+      (child => React.cloneElement(child, { className: cx(child.props.className, { hover }) })))}
   </div>
 );
 
 DropZone.propTypes = {
   children: PropTypes.any,
+  hover: PropTypes.bool,
 };
 
 DropZone.defaultProps = {
   children: null,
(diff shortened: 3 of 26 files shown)


src/behaviours/draggable.js
```diff
@@ -296,11 +297,13 @@ export default function draggable(WrappedComponent) {
...
     onDropped: () => {},
     onMove: () => {},
+    originIds: {},
   };
 
-  function mapStateToProps(state) {
+  function mapStateToProps(state, props) {
     return {
       zones: state.droppable.zones,
+      originIds: { promptId: props.promptId, stageId: props.stageId },
```
wwqrd: Does this mean that another draggable on the same screen would be considered the same origin? Suggest we either manually provide a unique id, or we could auto generate an id on creation using `uniqueId` in lodash?

src/containers/Elements/NodeProvider.js
```diff
@@ -60,30 +66,53 @@ class NodeProvider extends Component {
...
       nodes,
+      nodeColor,
+      showPanelAlways,
+      currentIds,
+      isOrigin,
     } = this.props;
 
     const label = node => `${node.nickname}`;
     const selected = node => isMatch(node, this.props.activePromptAttributes);
+    const nodeType = isOrigin ? null : 'EXISTING_NODE';
```
wwqrd: Seems like a work around, does accepted _type_ actually change?

src/containers/Elements/NodeProviderPanels.js
```diff
@@ -74,30 +74,51 @@ class NodeProviderPanels extends PureComponent {
...
-            minimise={!this.isPanelOpen(panelIndex)}
+            minimise={!this.isPanelAlwaysOpen(panelIndex) && !this.isPanelOpen(panelIndex)}
             highlight={getHighlight(panel, panelIndex)}
           >
             <NodeProvider
               {...panel}
               stage={stage}
               prompt={prompt}
               onUpdateNodes={nodes => this.onUpdatePanelState(panelIndex, nodes.length !== 0)}
+              showPanelAlways={show => this.showPanelAlwaysOpen(panelIndex, show)}
```
wwqrd: Node provider probably shouldn't have any knowledge of a panel. Maybe NodeProvider could have something like an `onDrag` prop? and

src/selectors/interface.js
```diff
@@ -72,26 +72,26 @@ export const makeNetworkNodesForSubject = () => {
 };
 
 export const makeNetworkNodesForPrompt = () => {
-  // used to check node attributes
-  const getIds = makeGetIds();
+  // used to check prompt ids
+  const getAttributes = makeGetAdditionalAttributes();
```
wwqrd: Why the re-revision?

Maybe we should create a couple of new methods with names that better describe what they do?

Also, if the notable change has been reverted we can probably remove the comment.

src/behaviours/draggable.js
```diff
@@ -254,7 +254,7 @@ export default function draggable(WrappedComponent) {
     }
 
     styles() {
-      return this.state.dragStart ? { opacity: 0, width: 0, height: 0 } : { opacity: 1 };
+      return this.state.dragStart ? { visibility: 'hidden' } : { visibility: 'initial' };
```
wwqrd: Might be an interesting place to try out something like the animejs library? This makes the dragged component take up space again right?

src/containers/Elements/NodeProviderPanels.js
```diff
@@ -45,7 +44,7 @@ const getPanelConfigs = createSelector(
 );
 
 const getHighlight = (provider, panelNumber) => {
-  if (provider.highlight) { return provider.highlight; }
+  if (provider.highlight) { return colorDictionary[provider.highlight]; }
```
wwqrd: What happens when `colorDictionary[provider.highlight]` isn't defined?

jthrilly: Should return `colorDictionary['panel-highlight']` in that case. It doesn't exist right now, and in fact there are no color dictionary items relating to panel highlight colors. These should be structured exactly the same as the node colors, with panel-highlight, panel-highlight--alt-1, panel-highlight--alt2, etc.

src/ducks/modules/network.js
```diff
@@ -39,16 +40,54 @@ function edgeExists(edges, edge) {
...
-      const node = { uid, ...action.node, id };
       return {
         ...state,
-        nodes: [...state.nodes, node],
+        nodes: getNodesWithAdd(state.nodes, action.node),
+      };
+    }
+    case ADD_OR_UPDATE_NODE: {
+      let nodes = state.nodes;
+      if (state.nodes.filter(node => node.uid === action.node.uid).length > 0) {
```
wwqrd: I'd probably use `find` for this, though we probably won't see any performance improvements for that :)

If you use the lodash variant there's a shorthand version you could use too:
`find(state.nodes, ['uid', action.node.uid])`

src/behaviours/draggable.js
rebeccamadsen: This is trying to keep track of which stage and prompt the draggable was created/originated on. This is to help distinguish between where nodes were created, so only nodes created on other prompts and pulled in from the side panel can be dragged back to the side panel. Nodes created on the current prompt need to be deleted, not dragged to a side panel.

src/behaviours/droppable.js
```diff
@@ -35,6 +35,12 @@ export default function droppable(WrappedComponent) {
       this.updateZone();
     }
 
+    componentDidUpdate(prevProps) {
```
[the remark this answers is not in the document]
rebeccamadsen: This relates to trying to toggle `acceptsDraggableType` in `NodeProvider`; changing the prop didn't update the zone without this.

src/components/Elements/DropZone.js
```diff
@@ -1,23 +1,28 @@
...
 import { droppable } from '../../behaviours';
 
 /**
   * Renders a container onto which a `draggable` can be dropped.
   */
-const DropZone = ({ children }) => (
+const DropZone = ({ children, hover }) => (
   <div className="drop-zone">
-    { children }
+    {React.Children.map(
```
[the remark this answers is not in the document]
rebeccamadsen: Makes sense. I think I could do that.

src/components/Elements/NodeList.js
```diff
@@ -10,69 +12,101 @@ const EnhancedNode = draggable(selectable(Node));
...
-  draggableType,
-  hover,
-}) => {
-  const classNames = cx(
-    'node-list',
-    { 'node-list--hover': hover },
-  );
+class NodeList extends Component {
+  componentDidMount() {
+    this.props.showPanelAlways(this.props.isDragging);
```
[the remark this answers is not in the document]
rebeccamadsen: Makes sense.

src/containers/Elements/NodeProvider.js
```diff
@@ -95,20 +124,28 @@ function makeMapStateToProps() {
   const getProviderNodes = makeGetProviderNodes();
 
   return function mapStateToProps(state, props) {
-    const interaction = (props.selectable && 'selectable') || (props.draggable && 'draggable') || 'none';
+    const currentIds = { promptId: props.prompt.id, stageId: props.stage.id };
+    const isOrigin = isEqual(state.draggable.draggingFromIds, currentIds);
```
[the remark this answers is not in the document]
rebeccamadsen: Um, I'll have to look at the `hit` calculator to answer that. I'm not sure I understand how that works fully.

src/behaviours/draggable.js
rebeccamadsen: I'm not familiar with the animejs library, but yes, this makes a dragged component take up space.

src/containers/Elements/NodeProvider.js
rebeccamadsen: So this could be done differently, I'm sure. I toggled the type only because you may have one node that can be dragged to a side panel, and another right next to it that cannot. It seemed clunky to add and remove an entire zone based on which node is being dragged. Maybe there should be a new field though or something to toggle whether to allow a node to be dropped or not.

src/containers/Elements/NodeProviderPanels.js
```diff
@@ -74,30 +74,51 @@ class NodeProviderPanels extends PureComponent {
...
+    map(this.state.panelsOpen, (value, index) => this.isPanelOpen(index)),
     open => open === true,
   );
 
   render() {
     const { panels, stage, prompt } = this.props;
 
     return (
-      <Panels minimise={!this.anyPanelOpen()}>
+      <Panels minimise={!this.anyPanelAlwaysOpen() && !this.anyPanelOpen()}>
```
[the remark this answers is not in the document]
rebeccamadsen: I think this needs a better name... This is an attempt to allow a side panel to show up, even if it is empty, so that you could drop a node from the main panel on it if you don't want it in the main panel anymore. There may be a case that the side panel(s) are empty, and should hide in general, but should show up as a droppable area if a valid node starts being dragged. That is what I'm trying to specify here -- whether this is an area that should be visible when nodes are being dragged, even if it is usually hiding.

src/containers/Elements/NodeProviderPanels.js
```diff
@@ -74,30 +74,51 @@ class NodeProviderPanels extends PureComponent {
...
         { panels.map((panel, panelIndex) => (
           <Panel
             title={panel.title}
             key={panelIndex}
-            minimise={!this.isPanelOpen(panelIndex)}
+            minimise={!this.isPanelAlwaysOpen(panelIndex) && !this.isPanelOpen(panelIndex)}
```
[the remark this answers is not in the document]
rebeccamadsen: Same explanation about this being slightly different and needing a better name -- whether this is an area that should be visible when nodes are being dragged, even if it is usually hiding.

src/containers/Elements/NodeProviderPanels.js
rebeccamadsen: Makes sense.

src/selectors/interface.js
rebeccamadsen: This related to using the attributes as a match for which nodes belonged. Since you could drag nodes from other prompts into the main panel, I was using the attributes instead of promptid.

src/behaviours/draggable.js
wwqrd: That should probably be handled elsewhere, since any component (not just nodes) can be a draggable(). Would a uid enable us to do that, or do we need to have a think about how this might work?

src/containers/Elements/NodeProviderPanels.js
rebeccamadsen: Right now, if you put some invalid name in, it depends on the component to have a default, I think. So nodes default to the "neon-coral" for example if you put a bad value in.

I think panel highlights are leaning on edge-alt-[num] palette definitions right now.

src/ducks/modules/network.js
rebeccamadsen: Sounds good.

src/containers/Elements/NodeProvider.js
wwqrd: Are we talking about hover functionality here? Dropping a node on a droppable zone doesn't do anything in itself, so this shouldn't cause any issues with still accepting a type even if it's not wanted (the dropping component itself can just do a noop).

src/containers/Elements/NodeProvider.js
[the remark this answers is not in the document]
wwqrd: It loops through the registered areas and returns a list of matches, I think there's a little bit of meta data provided so that the draggable can decide what to do with the event, it might be possible to add this origin information there, but I'm not sure.

src/containers/Elements/NodeProvider.js
rebeccamadsen: I hadn't considered a noop. I think the hover functionality can be handled elsewhere; I was just trying to get the active zones to be 0. I'll look and see if the dropping component can figure out when to do a noop.

src/behaviours/draggable.js
rebeccamadsen: What would a uid change? Wouldn't it still need to be connected to the promptid/stageid somewhere along the lines, whether it comes in as a prop or not?

src/containers/Elements/NodeProvider.js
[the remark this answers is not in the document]
rebeccamadsen: I'll take a closer look at it then and see what I can learn.

src/containers/Elements/NodeProviderPanels.js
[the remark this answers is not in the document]
wwqrd: Makes sense, but what I mean is does it actually need to distinguish the two kinds of open like it does here?

```jsx
isPanelOpen(index) => { panelHasNodes(index) || panelIsDragging(index); }
isAnyPanelOpen() => panels.map(isPanelOpen);

...

<Panels minimise={!this.isAnyPanelOpen()}>
```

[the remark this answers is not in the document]
rebeccamadsen: Ah, yeah, that makes sense. :)

src/behaviours/draggable.js
wwqrd: Oh! I think I'm missing something here?

Can't the draggable make this distinction in it's onDrop method? as the originator of the Node, doesn't it already have this information about what got dropped?

If that's not the case, maybe we should think about adding some kind of `meta` prop which is associated with draggable/droppables that can contain arbitrary data.

rebeccamadsen: I'll have to think about that. I know there are three places we need this information -- we need to know whether to show the droppable area when we start dragging, we need to know whether to show a placeholder node when hovering, and we need to know whether to allow a node to be dropped. And it varies by combination of node and prompt, because nodes could come from different prompts originally.

wwqrd:
Looking good!

I'm still looking at NodeProvider/NodeProviderPanels and thinking this needs refactoring, what do you think?

Maybe NodeProvider could be a HOC that wraps around a node list. And we could have a NodePanel that handles interactions?

PR #260, opened 2017-11-20
Fixed stray comma in webpack.config.dev.js

beckah:
Removed stray comma on line 80 which prevented the ability to run dev environment locally.


jthrilly:
Hey! Thanks for the PR. 

In this case, the comma is actually correct - it is known as a 'dangling comma', and is part of our linting rules (which we inherited from airbnb).

I think the issue you might be having is that this is a relatively new language feature, and you will need to be using the correct node version for it to work. The project specifies version 8.5.9, I think - can you check you got the error with this version? If you did, I will look into this further 👍

Issue #174, opened 2017-06-05
Multiple `Elements` folders are confusing and awkward to maintain.


wwqrd:
I want to have a rethink of the way our components are organised.

We’re using ducks/containers/components, which seems great.

However, inside both `containers` and `components` we have `Elements`, which is intended to contain components for direct use in interfaces. I am now thinking is slightly confusing due to it being split across both folders.

What I’d like to do is have a top-level (on the same level as containers and components), which contains *all* our interface elements (container or otherwise).

Are people cool with me making this change - any suggestions on what the folder should be called?

rebeccamadsen:
So, are you referring to the "interfaces" with a capital "I" like NameGenerator, or interfaces as in user interfaces (e.g. buttons, menus, checkboxes)? I think I had just assumed Elements was for...smaller functions...or helpers...or something. It sounds like I was wrong, so a more clear name would be helpful, for sure.

Some of the files in both Elements folders access state. If you combine them, would we be more explicit in the file name whether it is a container or component? Or maybe have a container/component division within the upper level folder? 

Could it make sense to go the other direction and break things down into smaller folders like Forms, Navigation, Nodes, etc?

wwqrd:
Yes, these Elements are intended for *direct* use in the Interfaces.

I was kind of thinking from the point of view of a contributor making an interface out of our components, without necessarily needing any knowledge of how our state is structured. They could look in a single place for all the components (and documentation) they would need.

For our end-users I imagine they wouldn't be concerned (or want to be concerned with), knowing if something was a 'smart' or 'dumb' component. So I don't think *more* organisation would make it clearer, or adding Container/Component to the name of every item that is used in an Interface.

So what I'm essentially suggesting is a change from:

```js
import { PromptSwiper, NodeProviderPanels, NodeForm } from '../../containers/Elements';
import { NodeList, NodeBin } from '../../components/Elements';
```

to:

```js
import { PromptSwiper, NodeProviderPanels, NodeForm, NodeList, NodeBin } from '../../Wigets';
```

To be used like:
```jsx
const MyShinyInterface = ({ prompts, panels, activeOriginNetwork }) => (
  <div className="my-shiny-interface">
    <PromptSwiper prompts={prompts} />
    <NodeProviderPanels config={panels} />
    <NodeList
      network={activeOriginNetwork}
      label={label}
    />
    <NodeBin />
  </div>
);
```

Not like:
```jsx
[...]

PR #262, opened 2017-12-07
Feature/refactor dnd

wwqrd:
Sorry to send this your way incomplete and not working. I did some pruning in preparation to show you all and it broke! (Maybe you can help fix it? 😁 )

I've been working on the stuff in `src/behaviours/DragAndDrop`, the idea is to formalise the API (https://github.com/codaco/Network-Canvas/wiki/Plan-for-new-dnd-interfaces, heavily borrows from react-dnd), develop it in a way to support some of the features we are going to depend on, and generally tidy it up a bit from the monolithic monstrosity it was before.

# Status

~The monitors aren't actually binding props to the components, so there aren't any hover interactions at the moment (they were implemented, but broke in the tidy, should have them back tomorrow). The actual dragging should work with the state echoed to the console.~

Monitors are binding props, but component ids are changing mid-flight messing up some of the hover detection.

Hit detection is working, but drop callbacks are not.

# What's new

1. Superficially the first thing you'll notice is the DragAndDrop folder with an index. Previously we've been using the containers/components set up which is starting to bulge at the seams a little. I still think we should try and separate our components into controller and view ones (😉 ), but perhaps we need to start to think about grouping by feature at the top level.
2. The store is separated out into it's own instance of a Redux store. This is shared as a global, I did start using recompose's withContext, but probably because of my implementation, the performance was really poor. Since we will only ever have one 'Drag' context it seems to be an okay thing to do? Unless you lot have objections?
3. I've split out the binding of drag props (for observing the state of the drag context), to the bit that makes things Draggable or Droppable. The idea is you could decorate anything in the app with these props for it to get awareness about the world of draggable components.
4. the onDrop event is now on the DropTarget, rather than the DragSource, I think this will make way more intuitive sense to us all. However, it's implemented in an evil way, with a method call in the reducer, i haven't thought about how to do that yet (probably with Rx), but it's on the roadmap.

The general thinking is to calculate all the things we're interested on the state, so if we need new features this is where they would live, we'd then add the respective props to the Monitors.

Please let me know what you think!


src/behaviours/DragAndDrop/dragState.js
```diff
@@ -0,0 +1,142 @@
...
+}
+
+const store = createStore(reducer);
+
+store.subscribe(() => console.log(store.getState()));
+
+const actionCreators = {
+  updateTarget,
+  dragStart,
+  dragMove: throttle(dragMove, 100000),
```
bryfox: `100000` is probably unintentionally high...

wwqrd:
Hey all, I reverted from the cool Rx store handler to a simple listener and performance is pretty good again, just ironing out some stuff with DropTarget ids but in theory it should all be working now.

wwqrd:
This seem to be all working now!

rebeccamadsen:
>  I still think we should try and separate our components into controller and view ones (😉 ), but perhaps we need to start to think about grouping by feature at the top level.

I agree we should think about grouping by feature, but still have controller and view separation. Perhaps related to #174?

wwqrd:
@bryfox @rebeccamadsen @jthrilly Okay folks, think this is ready to merge?

bryfox:
I'm seeing some console warnings/errors after resetting session and clearing cache (Chrome desktop).

When I load the demo protocol:

```
- Failed prop type: The prop `externalData` is marked as required in `NodePanels`, but its value is `undefined`. 
  in NodePanels (at Monitor.js:34)
- Warning: Failed prop type: Invalid prop `isOver` of type `boolean` supplied to `NodeList`, expected `function`.
  in NodeList (at scrollable.js:37)
- Warning: Failed prop type: The prop `id` is marked as required in `DropTarget`, but its value is `undefined`.
  in DropTarget (created by withProps(DropTarget))
```

Occasionally, when dragging (protocol screen):

```
- setState(...): Can only update a mounted or mounting component.
  ...
  _this.trackSize (withBounds.js:48)
```

And this console.log()... happens rarely, after dropping (protocol screen); but happens repeatedly/endlessly once started without further interaction:

```
Error in accept() function TypeError: Cannot read property 'itemType' of undefined
    at accepts (NodePanels.js:79)
    at willAccept (reducer.js:18)
    at markTargetHit (reducer.js:35)
    at reducer.js:39
    at Array.map (<anonymous>)
    at markTargetHits (reducer.js:39)
    at compose.js:32
    at reducer (reducer.js:97)
    at dispatch (createStore.js:166)
    at Object.dispatch (index.js:14)
    at DropTarget._this.updateTarget (DropTarget.js:63)
    at DropTarget._this.update (DropTarget.js:48) {x: 848, y: 347, dy: 5, dx: 0, dt: NaN, …}
```

rebeccamadsen:
This is a large undertaking! I like the general architecture.

I cannot click on nodes to edit or add a new node though. It just cycles register/unregister field over and over until I kill the process.

Also I notice a couple of changes from existing behavior, which I'm not sure if they are intentional. If "Already mentioned" panel is closed because it is empty, it opens if you start to drag, even if it cannot be dropped into. "People from previous visit" used to not be droppable -- the thinking was that users would have to "trash bin" nodes they want in that panel, making the interaction different because it affects other prompts that aren't visible.

Sometimes I get an "Error in accept() function TypeError: Cannot read property 'itemType' of undefined" on the console if I drop an item into the same panel I was dragging from. But not always. I haven't found what makes that error repeatable yet.

```
Error in accept() function TypeError: Cannot read property 'itemType' of undefined
    at accepts (NodePanels.js:83)
    at willAccept (reducer.js:18)
    at markTargetHit (reducer.js:35)
    at reducer.js:39
    at Array.map (<anonymous>)
    at markTargetHits (reducer.js:39)
    at compose.js:32
    at reducer (reducer.js:97)
    at dispatch (createStore.js:166)
    at Object.dispatch (index.js:14)
    at DropTarget._this.updateTarget (DropTarget.js:63)
    at DropTarget._this.update (DropTarget.js:48) {x: 930, y: 473, dy: -3, dx: 4, dt: NaN, …}```

wwqrd:
> I cannot click on nodes to edit or add a new node though. It just cycles register/unregister field over and over until I kill the process.

@jthrilly This seemed to be caused by the dependency update.

PR #263, opened 2017-12-08
Feature/change drag drop alt

wwqrd:
Fork of #259 which refactors the node panels.


wwqrd:
Oh drat, I suppose we can't merge this with the partial feature set. Maybe we merge in 259 first!

Issue #127, opened 2017-03-08
Roster-based name generator


jthrilly:
# Description
This interface is a variant of the name generator, that allows respondents to choose from a pre-defined list of alters instead of creating them from scratch.

A typical use would be in a setting where all actors in a network are known (such as a classroom).

We have no interface mockups available for this interface, since it was conceived off after our design consultation.

There are _three_ proposed interfaces for this, which suit different network sizes and configurations. We can create three interfaces, or we can bundle this functionality into one interface with some logic for switching between them. 

## Auto complete box
This will be suitable for large rosters, where it would be impractical to show all potential network members at once.

Essentially, the new node form is replaced by an autocomplete input, which is triggered after x keystrokes and only allows options from the predefined list to be selected. It is important to note that the autocomplete text might contain place names rather than simply people names, so this matching might need to work across multiple node properties.

## Node select
This interface will be suitable for smaller rosters, where it is practical to show all potential network members at once.

The participant will be able to select any nodes on this screen, with those selected being brought into the current interview.

## Pre-populating the interview
This is not really a roster interface, but is included here for clarity. Some interview protocols will want to include an entire external network from the start. That is, they will not ask the participant to nominate any alters - all alters will automatically be included from the start. In these cases, the alters will be stored within the protocol file directly when it is loaded. 

#Specification 
To be determined. 

# Progress

- [x] Implement side panels in existing name generator.
- [x] Implement node list select interface (#265)
- [x] Implement auto complete name generator (#264)

jthrilly:
Discussed this week was the need to have roster based name generation.

This needs to support different types of actor (i.e people, venues).

michellebee:
Some thoughts from meeting with school psychology phd students: 

1. roster-based name gen should be optimized for working with schools and/or organizations.  typical scenerios will be:
 - students nominating other students
 - teachers nominating students with whom they have a particular relationship with.
 - we may also want to support observers reporting the relationships they see other children having with each other. this might bypass the namegen stage entirely and go straight to the alter layout/alter connection stages.

2. we should have functionality for both name autocomplete and a pull down list for those that need/require it.

3. it will be important to be able to sort rosters by particular variables - for school research, these variables will likely include: name of school; grade; last name of teacher or some other identifier to find the classroom. 
- we will likely want to autosort rosters by data input by the researcher or interviewee in the stage0/intro screen.

pfsj:
[...]

PR #267, opened 2017-12-18
Autocomplete roster generator

### LINKED ISSUE #264: Roster Name Generator - Auto complete
# Purpose
There are situations where we want to limit the scope of new node creation to a predefined list. This is typically the case when when we know the boundaries of the network we are interested in, as for example in a classroom, or organizational setting.

Depending on the number of nodes the user is able to choose between, different interfaces are appropriate.

The autocomplete interface is designed for situations where there are too many possible nodes to display either in a side panel, or a node list.

It is important to remember that nodes might not always represent "people". They could represent other actors, such as places, or even medical conditions.

# Functionality
This interface should:
- Use the configuration object described below to accept parameters.
- Accept data from the protocol file in the format of a network
- Parse this network data and understand the properties of the nodes, using the variable registry where appropriate.
- Provide an autocomplete box that allows fuzzy matching of node attributes based on configuration parameters
 
# Configuration

``` Javascript
    {
      "id": "namegenroster1",
      "type": "NameGeneratorAutoComplete",
      "label": "NG Classmates",
      "prompts": [
        {
          "id": "7cl",
          "text": "Which classmates would you also think of as friends?",
          "additionalAttributes": { // Added to the node object when it is added to the main interview network
            "special_category": 46,
            "close_friend": true
          },
          "dataSource": "schoolPupils", // Key corresponding to external data section of the protocol file.
          "autoCompleteOptions": { // Configurable per prompt, but with sane defaults TBD.
            "fuzzyness": 5, // Can we control the level of fuzzy matching?
            "matchProperties": ['nickname', 'name'], // Properties to apply matching to
            "sortOrder": {
              "nickname": "DESC"
            },
          }
        },
        {
          "id": "7su",
          "text": "Which of these venues have you visited in the previous 6 months?",
          "additionalAttributes": {
            "primary_venue": true
          },
          "dataSource": "HIVServices",
          "autoCompleteOptions": { // Configurable per prompt, but with sane defaults TBD.
            "fuzzyness": 5, // Can we control the level of fuzzy matching?
            "matchProperties": ['Raw Address', 'Abbreviated Name'], // Properties to apply matching to
            "sortOrder": {
              "Abbreviated Name": "DESC"
            },
          }
        },
      ]
    },
```

# Sample Data

https://gist.github.com/jthrilly/3c44635576d423d58710b4b3c72542b2

# UI Mockups

Incoming.

# Sub-tasks

### LINKED ISSUE #269: Implement "Add Count" button
Two of the roster interfaces need a submit button which reflects the number of items to add.

Mockup:

addcountbutton

Requirements:

- Display dynamic count up to a reasonable limit (999?)
- Dynamic background color based on node type

bryfox:
This adds a new interface that implements a "fuzzy search" roster builder as described in #264. There are several known issues (noted below) but I'd like to get some eyes on this before proceeding too much more.

## Notes

By design, any nodes that have been added to a name roster are excluded from search results. These exclusions are not stored directly; rather, all displayed fields of a search result are used to define equality between a search result object and a node. I've assumed this is sufficient for uniqueness; otherwise a user would have no way to distinguish. But there may be a better way to handle this. (See Search.js: `isAllowedResult`.)

## WIP shortlist

- [x] Support node-specific background colors in AddCountButton
- [x] Customize the initial 'add [person]' button to node type
- [x] Rework Search form (redux-form not needed)
  - [x] `<input type=search>`
- [x] support node deletion from interface
- [x] stylized input
- [x] Refactor network reducer: see addNodes() comment
- [x] "AddCountButton" svg missing its gradient shape
- [x] adapting to screen size changes
- [x] integrate with Card/CardList
- [x] scss duplication
- [x] investigate data change on prompt switch ("undefined" labels)
  - Caching added. See #305 for further discussion.
- [x] prevent node duplication (uid discussion below)

## Known issues

- <strike>Human-readable palettes. See latest comment in #269. I had implemented a NodePalette util to handle the translation, but I can make protocol use the 'unfriendly' versions if needed short-term.</strike> Colors are read from the protocol's variable registry.
- Need better animation implementation (e.g., search box should expand from button, not slide in)
- I added a 'fix' for search results display when a soft keyboard scrolls the window out of view. I can look into more resilient solutions along with the animation changes.

## Feedback requested: SearchResults

<strike>My original intention was to keep searchResults state out of redux as the larger application context shouldn't/doesn't need to know anything about it. (Next step would be to move those into the SearchForm component.) However, I may be fighting reduxForm a bit here — resetting value is not working as expected. I'd like input on this part of the design. Do these results belong in redux state?</strike> Feedback is below. Thanks.

Resolves #264. Resolves #269.


src/other/demo.canvas/protocol.json
```diff
@@ -233,6 +469,51 @@
     }
   },
   "stages": [
+    {
+      "id": "namegenroster1",
+      "type": "NameGeneratorAutoComplete",
+      "label": "NG Classmates",
+      "form": "autocompletingSearchForm",
```
jthrilly: I think we want to rethink the way that this interface handles its relationship to the form in its config. 

In the standard name generator, we want form to be passed because it could contain an arbitrary redux form, that our Architect app will let users create. In the case of the autocomplete, there doesn't seem to be any customisation possible with this form.

Can we hardwire this in, for now, and remove the config property?

bryfox: Oops, this is a vestige of my original prototype. I removed the protocol-driven form based on conversations with @wwqrd ; just need to remove this reference.

wwqrd:
Made an error in my code example, I meant:
```
// <Search> is a <Field> compatible component which renders as a search box AND the results, and sets "value" to the array of results.
const AutoCompleteForm = () => (
  <Form>
    <Field component={Search} /> 
    <input type="submit" />
  </Form>
);
```

bryfox:
@wwqrd 

- network reducer - that looks good to me.
- thanks for the redux form links; I'll take a look
- form structure (`component={Search}`). Interesting! Looks like I have `initialValues` and `enableReinitialize` [at my disposal](https://redux-form.com/7.2.0/docs/api/reduxform.md/). Let me work on this a bit. Conceptually, I liked that the search container encompassed a form, since that described the dependency to me, but I think your approach might solve my problems... and I can rename Search to SearchField or something.

bryfox:
@jthrilly thanks for your button comments — I think those all together give me what I need. And I think we have a (better) plan for the search form now.

bryfox:
@wwqrd @jthrilly "Autocomplete" naming: now is probably a good time to change the name if it's confusing (and I agree it more resembles a filter). 

The [original issue](https://github.com/codaco/Network-Canvas/issues/127) mentioned potentially having a drop-down as well, which I could imagine turning into a [dropdown with autocomplete](https://harvesthq.github.io/chosen/)... would another name be better here?

wwqrd:
@bryfox 😆 It could be neat, but maybe it's a bit of a hack? I think maybe your approach makes more sense conceptually. It is really a self contained "view", it's not like this would be a field amongst other fields in a form. Given that, I'd be tempted not to even use Redux Form. @jthrilly perhaps you could clarify long term intentions for this component?

jthrilly:
> @bryfox 😆 It could be neat, but maybe it's a bit of a hack? I think maybe your approach makes more sense conceptually. It is really a self contained "view", it's not like this would be a field amongst other fields in a form. Given that, I'd be tempted not to even use Redux Form. @jthrilly perhaps you could clarify long term intentions for this component?

So as I said, I think not using redux-form for this is the way to go. There is no intended further use for this component outside of this interface at present.

> @wwqrd @jthrilly "Autocomplete" naming: now is probably a good time to change the name if it's confusing (and I agree it more resembles a filter).

Programmatically, filter is more accurate. However, the team (and our users) will probably not understand that naming. 

Perhaps we should take a different approach, and name these new interfaces after their intended design functions. They are both name generators, both roster-based, but differentiated by the size of the external data they are designed to handle. Perhaps `NameGeneratorRosterLarge` and `NameGeneratorRosterSmall` would be better? @rebeccamadsen? @wwqrd?

Feel free to name the actual component something to do with filter if that makes more sense to you.

bryfox:
@wwqrd see 0866bac for revisions to the network actions.

Based on your feedback, I consolidated `addNode` and `addBatch`, renaming to "addNodes". I left the updating version alone for now; I don't feel like I understand the big picture well enough to change the behavior in that way. We can discuss that further here, or address in a follow-up issue — I'd prefer the latter.

bryfox:
re: organizing `<Search />` like `./ConcentricCircles/`. In general, I like the pattern.

I need to take a closer look at what's happening over in #273, but my hunch is the `<SearchResults>` could be replaced by the  `CardList` with some minor changes. (And `<SearchResult>` will be replaced with `<Card>`).

I had left a TODO for moving search results management into `<SearchResult>` but since we're not intending to use that functionality elsewhere, my inclination is to leave the result(s)/card(s) as shared pure components. If so, then `<Search>` would stand alone.

I'll resolve one way or the other.

bryfox:
This is ready for review again; I've removed the WIP label. Note that the 'autocomplete' demo is now the final stage in the protocol.

Since last time, this:
- has Search containers/components grouped together
- integrates with Card/CardList
- Fixed some behavioral problems
- adapts better to screen size changes
- removes some scss duplication
- add some basic Search component tests

There are still a couple of known issues in the description, but (pending feedback) I'm OK following up with a separate PR to address those.

I may need to make a couple of tweaks once #293 is merged, but nothing too significant.

jthrilly:
Great work Bryan. Very close indeed, and might even be mergable as-is depending on if others uncover anything. 

Here are a couple of thoughts:

- I added some styling and animation to the card component on Rebecca's branch, that you might want to test works ok here. Nothing too dramatic. She also did some work with making the card grid more responsive, which we might want to look at, but to be honest I think we should actually just reduce the overall size of the card container/search results UI in your interface, and keep it as one full width card per line. We can handle this in the final CSS tweaks before merging.  
- When switching prompts, the node labels change to 'undefined' briefly as they animate out. Obviously that's not great aesthetically, but I also worry that if this property is changing we might be impacting animation performance, so I want to understand if this is potentially happening elsewhere, too. Can you shed any light on it so we know if we should investigate other node lists?
- Great job implementing the node color for the count button inside the search results. I noticed the nodes in the node list aren't displaying the correct colors, but that's probably because we didn't implement that feature yet. I imagine it would be a small tweak to node list. Do you think you could make it here, since you are ahead of the other branches in having implemented colors?

src/utils/getNodeIconName.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Simple resolver for icon naming.
+ *
+ * Reconcile protocol "venue" naming with Network-Canvas-UI (svg) "place"
```
rebeccamadsen: Can we just update the svg in Network-Canvas-UI if we prefer "venue", or is there a reason for having different naming schemes?

bryfox: I'd personally prefer to use one or the other... I'll open an issue on UI if others agree that 'venue' is better.

jthrilly: Hm. so something has gone partially awry here. 

So, the node types (and colors) are supposed to have one canonical point of definition: the variable registry section of the protocol file:

``` JSON
"variableRegistry": {
  "node": {
    "person": {
      "label": "Person",
      "color": "coral",
      "displayVariable": "nickname",
      "variables": { ... }
    },
  },
}
```
As you can see, this defines both the node type and the corresponding node color, with reference to the existing palette colors (which the node component itself should accept as a parameter: `<Node color="coral" />`). Where no color is defined, we don't have specified behavior yet. A given prompt on an interface will refer to a node type, and the interface should look up the corresponding meta-data (such as node color) here.

To handle the icons, we should introduce an additional property here, something like `iconVariant`. Individual interfaces can then use `getNodeIconName.js` to resolve this. The Name Generator and other interfaces that show a new node button will need to have this functionality added.

I might have misunderstood your discussion, so sorry if I've just confused matters.

bryfox: - [x] look up color in registry
- [x] look up iconVariant in registry

bryfox:
@rebeccamadsen Can you elaborate on this?

> If you add a second prompt in this interface with the same data, it is possible to end up with two nodes in the network with the same uid.

The Search result should be excluding any results that 'match' nodes (based on the definition of uniqueness we discussed elsewhere), so I'd like to understand how this was happening.

---

I did just push an update to the `uid` generation — rather than having SearchResults creating those on nodes (which were getting persisted), I've added an optional uid mapper to CardList, which defaults to reading the `uid` prop.

Are you still seeing the above issue?

bryfox:
[x] Updated card styling to use a single column (in search results)

bryfox:
> the node labels change to 'undefined' briefly

The difference between this & the primary name generator is that this one drives the label from a `displayLabel` variable in the protocol, which changes with the prompt. I'll need to investigate further to see how to best handle this — possibly caching values on the displayed nodes.

> might be impacting animation performance

I don't think this is the case at first glance; the state changes are run between timeouts during which time the animations run.

rebeccamadsen:
The duplicate nodes does not happen on one prompt. You have to create two prompts that both use the same data set to see this. So, steps to see it are: create a second prompt with the same dataset; add a node on the first prompt; add that same node on the other dataset. Then in the network on the console (or on other stages that show network nodes) you see two nodes that are nearly identical, except for additional attributes and promptid. Does that help?

jthrilly:
Sorry Bryan, I wrote the following out first thing but apparently never hit enter. Might be redundant now depending on what you've done since then.

> The difference between this & the primary name generator is that this one drives the label from a displayLabel variable in the protocol, which changes with the prompt.

Good to know. That's also interesting that the name generator (and presumably also the sociogram) do not correctly derive the label from the variable registry, yet. Does the way you implemented this generalize without much trouble? I am thinking of quickly adding this functionality to the other interfaces, too.

>I'll need to investigate further to see how to best handle this — possibly caching values on the displayed nodes.

@wwqrd  - any suggestions?

> I don't think this is the case at first glance; the state changes are run between timeouts during which time the animations run.

My worry was it looked sort of like: animation starts -> props change before animation completes -> animation ends. As far as I knew, updating the text content of an element would require recalculating styles. I didn't check in the profiler though, so you might well be right.

bryfox:
@rebeccamadsen thanks catching the 'duplicate' issue. Based on your insights in #304, for now I'm going to exclude network nodes from search results. If we want to use subsequent prompts to update node attributes, then I can address in the future.

bryfox:
@jthrilly I think display label was a red herring; the promptId was duplicated in the demo protocol, which was the real problem.

src/selectors/protocol.js
```diff
@@ -2,7 +2,26 @@
...
+  state => state.protocol.forms,
+  forms => forms,
+);
+
 export const getExternalData = createDeepEqualSelector(
   state => state.protocol.externalData,
   protocolData => protocolData,
 );
+
+export const getNodeColor = createDeepEqualSelector(
```
wwqrd: Any selectors using props need to be function factories to cache the results properly. Otherwise all instances of the components will share the same cache. e.g.

```js
export const makeGetNodeColor = () =>
  createDeepEqualSelector(
    protocolRegistry,
    (state, props) => props.nodeType,
    ({ node }, nodeType) => node && node[nodeType] && node[nodeType].color,
  );
```

src/selectors/search.js
```diff
@@ -0,0 +1,33 @@
...
+import { getExternalData } from './protocol';
+
+const getDatasourceKey = (_, props) => props.dataSourceKey;
+const getSearchOpts = (_, props) => props.options;
+
+/**
+ * Look up the search set for a search based on a data source in the protocol.
+ * The value of this key should point to an attribute in the protocol's externalData.
+ */
+export const getSearchData = createSelector(
```
wwqrd: Should be factory method:

```js
makeGetSearchData = () =>
   createSelector(
      ...
```

src/selectors/name-generator.js
```diff
@@ -78,3 +84,12 @@ export const getDataByPrompt = createSelector(
   (externalData, key) => externalData[key].nodes.map(
     (node, index) => ({ uid: nextUid(externalData[key].nodes, index), ...node })),
 );
+
+export const getNodeIconName = createSelector(
```
wwqrd: I think this one also needs to be a factory method?

src/ducks/modules/rootReducer.js
```diff
@@ -10,6 +10,7 @@ import draggable from './draggable';
 import droppable from './droppable';
 import modals from './modals';
 import menu from './menu';
+import search from './search';
```
wwqrd: Is this module defunct now?

src/selectors/rehydrate.js
```diff
@@ -1,14 +1,7 @@
 /* eslint-disable import/prefer-default-export */
```
wwqrd: Without the other methods I think this might make more sense called something like ` src/selectors/forms.js`?

src/selectors/search.js
bryfox: OK. Can you explain why this one needs the factory? The underlying memoize & recalculate seems to be working as I'd expect.

src/ducks/modules/rootReducer.js
bryfox: It is being used, though arguably we don't really need to keep the toggle state of the search panel.

bryfox:
> I'm not keen on the fix for display labels, do we know why it's returning undefined?

Agreed; see #305.  It's returning undefined because the display attr is driven by props. But I think there's some deeper problem with state/rendering there which I haven't understood.

src/selectors/search.js
bryfox: Nevermind; I see based on your above comments. Thanks!

bryfox:
@wwqrd I pushed fixes for your comments above — thank you.

- I left getSearchData alone, but stopped exporting it. "fuse" is the factory we'll be using. Others are now exporting factories as noted.
- I left the search state reducers in for now.

bryfox:
I've pushed the latest: fixed conflicts and updated protocol to match that described in #300. (I kept "cardOptions" but renamed "fuzzyness"->"fuzziness").

bryfox:
Latest version has "v2" animations; something fairly close to the wireframes.

I've also bumped the version of sass-lint because of an error (and, unintentionally, the UI dependency in package-lock).

PR #270, opened 2017-12-20
Fix/node version

jthrilly:
I thought I would just bump the node and npm versions for the project (given the package.json bug), and switch to `nvm` for handling it, since I think this is what most of us use already.

You should be able to just do `nvm use` now from within the project directory, and have it pick up the correct version. You can integrate this with your shell to have it switch automatically, but I'll leave this up to you. 

The `.nvmrc` currently specifies the latest LTS release, which I think is a good choice for stability.


jthrilly:
@codaco/network-canvas-team - any objections to this approach?

rebeccamadsen:
Being a windows user, I don't use nvm, but I'm okay with this approach.

wwqrd:
This doesn't work with the tool I use manage node versions either. Suggest adding both these files to `.gitignore` and referring to the `package.json` instead?

jthrilly:
Which tool do you use, Steve?

wwqrd:
I use nodenv because it parallels rbenv which is the ecosystem I migrated from (that's probably where the `.node-version` comes from).

If we went ahead with this I could implement a workaround for my system.

But since all of our systems are different it might make sense to use a standard? The closest thing to a standard on these things seems to be `engines` in `package.json`, which unfortunately isn't used by any of the tools...

jthrilly:
This is a really weak point of the current ecosystem by the looks of things. No cross-platform version manager, and no single point of truth for a project's node or npm versions. I have added a small bit to the project readme to try and tie this together a bit, but we will need to update three files for the time being whenever we want to change the node version.

PR #278, opened 2017-12-21
Node panels only shows if panels accept the node type.

### LINKED ISSUE #275: Minor change to panel show/hide behavior and node drag
On the Name generator interface, all panels expand when a node drag starts. This is well intentioned: the panels may contain a drop target for the node, and expanding them allows the user to use these drop zones.

However, this behavior is also confusing. It expands panels that do not contain a valid drop zone for the node, but the expansion implies to the user that the panel that has expanded is somehow relevant to the interaction.

The ideal solution would be to only have the panel expand if it contains a valid drop target for the currently dragging node. However, this would be extremely difficult to implement, since it would require the panel itself to know about the drop target it contains, and the node that is being dragged. 

I think for now a better solution is to not trigger the panel expansion at all. If the user has chosen to hide a panel, it will be because it is not useful for them in a given context. Panel hiding and expansion resets between screens as the interview progresses, so this is not a huge issue. We can revisit if we get further feedback.

wwqrd:
### Known issues:

- Both panels can sometimes be valid, is that confusing?:
    1. Node originally in 'previous' panel, dragged into main list
    1. On a different prompt that node now shows in 'existing' panel, and again can be dragged to the main list
    1. If that node is now picked up from the main list it can be a) dropped into existing to de-select prompt attributes b) be dropped into 'previous' to be deleted.

Fixes #275

### DIFF
--- a/src/containers/Elements/NodePanels.js
+++ b/src/containers/Elements/NodePanels.js
@@ -39,10 +39,12 @@ class NodePanels extends PureComponent {
     newNodeAttributes: PropTypes.object.isRequired,
     isDragging: PropTypes.bool,
     panels: PropTypes.array,
+    meta: PropTypes.object,
   };
 
   static defaultProps = {
     panels: [],
+    meta: {},
     isDragging: false,
   };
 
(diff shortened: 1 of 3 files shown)


jthrilly:
Just to make sure: if a node has been created on the current propmpt, you shouldn't be able to drag it into 'existing' - you should only be able to delete it from that prompt. Is that working?

wwqrd:
- If node was authored on a prompt i can only be deleted (Whilst on that prompt)
- If a node was created elsewhere but had the current prompts attributes added, it can be dragged back to the panel to remove them
- If a node originally came from 'previous' nodes, it can be dragged back there from any stage to reappear in that panel (effectively deleting it from the network)

Issue #236, opened 2017-09-05
Upgrade to v2 spec of react transition group.


wwqrd:
# Purpose

There are certain limitations to the old react transition spec which effect how animations work in our app. For example, only one kind of animation is allowed within each group. In `<NodeList>` whereas ideally we might animate the "ghost" new node instantly, and the initial nodes staggered, we have the same animation for both, which is a bit jarring.

https://github.com/reactjs/react-transition-group

jthrilly:
Staggered animation for node entry could look like this: https://examples.framerjs.com/#animation-staggered.framer

jthrilly:
Hi fidelity mockup of staggered node animation:

![panel animation](https://user-images.githubusercontent.com/1387940/34522221-890f36aa-f089-11e7-9f7f-f918dd2b22bb.gif)

PR #280, opened 2017-12-21
Updated packages for react 16

### LINKED ISSUE #274: Upgrade to React 16
React 16 includes many useful features, as well as speed improvements. 

Perhaps most importantly, it changes the license of the code (removing the controversial patent clause), which makes the future of Network Canvas development more secure. Because of this, we want to switch to React 16 as soon as possible. 

It is difficult to estimate the work involved in this change, but given we have no deprecation warnings in master, I am hopeful it won't be a nightmare experience. 

Please feel free to add notes about compatibility if anything occurs to you.

wwqrd:
Known issues:
- Warning about the version of react-transition-group we're using. It compiles and seems to work fine visually, so suggest we leave upgrading this package as part of #236

Resolves #274


wwqrd:
This went eerily smoothly for me. Would be great if people could try this out on their machines before I merge it. Ideally with a fresh install of `node_modules`.

jthrilly:
I'll give this a go.

wwqrd:
I think I've fixed the field registration loop.

Dynamic validate function causes infinite re-renders/re-registration:
https://github.com/erikras/redux-form/issues/2412#issuecomment-322431972

jthrilly:
Works perfectly for me. I've tested all fundamental interactions. Only issue I've observed is a bug with the sociogram interface where creating an edge doesn't turn off the selected state of the two nodes involved, but that is probably separate issue.

I also tested on the newer node and npm versions from the PR related to that, and everything was good. 

I say merge.

PR #287, opened 2018-01-03
iOS panel animations

### LINKED ISSUE #282: Name Generator: panels do not animate their collapsed/open state on safari and iOS
As the title says, the side panels on the name generator do not animate their collapsed and open state on iOS or Safari. 

I haven't investigated, but it is hopefully just a CSS bug we can work around.

wwqrd:
iOS/Safari can't animate the flex properties we're using in the layout, so this is a work around using `max-width/max-height`.

Known issues:
- Because we animate from an arbitrary max-height, the border animation/margin animations appear out of sync. What's actually happening is that the max height is being transitioned between that higher arbitrary figure and the actual height, so visually nothing changes:

    ```
    start                                             end
    |A--------------------------B----------------------C| max-height
    |D-------------------------------------------------E| border-width
    ```

    A: 1000px
    B: the actual displayed height of the panel, say 400px, at this point the animation appears to start.
    C: 0px

Another way to do it would be to set these figures dynamically using getBoundingRect/animate.js if we would prefer a more precise look.
 
Resolves #282

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -39,9 +39,9 @@
     <preference name="Orientation" value="landscape" />
     <preference name="target-device" value="tablet" />
     <preference name="DisallowOverscroll" value="true" />
-    <engine name="android" spec="^6.2.3" />
-    <engine name="ios" spec="^4.5.0" />
     <plugin name="cordova-plugin-app-version" spec="^0.1.9" />
     <plugin name="cordova-plugin-file" spec="^4.3.3" />
     <plugin name="cordova-plugin-x-socialsharing" spec="^5.2.0" />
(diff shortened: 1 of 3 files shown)


rebeccamadsen:
I can collapse/expand the side panels, but they are not animated in Chrome. I can't test the ios though.

jthrilly:
Relevant webkit bug: https://bugs.webkit.org/show_bug.cgi?id=180435

wwqrd:
So we're going to leave this as it was, not animated on iOS. It's supported on other platforms and should come to iOS eventually, as it's in the spec for flexbox.

PR #290, opened 2018-01-04
Update electron scripts to use project version of electron

jthrilly:
I just encountered an issue where I was using a globally installed copy of electron, rather than the one the project specifies. They were different versions, and that could get pretty bad after a while.

AFAIK, pointing to the project's copy in `node_modules/.bin/` is safe, but someone else might know better.


wwqrd:
What command are you using electron with? npm adds node_modules/.bin/ to the PATH environment variable so this shouldn't be necessary (though I'm not sure which has higher precedence).

jthrilly:
You're absolutely correct. I was assuming that the npm script was literally just running `electron ./www` without changing the path, but it does add `.bin`, and at a higher precedence. No need for this change.

Issue #285, opened 2018-01-02
Drag and drop: delay in dropped node appearing


jthrilly:
There is a significant (1+ second) delay between a node being dropped and it appearing in its new location. This delay also appears when a node being dragged leaves a drop zone, but this time it affects the time it takes for the "ghost node" (placeholder node) to disappear. It would be great if we could reduce this or eliminate it entirely.

On a related note, the 'ghost' node always appears at the end of the node list, but the node list is sorted alphabetically by nickname (and I believe will soon have a sort parameter - @rebeccamadsen?). This means that once the node is dropped, it likely won't appear where the ghost node is positioned. There are several possible solutions to this, but I think the simplest is to remove the placeholder node entirely, and rely on the existing visual indicators of a valid drop target (which we can make more explicit if we need to).

wwqrd:
This is a declared with `<StaggeredTransitionGroup />` in `<NodeList />` https://github.com/codaco/Network-Canvas/blob/master/src/components/Elements/NodeList.js#L42,L48

We could probably refine that to play more nicely with adding/removing nodes but I'd suggest we bundle this issue with https://github.com/codaco/Network-Canvas/issues/236

Issue #283, opened 2018-01-02
Name generator: staggered animations do not work correctly when switching between prompts


jthrilly:
Changing prompt on the name generator should trigger animation on all node lists where the contents have changed. This includes the main node list, and the side panels.

The animation is defined as: (replaced by mockup in next post)

~~All nodes in the container simultaneously animate out, with a scale(0) and opacity(0) of duration medium.~~

~~The changed nodes then animate in, in a staggered fashion, using the reverse animation but of duration fast.~~

Currently (speaking purely from a visual perspective), the main node list implements the "in" animation somewhat correctly, but with inconsistent time gaps between incoming nodes, and some visual glitches (nodes appearing instantly, or flashing in and then out). It does not seem to implement the out animation.

The side panels seem to show a brief "in" animation the first time they are shown, but do not animate after that. I can't be positive, but I believe the contents of the side panel are not updating correctly (will investigate this further and write up a separate bug if so).

On iOS this is particularly noticeable, and it might even be that some of the transitions are not functioning on iOS at all (we can test in safari).

If this is going to be fixed, we should take the opportunity to upgrade this animation to use animejs and the v2 api of react-transition group (as per issue #236).

jthrilly:
Hi fidelity mockup of intended animation:

![panel animation](https://user-images.githubusercontent.com/1387940/34522316-f0b3910c-f089-11e7-8979-5cb68f1f43a3.gif)

wwqrd:
Note: use easeOutElastic for easing.

PR #291, opened 2018-01-05
Feature/transition group v2

### LINKED ISSUE #236: Upgrade to v2 spec of react transition group.

wwqrd:
Updates our use of `react-transition-group` to version 2, which has an improved interface and should make it more straightforward to implement our more complex transitions.

Also implements the new staggered transitions outlined in #283 

Any ideas about the poor performance of panel node lists?

### Known issues

- Some slow performance for large lists of nodes, which has been mitigated by only animating the first X nodes. We need to find a better fix!
- ~Modal transitions not implemented yet~
- ~Not linted/tested~

Resolves #236, #285, #283


src/components/Modal.js
```diff
@@ -20,30 +19,26 @@ function Modal(props) {
 
   const classnames = cx('modal', className);
 
+  // if (!show) { return null; }
```
rebeccamadsen: Not needed?

src/components/NodeList.js
```diff
@@ -18,55 +19,100 @@ const EnhancedNode = DragSource(selectable(Node));
...
+    this.refreshTimer = null;
+  }
+
+  componentWillReceiveProps(newProps) {
+    // Don't update if nodes are the same
+    if (isEqual(map(newProps.nodes, 'uid'), map(this.state.nodes, 'uid'))) {
+      return;
+    }
+
+    // if we are on the same prompt/provided same id, then just append the node
```
rebeccamadsen: Is prompt id enough, or do we need stage and prompt?

src/containers/Interfaces/NameGenerator.js
```diff
@@ -112,7 +114,7 @@ class NameGenerator extends Component {
             <NodeList
               nodes={nodesForPrompt}
               label={label}
-              id="MAIN_NODE_LIST"
+              id={`${prompt.id}_MAIN_NODE_LIST`}
```
rebeccamadsen: Ditto with the question whether prompt id versus prompt+stage.

src/components/NodeList.js
```diff
@@ -18,55 +19,100 @@ const EnhancedNode = DragSource(selectable(Node));
...
-      className={classNames}
-      component="div"
-      delay={animation.duration.fast * 0.2}
-      duration={animation.duration.slow}
-      start={animation.duration.slow + 3}
-      transitionName="node-list--transition"
-      transitionLeave={false}
-    >
-      { isOver && willAccept &&
-        <Node key="placeholder" placeholder />
```
rebeccamadsen: If we don't have a placeholder node, do we need a different/stronger highlight effect on hover to differentiate between hover and dragging?

src/components/NodeList.js
```diff
@@ -18,55 +19,100 @@ const EnhancedNode = DragSource(selectable(Node));
...
+        {
+          nodes.map((node, index) => (
+            <NodeTransition
+              key={`${node.uid}`}
+              index={index}
+              stagger={stagger}
+            >
+              <EnhancedNode
+                color={nodeColor}
+                label={`${label(node)}_${index}`}
```
rebeccamadsen: The names with the _[number] is certainly disconcerting when the numbers bounces around based on which panel the node is in!

src/components/NodeList.js
wwqrd: Yes good point, it's yet to be determined so I've added it as an issue.

src/components/Modal.js
wwqrd: 👍

src/components/NodeList.js
wwqrd: The NodeList id isn't required to reference prompt/stage. I've updated the comment to express that, but updated our usage to include stage. It isn't actually needed because components are not reused between stages, but I think it makes the intention clearer.

src/components/NodeList.js
jthrilly: I think we will probably just want to visually increase the intensity of the effect for a valid drop target when hovering.

rebeccamadsen:
I looked at the changes since my last comment, and it looks like most of them are resolved. Although I still have issues with moving a node back to "already mentioned". I seem to have no problem if I am using mock nodes (which calls `TOGGLE_NODE_ATTRIBUTES` three times), but I have problems with new nodes.

E.g. I create a new node from the "Add a Person" dialog, and then in a different prompt drag it to the main nodes. Dragging back highlights "Already mentioned" as if I should be able to remove the node from main nodes. However, when I drop it, it does not remove itself (it calls `TOGGLE_NODE_ATTRIBUTES` two times). It is similar if I drag a node from "previous visit".

jthrilly:
I see the behavior Rebecca is describing, too.

wwqrd:
Am I right in thinking that the expected behaviour is that in both cases it should call that method once?

rebeccamadsen:
One would think.... Unless there is an edge case/no-op going on?

wwqrd:
@rebeccamadsen Turn out the drop areas where being left in state, so navigating between prompts was creating multiple targets in the same place on the screen!

Should be fixed in the last couple of commits.

Think this is ready to merge now.

PR #293, opened 2018-01-11
Feature/roster list

### LINKED ISSUE #265: Roster name generator - list select
screen shot 2017-12-12 at 5 21 35 pm

# Configuration
    {
      "id": "namegenroster2",
      "type": "NameGeneratorListSelect",
      "label": "NG Classmates",
      "prompts": [
        {
          "id": "7cl",
          "text": "Which classmates would you also think of as friends?",
          "additionalAttributes": { // Added to the node object when it is added to the main interview network
            "special_category": 46,
            "close_friend": true
          },
          "dataSource": "schoolPupils", // Key corresponding to external data section of the protocol file.
          "sortOrder": {
            "nickname": "DESC"
          },
        },
      ]
    },

rebeccamadsen:
Resolves #265. This adds a name generator interface using selection from a list, roster style. The "add count" button is just using a placeholder button to add nodes.

untitled


jthrilly:
I took a look at this and it looks great! it was actually not what I was expecting (I had thought you would replace the main node list with the card select view that you get when you click the add new node button), but this actually works pretty well.

The only implementation thing I saw was to do with multiple prompts. I noticed that once you have added some nodes on a given prompt, if you go back to add more, the nodes you have already added aren't there (which is perfect). However those same nodes _can_ be added again on a subsequent prompt. I think we need to reimplement the same logic that we have in the external data side panel, which basically checks if a node with the same attributes already exists in the interview, and if so filters it out of the external data source.

What do you think?

rebeccamadsen:
If we make it so that you cannot add previously used nodes on a subsequent prompt, does that mean we would need the side panels in this interface, so the user could include previously used nodes?

We could just make sure we don't create a duplicate node in the network, but update a node instead of adding a new one.

rebeccamadsen:
This is ready for another look. I have moved the list view to the main panel, removing any add node buttons. Toggling a card adds/removes a node from the network directly.

In the protocol, you can use `showExistingNodes` to alternate between modes. Enabling it means that selected nodes will remain in the list on subsequent prompts, disabling it removes nodes that were selected on other prompts from the current prompt. I am not sure if it matters, but I did wonder whether the "enabled" view ought to add the other prompt attributes on subsequent prompts, but I'm not sure the implications of that. As it is, only the prompt attributes when the card is "selected" are added to the node, even if the card is shown as selected later on other prompts.

The details shown in a card and the fields for sorting can be specified in the protocol:
```
"nodeDisplay": {
    "primary": "nickname",
    "secondary": [
        {
            "label": "Name",
            "variable": "name",
            "sortable": true
        },
        {
            "label": "Age",
            "variable": "age",
            "sortable": true
        }
    ]
}
```

I made the number of columns in grid dependent on width, but I also added flexbox as a fallback because the android version I'm using on cordova doesn't support grid, which surprised me.

I ran into an issue when I experimented with removing the `uid` from the protocol, because we don't add uid's until a node is added to the network. Without a uid, it is hard to know whether a node and the card clicked on are the same. The cards can be filtered and sorted, so order means nothing. For the record, there is also an issue when the uid for the nodes in the main NameGenerator are not provided (e.g. try using a dataset without uids, and drag a node from "previous visit"). If uids aren't required for the data, then this should be a new issue.

jthrilly:
I'll take a look at this today!

bryfox:
This looks good to me!

FWIW, part of #267 may be relevant to the `uid` discussion. When search results are rendered using cards (from master), I manually generate a UID — since what I'm rendering isn't actually a Node. [See comments for `uidForResult` and `isAllowedResult`](https://github.com/codaco/Network-Canvas/commit/2449007baad16e46074aad5e95b1e5a08cc9e402#diff-d7da39c09afc3fb197d8a703ce419fb1R115); I needed a concept of uniqueness anyway to remove added nodes from search results.

Also, as a possible future improvement — with a little refactoring, we could unify the filtering functionality between this interface and the 'autocomplete' interface, unless there's a reason not to.

jthrilly:
I think combining the filtering functionality would be very smart. It has other configurable parameters that could be used to make the feature more useful. We can add this to the issue tracker shortly.

jthrilly:
@rebeccamadsen - if you want to address Bryan's other comments regarding UIDs, I think we are good to merge.

bryfox:
@jthrilly I opened #300 to cover that. It probably makes sense for me to handle it; please update that issue as needed.

@rebeccamadsen I don't think anything needs to change. Requiring `uid` is fine from my point of view; if we wanted something more flexible in the future, we can address later.

rebeccamadsen:
@bryfox Yes to the uniqueness! Your approach works too. I opted for generating a uid in the selector primarily for the rare case that the display isn't enough to disambiguate between nodes. I think the display fields *ought* to be enough most of the time. After all, if the display isn't enough to disambiguate for our purposes, then how can the user be expected to disambiguate??

Though my husband and his father share the exact same first, middle, and last name. They did not use "Jr" or "Sr", so when they were still in the same house, they could have been an edge case!

Issue #300 looks great to me.

bryfox:
>  if the display isn't enough to disambiguate for our purposes, then how can the user be expected to disambiguate

Exactly!

PR #294, opened 2018-01-14
Adding diont library and adding server discovery component.

jamieshark:
screenshot 2018-01-14 14 33 56
Example of the information that's listed from the advertising Server.

Would appreciate next steps to figure out how to get this info over to the UI for both Cordova and Electron. @wwqrd @jthrilly


jamieshark:
FYI, cordova plugin is available here: https://github.com/willemmulder/Diont-for-Cordova

wwqrd:
Electron-wise we probably want to run this in the renderer just to keep it simple, unless it won't run there in which case we'll need to use IPC 😅

Looks like Cordova can just nestle in with the rest of the app code, like in `src/utils`.

I've been working on a thing for the cross platform protocol importing code (https://github.com/codaco/Network-Canvas/tree/feature/protocol-bundle) which might be useful here:

Syntax looks something like:
```javascript
const myModule = inEnvironment((environment) => {
  if(environment === environments.CORDOVA) {
    // ...cordova version
  }

  if(environment === environments.ELECTRON) {
    // ...electron version
  }
});

exports default myModule;
```

jamieshark:
@wwqrd - okay, I'll update using that syntax!

jthrilly:
Another thing Jamie: Would you mind implementing the UI side of this inside a Modal Dialog (component we already have)? Just so that we can test this functionality before we build the new start screen. 

The way I imagine it would be you would add a menu item in the session menu ("Pair with server" or something) that would trigger a dialog that has a simple element like a textarea to display the output of the service discovery. The listening code would only run when this dialog is open.

Does that make sense?

jthrilly:
I have concluded the implementation of this feature in its current form. Known limitations are as follows:

- There is no iOS version of the Diont library, meaning one would need to be created from scratch.
- The functionality does not work in the Android emulator for testing. 
- ChromeOS does not support passing through generic UDP multicast packets. This feature will never work on this platform.

PR #298, opened 2018-01-25
tweak scrollable styles to be more visible and fix regression

jthrilly:


rebeccamadsen:
This looks fine to me in electron and chrome. I don't see the scrollable styles in android, but then I don't in master either, so.

PR #299, opened 2018-01-26
Feature/protocol bundle

### LINKED ISSUE #238: Bundled protocol resources (images, data, audio)
# Overview

Several pieces of planned functionality include the ability for the researcher to embed or "bundle" resources not included in Network Canvas within their interview protocols.

At present we do not have an approach to facilitating this.

Examples of interfaces that would need this functionality are:

- #157 Map Interface (external geojson or CSV files to be used as marker or polygon feature layers)
- #218 Instruction Interface (external images, audio, and video)
- Others?

# Considerations

- We cannot simply use web accessible URLs for these resources (and ask researchers to host them privately) because the app must function without an internet connection.
- The protocol "bundle" (including all assets) must be able to be transferred to the device in a relatively timely manner. 
- Once the resources are available, specifying their use within (for example) an interface could be accomplished using a helper method, like a loader:
``` javascript
import { protocolResourceHelper } from 'protocolLoader';

mapConfig: {
  markers: protocolResourceHelper.resolve('markers.csv'),
}
```

# Implementation
- One options is to use some sort of container format, which is compatible with all our platforms. ZIP seems a good candidate, with the added benefit of compression.
- Alternatively, we could use a 'catalog' file, which would list all resources to be downloaded. These individual files could then be fetched separately. Protocols and their assets could still be transferred manually, by moving the containing folder. 
- We need to be mindful of how local file loading works on our different platforms. In electron we can use node, but in Cordova file APIs are quite different. There is also the question of how/if this would work in development (browser) contexts.

wwqrd:
This feature allows us to import a zipped protocol bundle in the format:

```
myprotocol.canvas/
myprotocol.canvas/protocol.json
myprotocol.canvas/custom.js
myprotocol.canvas/assets/
myprotocol.canvas/assets/example.jpg
myprotocol.canvas/assets/...
```

There's still more work to go to support the external csv feature, but this is mainly an implementation detail now, as this change includes a HOC that can be used to load assets.

# Known issues

It currently works for a fixed archive which exists it `public/demo.canvas`, the idea being that once we have the server distribution working it will load from some kind of temporary directory on the device.

Partially resolves #238


jthrilly:
Followup from Slack conversation: this runs on iOS and Electron (with errors) and doesn't run on Android (with errors). Details below:

## Android

screen shot 2018-01-29 at 4 17 25 pm

## Electron

screen shot 2018-01-29 at 4 20 21 pm

## iOS

screen shot 2018-01-29 at 4 21 12 pm

Issue #300, opened 2018-01-30
Node/Result Filtering is duplicated


bryfox:
We have two new interfaces, each of which involves filtering Nodes or results based on a user's text input. See #293 and #267.

Fuzzy search is required for #293, but is configurable as implemented, so it could be disabled for roster list to match current behavior.

rebeccamadsen:
Does the fuzzy search allow filter to work across multiple fields? In the current implementation, the roster list is filtering based on the values in any displayed field on the cards.

bryfox:
It does; the 'autocomplete' protocol specifies fields like this:

```
          "displayLabel": "nickname",
          "displayProperties": ["name", "age"],
```

... which appears to differ from the roster list protocol.

Perhaps we want to standardize on a format here? I like the one used in roster list (`nodeDisplay`...).

jthrilly:
Either approach is fine! Sorry for the mismatch...

rebeccamadsen:
I see that now. A standard would be great! I'm not picky on format. I know I personally debated whether to have a `sortable` true/false param for each display variable, or whether to have a `sortableFields` array elsewhere. I leaned towards the true/false param so there was less chance of typos, but the array has merits too.

jthrilly:
Okay, here is how I propose we standardize these:

## List

``` JSON
{
  "id": "namegenroster2",
  "type": "NameGeneratorList",
  "label": "NG Classmates",
  "showExistingNodes": false,
  "prompts": [
    {
      "id": "7tk",
      "text": "Which classmates would you talk to outside of class?",
      "additionalAttributes": {
        "talk_friend": true
      },
      "subject": {
        "entity": "node",
        "type": "person"
      },
      "dataSource": "schoolPupils",
      "cardOptions": {
        "displayLabel": "nickname",
        "additionalProperties": ["name", "age"]
      },
      "sortOptions": {
        "sortOrder": {
          "name": "ASC"
        },
        "sortableProperties": [
          {
            "label": "Name",
            "variable": "name"
          },
          {
            "label": "Age",
            "variable": "age"
          }
        ]
      },
      "searchOptions": {
        "fuzzyness": 0.4,
        "matchProperties": ["nickname", "name"]
      }
    }
  ]
}
```

## Autocomplete

``` JSON
{
  "id": "namegenroster1",
  "type": "NameGeneratorAutoComplete",
  "label": "NG Classmates",
  "prompts": [
    {
      "id": "7cl",
[...]

PR #303, opened 2018-01-31
update roster list protocol api

rebeccamadsen:
This updates the protocol api that the roster list uses. Cards and Sort buttons both use the provided labels instead of the variable names. Preparatory for #300.

### DIFF
--- a/src/selectors/name-generator.js
+++ b/src/selectors/name-generator.js
@@ -1,7 +1,7 @@
 /* eslint-disable import/prefer-default-export */
 
 import { createSelector } from 'reselect';
-import { has, omit } from 'lodash';
+import { has } from 'lodash';
 import { makeGetSubject, makeGetIds, makeGetAdditionalAttributes } from './interface';
 import { getExternalData } from './protocol';
 import { nextUid } from '../ducks/modules/network';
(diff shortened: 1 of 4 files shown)


bryfox:
Looks good to me, but I'll let @jthrilly chime in since this is based on his comments/data format.

src/selectors/name-generator.js
```diff
@@ -38,38 +38,30 @@ export const makeGetPromptNodeAttributes = () => {
...
-  ),
+  propSortOptions,
+  sortOptions => (has(sortOptions, 'sortableProperties') ? sortOptions.sortableProperties : []),
 );
 
 export const getSortOrderDefault = createSelector(
-  getSortDefault,
-  sortData => sortData && (Object.keys(sortData)[0] || ''),
+  propSortOptions,
+  sortOptions => sortOptions.sortOrder && (Object.keys(sortOptions.sortOrder)[0] || ''),
```
bryfox: This isn't a comment on this PR, since it existed already, but is the intent to only allow a single sort field? If so, should we make that explicit with something like "sortField" and "sortDirection"? If not, should we use an array?

I imagine key ordering is implementation-specific.

rebeccamadsen: If there is only one then key ordering wouldn't be a problem. ;) But yes, at least I can't imagine a scenario where more than one would make sense, so maybe it should be explicit in the protocol.

jthrilly: I can see that the API as it is now doesn't really lend itself to supporting multiple sort properties, although this would be a likely candidate for a feature request in the future. 

I think the reason we did it this way was just brevity, but I agree that it does introduce readability problems. Since it is consistent throughout the app right now, let's leave it for a separate issue.

Issue #305, opened 2018-01-31
Incorrect state of Nodes on prompt change (edge cases)


bryfox:
NodeList#componentWillReceiveProps() involves some nested state setting and timeout-driven animations. I'm opening this issue to better understand what's happening here.

- <strike>It appears that I need to specifically cache label strings on nodes, in the case where the displayed field differs between prompts, but that shouldn't be necessary.</strike> Update: This separate issue was moved to #311.
- It appears that double-clicking the prompt can produce an incorrect state, where a node appears both in 'already mentioned' and 'nodesForPrompt'

Here's one way to reproduce the latter:

1. Reset Session
2. Load demo protocol
3. On Name generator...
4. Prompt0: Drag one or more nodes to the main list
5. Click prompt swiper to advance to Prompt1
6. Click prompt swiper to advance to Prompt2
7. Double-click prompt swiper to advance through Prompt0 to Prompt1

bryfox:
#### 1. setState called within timeouts may result in incorrect state.

The NodeList is shared between prompts. To make animations work, rendering is driven from component state rather than directly from prompts. Timeouts are used to clear and then show nodes. If a user's second click is faster than the timeout, then the (previous) state will overwrite the 'new' state. If the 'new' state doesn't also trigger a timeout, then the displayed state will be inconsistent until further user action.

Solution: remove timeouts; adjust transition code to make animations continue working as expected.

#### 2. "undefined" labels during transition

NodeList state is updated aynchronously from props. The display label formatter/callback is formatting a node from an 'old' prompt (see above). The formatter could theoretically use the promptId to look up the correct `displayLabel` for the node it's formatting, but (at least presently) it would need to check 

Solution: ideally, NodeList's rendering will be driven from props, so that everything is driven from React props as expected. I'm working on getting transitions to work correctly here.

#### 3. potential for "undefined" labels

Given that currently:

- a prompt's nodes (`networkNodesForPrompt`) are driven from the prompt's `additionalAttributes` in the protocol definition,
- a protocol `dataSource` is associated with a prompt, not a stage (a network can consist of different node types), and
- no business logic prevents a protocol from re-using `additionalAttributes` between prompts

...a prompt's displayLabel property may refer to a field that doesn't exist on one or more nodes.

<img width="622" alt="undefined" src="https://user-images.githubusercontent.com/39674/35745990-4061d560-0813-11e8-9931-9f7bd823b52f.png">

bryfox:
[...]

PR #313, opened 2018-02-06
Render NodeList without timeouts

bryfox:
This removes the 'timeout' used to set animation transitions between
nodes in the node list. The transition is now achieved by a timeline
animation in the NodeTransition itself.

Resolves the first 'timeout' issue described in #305.

### DIFF
--- a/src/components/NodeList.js
+++ b/src/components/NodeList.js
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 import { find, get, isEqual } from 'lodash';
 import cx from 'classnames';
 import Color from 'color';
-import { Node, animation, colorDictionary } from 'network-canvas-ui';
+import { Node, colorDictionary } from 'network-canvas-ui';
 import { TransitionGroup } from 'react-transition-group';
 import { Node as NodeTransition } from './Transition';
 import { scrollable, selectable } from '../behaviours';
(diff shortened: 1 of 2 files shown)


wwqrd:
I like the direction this is going (consolidating the animations into the transition component), but it looks to me like this introduces some visual artefacts that weren't there before (try clicking through the prompts rapidly)?

I'm struggling to reproduce the original issue too (though it looks to be correct theoretically), can you give me some precise steps?
> The NodeList is shared between prompts. To make animations work, rendering is driven from component state rather than directly from prompts. Timeouts are used to clear and then show nodes. If a user's second click is faster than the timeout, then the (previous) state will overwrite the 'new' state. If the 'new' state doesn't also trigger a timeout, then the displayed state will be inconsistent until further user action.

Would moving the `clearTimeout` fix it?

```diff
diff --git a/src/components/NodeList.js b/src/components/NodeList.js
index e64b28196..eaa4134f1 100644
--- a/src/components/NodeList.js
+++ b/src/components/NodeList.js
@@ -34,6 +34,8 @@ class NodeList extends Component {
   }

   componentWillReceiveProps(newProps) {
+    if (this.refreshTimer) { clearTimeout(this.refreshTimer); }
+
     // Don't update if nodes are the same
     if (isEqual(newProps.nodes, this.state.nodes)) {
       return;
@@ -54,7 +56,6 @@ class NodeList extends Component {
       this.setState(
         { nodes: [], stagger: true },
         () => {
-          if (this.refreshTimer) { clearTimeout(this.refreshTimer); }
           this.refreshTimer = setTimeout(
             () => this.setState({
               nodes: newProps.nodes
```

bryfox:
I've updated #305 with repro steps.

> Would moving the `clearTimeout` fix it?

It definitely fixes the case identified by the above steps. And it removes the visual artefacts when clicking fast.

With that modification, is it possible for a timeout to be set, some prop (other than nodes) to change, so that the timeout is canceled, the next line returns and the correct nodes are never set (from the previous timeout)? I don't know what would cause that, but it looks theoretically possible (either now or in the future). Still, it might be a better approach at least for now.

bryfox:
I'm closing this one. Visual artifacts may be solvable, but we'll discuss other possibilities first.

Issue #312, opened 2018-02-06
Development workflow for demo protocol.


jthrilly:
For electron/cordova:
- New protocols are installed to a specific directory on device. This installation consists of (1) transfer, (2) extraction, and (3) copying.
- On first run, "factory" protocols get installed (extracted to device storage) through this method. The app knows to look for them in a specific predefined location, such as `public/protocols/`. They will be stored here as folders, so no unzip process will be required, just a straight copy to the device storage.
- After this, factory protocols can be run as normal.

For web:
- New Protocols cannot be installed.
- "factory" protocols remain in `public/protocols` where they can be loaded directly.

PR #315, opened 2018-02-06
Feature/factory protocols

wwqrd:
This feature allows us to bundle multiple fully fledged protocols with the app so that we can:

1. Provide different full-featured scenarios to potential new users
1. Work with protocols that more closely match production spec in development and testing.

**Context**: Previously our "demo protocol" feature loaded a hard-coded JSON file and didn't support assets.

### Details

"Factory" protocols are bundled in the app directory, and loaded directly from there.

We have a new action to load these protocols, `LOAD_FACTORY_PROTOCOL`.

For electron/cordova this is achieved with a combination of: alternative paths; updating the Electron `asset://` protocol to check both locations (`app`/`appData`) directories for assets, defaulting to `app`.

For web (dev), we only needed to add implementations for `assetUrl` (which points to localhost) and `loadProtocol` (which uses `fetch()`).

Resolves: https://github.com/codaco/Network-Canvas/issues/312


public/components/assetsProtocol.js
```diff
@@ -1,14 +1,25 @@
 const path = require('path');
-const electron = require('electron')
+const electron = require('electron');
+const fs = require('fs');
 
 const userDataPath = (electron.app || electron.remote.app).getPath('userData');
+const appPath = (electron.app || electron.remote.app).getAppPath();
+
+// check for factory protocl first?
```
rebeccamadsen: is this still a question?

public/protocols/demo.canvas/protocol.json
```diff
@@ -180,218 +180,6 @@
           "name": "Local Bar"
         }
       ]
-    },
-    "schoolPupils": {
```
rebeccamadsen: Is it worth having a development protocol as well so we don't need to recreate the data for testing? Although then we run the risk of having only one protocol be updated when we make changes to the api and not all of them, so maybe not...

bryfox:
Seems to import fine for me on iOS & Android. 👍 

I also tried testing local (imported protocol) changes by updating & replacing the  iOS app container, but wasn't able to see changes — but I'm probably just looking in the wrong place for cordova file imports?

public/protocols/demo.canvas/protocol.json
jthrilly: I think we need a plan for changing the development protocol in feature branches more generally. 

We can discuss this at the kickoff meeting for the next release.

In general, the plan is to have a minimum of two factory protocols: one for developers to test edge cases and new features, and one for the rest of the team and our test sites to have something that somewhat "makes sense", so that they can evaluate and provide feedback.

rebeccamadsen:
No more errors on my end!

Issue #327, opened 2018-02-08
iOS asset loading issues


bryfox:
In the demo protocol, the new instruction interface should display an image and have an audio playback feature.

On an iPad running iOS 11, the audio player displays "Error" and cannot play.

![device](https://user-images.githubusercontent.com/39674/35981974-9a84feca-0cbc-11e8-9483-227ad446dfbd.jpeg)

Additionally, on the latest simulator, images don't show on either the instructions or sociogram.

![simulator](https://user-images.githubusercontent.com/39674/35982035-c8434678-0cbc-11e8-8648-c374be04ff01.png)

Tested: master branch @ 13cbb3a after re-running `npm run dist:ios`.

rebeccamadsen:
I see something similar running android in the simulator. Loaded demo in android, on instruction interface image/audio are not loaded. Also on sociogram, the duck image is not loading, and on console I see this error:
```Failed to load resource: net::ERR_FILE_NOT_FOUND
Uncaught (in promise) FileErrorcode: 1__proto__: Object
index.html#/protocol:1 
```
It looks great in Chrome and Electron, which means I can't find a better error message, sadly.

wwqrd:
Audio not working on an actual device either.

wwqrd:
Factory protocols are loading assets from the imported protocol location. So only assets that are present (say from an older import) are available. Looking into a fix.

Issue #333, opened 2018-02-08
Protocol load error displays could be improved


bryfox:
When a protocol fails to load (or process to completion), the error message presented in the GUI may be unhelpful (e.g., `[Object object]`). The basic test case is to enter a custom URL and tap load protocol.

Note, I see a reasonable error message on electron for 404s, although the displayed URL ("stuk.github.io/jszip/documentation/howto/read_zip.html") looks like developer documentation...

I wonder if having our own generic error message for all of these cases is preferable (or perhaps one for network/server and one for processing).

## [Electron] When URL is a directory

![electron-folder](https://user-images.githubusercontent.com/39674/35983045-381b8238-0cbf-11e8-9dc2-f678d153c438.png)

## [iOS] URL that 404s

![ios-404](https://user-images.githubusercontent.com/39674/35983496-2320f1c8-0cc0-11e8-88af-22cf263617f9.png)

(Any 404 should do, but URL tested was the default, minus the final "s".)

## [Electron] SSL trust error

This may not be an issue in practice, but can be reproduced by importing from a server using a self-signed cert (e.g., the dev server in https mode). In this case, no error is presented. There's an uncaught error in the console and the app spinner displays indefinitely until the app is reloaded.

jthrilly:
The `<Dialog />` component has `additionalInformation`, which adds this to a textarea. This could be enhanced to be scrollable, and to allow easy copy and paste for bug reporting. 

Could take place as part of #321

wwqrd:
This part of the app is all handled with promises, with a catch all for errors.

We could either expand this catching functionality to better interpret the errors coming in.

Default to a plain text message for now "Sorry for the inconvenience etc.

The *actual* error is still logged to the regular console.

jthrilly:
I would be in favor of expanding the catching to interpret the platform specific errors better. I also think we should keep debug info in there for errors that are unexpected. Combining the two would mean basically "parsing" the error format specific to each platform, and putting it in a format that can be displayed in the `additionalInformation` box (which is the pattern I want to use for extra debug information - @wwqrd  is going to be creating an issue on improving this design later today).

To go along with that, we should have a series of human readable text responses for the most common errors. For example:

- 404 - "We couldn't find a Network Canvas protocol at the location you gave us. Check the location, and try again."
- Invalid path/URL - "The location you gave us doesn't seem to be valid. Check the location, and try again."
[...]

PR #338, opened 2018-02-08
In Cordova attempt to load assets from factory location first

wwqrd:
In Cordova attempt to load assets from factory location before the app data location.

This means that factory protocol assets should now load correctly

# Known issues

If a factory protocol name matched an imported protocol name, the factory protocol assets would be loaded (this is already the case for electron).

Resolves: #327 
Partial: #333

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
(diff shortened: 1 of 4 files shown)


rebeccamadsen:
Looks good to me!

bryfox:
Looks good here as well — tested on iOS & Android.

PR #340, opened 2018-02-08
Fix withBounds() throttle cancellation

### LINKED ISSUE #328: setState warning message
On Chrome, I see this warning message whenever I leave the sociogram stage:

```
proxyConsole.js:56 Warning: Can only update a mounted or mounting component. This usually means you called setState, replaceState, or forceUpdate on an unmounted component. This is a no-op.
```

It doesn't seem to affect anything...

bryfox:
Cancellation was not working in conjunction with throttle.

Timer function should use either throttle or reqAniFrame (not both),
as throttle calls setTimeout internally.

Using throttle makes sense for 'infrequent' updates (where
`maxUpdatesPerSecond` much less than 60fps), and the API is clearer.

Fixes #328.

### DIFF
--- a/src/behaviours/withBounds.js
+++ b/src/behaviours/withBounds.js
@@ -22,7 +22,7 @@ export default function withBounds(WrappedComponent) {
       this.state = initialState;
       this.lastState = initialState;
 
-      this.trackSize = throttle(this.trackSize, 1000 / maxUpdatesPerSecond);
+      this.throttledTrackSize = throttle(this.trackSize.bind(this), 1000 / maxUpdatesPerSecond);
     }
 
     componentDidMount() {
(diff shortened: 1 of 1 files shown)


jthrilly:
Great catch, Bryan. Just wondering if you could say more about why throttling is better than requestAnimationFrame for infrequent updates. I was under the impression that they solve different (but sometimes overlapping problems).

bryfox:
> they solve different (but sometimes overlapping problems)

I agree with that, and we could certainly use the latter here.

I'd replace "better than..." with "sufficient" in your statement. I picked throttle() because the API & intent seemed clearer. My comment was more a justification for that choice; I don't think we really need the resolution or accuracy of reqAniFrame here.

jthrilly:
Okay, excellent clarification. 

The reason I ask is that the intent here is to reduce animation jank, which is one of those overlapping use cases. Sometimes a heavy function call blows a rendering budget. You can either throttle it, which will help if it interacts with over, infrequent, events. OR make sure to requestAnimationFrame to give it a chance of not interacting poorly with a UI process. 

I think if we were being super proper about about we would probably profile the execution time for the function and do some quick arithmetic. If it is under our frame budget, we are fine using throttle. If it is over, I would prefer requestAnimationFrame.

jthrilly:
Either way, thanks for the explanation!

wwqrd:
Let's :shipit:

PR #341, opened 2018-02-09
Server/server discovery mdns

jthrilly:
## Overview of issue

Because of the limitations of the first attempt at Server discovery within Network Canvas (see #294), I looked for a different implementation strategy. Given that:

- ChromeOS does not support UDP multicast packet forwarding to Android apps.
- Diont does not have a Cordova plugin for iOS.

I have implemented a different protocol, both on the node side and the Cordova side.

## Technical solution
Chromecast devices use multicast DNS (which is essentially Apple's Bonjour protocol, and also known as zeroconf), and Google supports this functionality on ChromeOS and Android. This is the preferred method of service advertisement/discovery across macOS, Android, ChromeOS, and iOS.

Windows and Linux can have Bonjour support installed, either by installing Bonjour directly (already installed with iTunes) on Windows, or by installing `libavahi-compat-libdnssd-dev` on Linux. This will become a suggested dependency of both Network Canvas and Server, with a fallback mode for when it fails. There are several options with regards to how to handle this dependency, from creating combined installers, to providing additional installation files, to providing instructions on our website and a link to relevant downloads.

The development side is more complex. The MDNS node module must be built using `node-gyp` against the necessary bonjour/avahi headers when it is installed. This means that all developers must have these headers installed before running `npm install`. 

MacOS has them by default. Linux devs just need to install the same single lib as end-users. Windows is a real pain. You need visual studio, the Bonjour SDK, the Bonjour service, and potentially a specific node version (TBD). 

## Added by this branch
This PR implements MDNS, using `node_mdns` and `cordova-plugin-zeroconf`. It is tested and working in the following configuration:

> macOS host advertises a service
- [x] Seen by NC on Android
- [x] Seen by NC on ChromeOS
- [x] Seen by NC on Windows
- [x] Seen by NC on Linux
- [x] Seen by NC on macOS

## Todo

- [x] Need error handling for the case where a user doesn't have the dependencies installed (fallback mode). How gracefully can we handle this? Will the app crash on launch?
- [x] Need error handling for edge cases, such as needing to add `mdns6_minimal` to `etc/nsswitch.conf` on Linux. Can we catch these errors effectively?
- [x] Need to test npm install process for Windows based developers. My testing indicated we could not build the MDNS module on node 8.5.x, but 9.5.x was functioning. There are several other environment requirements.
- [x] Need to verify that service advertisement on other platforms (Windows in particular) functions correctly, and is received by all platforms.


src/utils/serverDiscoverer.js
```diff
@@ -0,0 +1,79 @@
...
+  constructor() {
+    if (!isElectron() && !isCordova()) {
+      return {
+        on: () => { /* noop */ },
+        off: () => { /* noop */ },
+        init: () => { /* noop */ },
+      };
+    }
+
+    const EventEmitter = require('eventemitter3');
```
wwqrd: Can this be an import?

src/utils/serverDiscoverer.js
```diff
@@ -0,0 +1,79 @@
...
+      this.browser.stop();
+    }
+
+    if (isCordova()) {
+      this.zeroconf.unwatch('_network-canvas._tcp.', 'local.');
+    }
+  }
+
+  init() {
+    if (navigator.onLine) {
```
wwqrd: Can we add `window.`? I found this unclear.

src/utils/serverDiscoverer.js
```diff
@@ -0,0 +1,79 @@
...
+          this.events.emit('SERVER_ERROR', error);
+        }
+      }
+
+      if (isCordova()) {
+        this.zeroconf = window.cordova.plugins.zeroconf;
+        try {
+          this.zeroconf.watch('_network-canvas._tcp.', 'local.', (result) => {
+            const action = result.action;
+            const service = {
```
wwqrd: Cordova and Electron implementations need to match so that the consumer of ServerDiscovery doesn't need to know the internals.

src/components/ServerList.js
```diff
@@ -0,0 +1,121 @@
...
+      }));
+    });
+
+    this.serverDiscoverer.on('SERVER_REMOVED', (response) => {
+      console.log('Looking for:');
+      console.log(response);
+
+      this.setState(prevState => ({
+        // eslint-disable-next-line
+        servers: prevState.servers.filter(item => !(item.name == response.name && item.index == response.index)),
```
bryfox: I'm testing on iOS. When a service is removed, its index comes through as `-1`, so this never matches.

jthrilly: Thanks Bryan - I've caught and fixed this in a later commit which I'll push up tomorrow.  React is a cruel mistress.

jthrilly:
@wwqrd - I've updated all the stuff you caught, albeit with some small modifications. If this checks out for you, I'd like to merge.

src/utils/serverDiscoverer.js
```diff
@@ -0,0 +1,103 @@
...
+            {
+              name: service.name,
+              host: service.hostname,
+              port: service.port,
+              addresses: [...service.ipv4Addresses, ...service.ipv6Addresses],
+            }
+          );
+
+          if (action === 'added') {
+            // Do nothing - we need a resolution which provides an IP address.
```
wwqrd: Could use a guard `return` at beginning of function?

src/utils/serverDiscoverer.js
```diff
@@ -0,0 +1,103 @@
...
+
+    if (isCordova()) {
+      this.zeroconf.unwatch('_network-canvas._tcp.', 'local.');
+    }
+  }
+
+  handleNetworkChange = () => {
+    this.checkEnvironment();
+    if (window.navigator.onLine) {
+      this.events.emit('SERVER_RESET');
```
wwqrd: I assume this means that we are now relying on mdns errors and "Automatic server discovery unavailable" for "no network" warnings?

wwqrd:
Looks fine to me, only an optional style suggestion.

src/utils/serverDiscoverer.js
jthrilly: Yeah, I changed it so that the ServerList doesn't have to be concerned with network status (ServerDiscoverer handles it, and emits events accordingly), which I think is a better design.

wwqrd: I think that's a good shout!

PR #348, opened 2018-02-10
Fix/android bars

### LINKED ISSUE #332: search field covered up by keyboard
On Android simulator, I noticed that the search field in the autocomplete roster is completely covered up by the soft keyboard. I can't see what I'm typing in the field at all without closing the keyboard, which defeats the purpose. Obviously this is not an issue in Electron or Chrome.

This is not a bug, as this was the design of this feature from initial concept. If I only want to type a character or maybe even two, this may not be a problem. But longer words are prone to mistake without the visual feedback of what I've already typed.

bryfox:
This customizes the Android Activity to set status & navigation bar colors per #332 (and fixes the problems with fullscreen described there).

Modifying NC's MainActivity is the most straightforward way to make this happen. Since the app is [git-]ignored and generated on the fly, the customized activity is copied into the app during the build phase. I'm certainly open to other suggestions here.

I've tested this according to the [updated] comments and docs, which are how I've been working with cordova here. If anyone runs into anything unexpected, please let me know.

Note: tested only in 8.0 Pixel C emulator. Min Android SDK is now 21.

Fixes #332.

android-bars


config.xml
```diff
@@ -30,31 +37,24 @@
...
+    <preference name="android-maxSdkVersion" value="21" />
     <plugin name="cordova-plugin-app-version" spec="^0.1.9" />
     <plugin name="cordova-plugin-x-socialsharing" spec="^5.2.0" />
     <plugin name="cordova-plugin-wkwebview-engine" spec="^1.1.4" />
     <plugin name="cordova-plugin-file" spec="^5.0.0" />
     <plugin name="cordova-plugin-file-transfer" spec="^1.7.1" />
     <plugin name="cordova-plugin-whitelist" spec="^1.3.3" />
     <engine name="android" spec="^7.0.0" />
-    <engine name="ios" spec="latest" />
+    <engine name="ios" spec="^4.5.4" />
```
bryfox: This change was automatic from running cordova scripts; I assume OK.

jthrilly:
Only thing I found is that you need to do a specific `cordova build android` before testing, whereas up until now I've just done `cordova run android --device` and that has included a build. 

The actual functionality looks great.

Issue #331, opened 2018-02-08
Ignored attempt to cancel a touchmove


rebeccamadsen:
I saw this on the remote console while running Android in debug mode on a simulator, but there don't seem to be any ill effects that I can see, so this may be something we could ignore. Especially if it doesn't show up on device, but I can't test that myself.

While dragging nodes around, this error crops up multiple times, over and over as long as you keep dragging:
```
main.d96239b5.js:218262 Ignored attempt to cancel a touchmove event with cancelable=false, for example because scrolling is in progress and cannot be interrupted.
detectMoveType	@	main.d96239b5.js:218262
onMove	@	main.d96239b5.js:218304
```

It does not show up in Electron or Chrome.

bryfox:
I looked at this a bit; code seems correct in use of `passive`. 

Further, if I comment out the line it claims to be ignoring, then the drag doesn't work properly; I can 'pick up' a node and then have my touch scroll the container instead of tracking drag.

I'm not sure there's anything we should do here other than leave a related comment in the code, but @wwqrd knows more about drag/drop...

jthrilly:
I'm going to close this because it doesn't seem to be an actual bug.

PR #349, opened 2018-02-12
Fix/improved errors

### LINKED ISSUE #333: Protocol load error displays could be improved

wwqrd:
Resolves #333

Adds human readable errors for download/import/load protocol, and pattern for handling errors in future.

 # Known issues
- [x] Functions that aren't defined for a particular environment throw errors. Do they need to return promises for correct handling? If so update to Promise.reject()
- [x] Re-test on devices

### DIFF
--- a/src/utils/protocol/downloadProtocol.js
+++ b/src/utils/protocol/downloadProtocol.js
@@ -69,12 +69,11 @@ const downloadProtocol = inEnvironment((environment) => {
         )
         .catch((error) => {
           const getErrorMessage = ({ code }) => {
-            if (code === 1) return "We couldn't find a Network Canvas protocol at the location you gave us. Check the location, and try again.";
-            if (code === 3) return "Your device doesn't have an active internet connection, so we weren't able to fetch your protocol at this time. Connect to a network, and try again.";
-            return "The location you gave us doesn't seem to be valid. Check the location, and try again.";
+            if (code === 3) return networkError;
+            return urlError;
           };
 
-          friendlyErrorMessage(getErrorMessage(error))(error);
+          getErrorMessage(error)(error);
         });
   }
 
--- a/src/utils/protocol/downloadProtocol.js
+++ b/src/utils/protocol/downloadProtocol.js
@@ -18,6 +18,9 @@ const getURL = uri =>
 
 const getProtocolNameFromUrl = url => url.pathname.split('/').pop();
 
+const urlError = friendlyErrorMessage("The location you gave us doesn't seem to be valid. Check the location, and try again.");
+const networkError = friendlyErrorMessage("We weren't able to fetch your protocol at this time. Your device may not have an active network connection, connect to a network, and try again.");
+
 const downloadProtocol = inEnvironment((environment) => {
   if (environment === environments.ELECTRON) {
     const request = require('request');
@@ -32,9 +35,9 @@ const downloadProtocol = inEnvironment((environment) => {
           url.href,
           path.join(tempPath, getProtocolNameFromUrl(url)),
         ])
-        .catch(friendlyErrorMessage("The location you gave us doesn't seem to be valid. Check the location, and try again."))
+        .catch(urlError)
         .then(([url, destination]) => writeStream(destination, request({ uri: url })))
-        .catch(friendlyErrorMessage("We weren't able to fetch your protocol at this time. Your device may not have an active network connection, connect to a network, and try again."));
+        .catch(networkError);
     };
   }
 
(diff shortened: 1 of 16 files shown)


src/utils/protocol/importProtocol.js
```diff
@@ -86,6 +77,28 @@ const extractZip = inEnvironment((environment) => {
...
 
+const loadZip = inEnvironment((environment) => {
+  if (environment === environments.CORDOVA || environment === environments.ELECTRON) {
+    return source =>
+      readFile(source)
+        .then(data => Zip.loadAsync(data))
+        .catch(friendlyErrorMessage("We couldn't find a Network Canvas protocol at the location you gave us. Check the location, and try again."));
+  }
+
+  throw new Error(`loadZip() not available on platform ${environment}`);
```
jthrilly: Do `loadZip()` errors not trigger a friendly error? What happens to them? What should the user do?

src/utils/protocol/downloadProtocol.js
```diff
@@ -23,27 +27,37 @@ const downloadProtocol = inEnvironment((environment) => {
...
-              encodeURI(uri),
+              url,
               destination,
               () => resolve(destination),
               error => reject(`Error code: ${error.code}. (${error.source})`),
             );
           }),
-        );
+        )
+        .catch(friendlyErrorMessage("Your device doesn't have an active internet connection, so we weren't able to fetch your protocol at this time. Connect to a network, and try again."));
```
jthrilly: Should this be "active network connection", since some protocol URLs could be local?

src/utils/protocol/importProtocol.js
wwqrd: There are two workflows going on here:

1. You called a method that isn't supported on the platform
1. One of the steps threw an error.

This is the first one, which will probably go straight to the console at the moment, since it's the wrong type for our promise chains (separate issue: it's generally good practice to always return the same type, so maybe we should return a promise rejection here?).

Should we capture this kind of error for the user? I'm not sure. What do you thing?

src/utils/protocol/downloadProtocol.js
```diff
@@ -23,27 +27,45 @@ const downloadProtocol = inEnvironment((environment) => {
     return (uri) => {
       const tempPath = (electron.app || electron.remote.app).getPath('temp');
 
-      return getProtocolNameFromUri(uri)
-        .then(protocolName => path.join(tempPath, protocolName))
-        .then(destination => writeStream(destination, request(uri)));
+      return getURL(uri)
```
bryfox: There's another case here that isn't caught; I'm not sure why. It's described by the SSL trust case in #331 but happens more generally any time a remote server can't be reached.

Test by hitting a well-formed URL at `localhost:7788` (any non-running local port) or some hostname known to not exist.

In these cases, an uncaught error shows in the console and UI spins indefinitely until reload. If we expect users to be able to import from arbitrary servers, it would be good to catch this case.

(I'm on electron/macOS.)

src/utils/protocol/importProtocol.js
jthrilly: I'm just thinking in terms of testing. Users won't have the console open, so will have no means of knowing what happened. I realise this is basically a noop, but we want to show an error when things that shouldn't happen do. I think returning a promise rejection is the right thing to do.

rebeccamadsen:
On Chrome, when the app loads, the console shows the following warning:
```
Warning: Failed prop type: Invalid prop `additionalInformation` supplied to `Dialog`.
    in Dialog (at ErrorMessage.js:35)
    in ErrorMessage (created by Connect(ErrorMessage))
    in Connect(ErrorMessage) (at App.js:40)
    in div (at App.js:18)
    in App (created by Connect(App))
    in Connect(App) (at index.js:17)
    in Provider (at index.js:16)
__stack_frame_overlay_proxy_console__ @ bundle.js:227145
```

src/utils/protocol/downloadProtocol.js
wwqrd: Great catch!

wwqrd:
@rebeccamadsen Chrome issue should be resolved
@jthrilly The environment errors now respect type, so will get captured in the promise methods
@bryfox I've updated this to use promises rather than streaming, which proved very difficult to capture. The electron errors are now much more fine grained, the Cordova ones are still a bit indiscriminate.

jthrilly:
Looks great to me.

bryfox:
Confirmed chrome issue is fixed as well.

PR #363, opened 2018-02-14
Teaching Protocol

jthrilly:
This PR adds the educational/teaching protocol developed by the team (closes #322). Some features, notably the ordinal bin, are not yet implemented, but the protocol configuration has been written. This requires the latest version of Network-Canvas-UI. 

One change is outstanding: create a short movie that shows the node linking process

Please file issues for any bugs you see with this PR that **do not result from changes added in this PR**. These will be fixed in the next release. Issues introduced here will be fixed in this PR as usual.

PR #365, opened 2018-02-14
Fix/sociogram highlighting

### LINKED ISSUE #358: Issues with highlighting on Sociogram
The Sociogram has issues related to the `highlight` configuration parameter.

The highlight parameter allows nodes to be put into the selected state based on their properties. It also controls whether these properties should be toggle-able on the node (by tapping a node, these properties are created or removed).

Example:
``` JavaScript
"highlight": {
  "variable": "has_given_advice", // Node property to examine
  "value": true, // Value to be matched if the node is to be highlighted
  "allowHighlighting": true // If toggling this property and value are to be permitted.
}
```

Intended implicit default:
``` JavaScript
"highlight": {
  "allowHighlighting": false
},
```

### Issues

- Omitting the `highlight` property entirely makes all nodes enter the selected state. The default should be all nodes deselected, and no selection possible.
- Setting `allowHighlighting: false` fixes the above, but when creating edges, nodes do not de-select once an edge has been successfully created/removed. Nodes remain in the selected state.

### Intended behaviour
- The implicit default for the interface should be as above, with no nodes highlighted, and no highlight toggling.
- Edge linking behaviour (the presence of `edges.create` should override `highlight.allowHighlighting`, and set it to false.
- Edge creation should use a different state (propose: `linking`) to avoid confusion/reuse of the `selected` state. The new `linking` state will trigger a repeating animation to visually indicate that the node is in a temporary state. This animation will not interfere with the current selected state animation. Better naming welcome.

wwqrd:
Resolves #358

Changes behaviour of Sociogram configuration:
- make highlight disabled by default, 
- make edge creation override toggling highlight
- Use new linking prop on Nodes to indicate edge creation

### DIFF
--- a/src/containers/ConcentricCircles/LayoutNode.js
+++ b/src/containers/ConcentricCircles/LayoutNode.js
@@ -17,6 +17,7 @@ class LayoutNode extends PureComponent {
     const {
       node,
       selected,
+      linking,
       allowPositioning,
       allowSelect,
       layoutVariable,
@@ -39,6 +40,7 @@ class LayoutNode extends PureComponent {
           label={label(node)}
           onSelected={this.props.onSelected}
           selected={selected}
+          linking={linking}
           allowDrag={allowPositioning}
           allowSelect={allowSelect}
           meta={() => ({ ...node, itemType: 'POSITIONED_NODE' })}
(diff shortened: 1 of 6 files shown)


jthrilly:
Do you mind testing this with my PR (#363) since it implements these options?

PR #368, opened 2018-02-14
allow consecutive name generators to open node forms

### LINKED ISSUE #360: Node form doesn't open with consecutive Namegenerators
If you place two Name Generator interfaces next to each other in an interview, the new node form does not open on the second of the two. Navigating directly to the second interface on a fresh load allows the modal to open as normal.

This seems to be to do with the modal unregistering after the second interface has loaded (just based on looking at the console activity).

rebeccamadsen:
Fixes #360.

Moving the Symbols for the forms to a class level variable keeps the name generators from clobbering each others' modals. Test by editing the protocol to have two NameGenerator interfaces one after another.


bryfox:
Nice. Fixes for me. 👍

PR #380, opened 2018-02-15
Add logging feature to electron

wwqrd:
So we can get to the bottom of the asar issue, here's a logging method which dumps to a file.


wwqrd:
Not required, we already have one in the project 'electron-log'

Issue #379, opened 2018-02-15
Android emulator videos


rebeccamadsen:
On Android emulator, the videos do not play. See the "Drawing Ties" screen in the teaching protocol for an example. Fortunately it plays on devices, so this is not urgent, and possibly something not to fix at all.
```
Refused to load the image 'android-webview-video-poster:default_video_poster/-141442556822233072' because it violates the following Content Security Policy directive: "img-src asset: 'self' data: cdvfile: file:".

index.html#/protocol:1 Uncaught (in promise) FileError
```

wwqrd:
How do I recreate this bug? Is it a specific protocol?

rebeccamadsen:
The teaching protocol has a video on the "Drawing Ties" screen. When I get to that screen in the android emulator, it briefly shows video controls and then even those disappear.

wwqrd:
Adding `android-webview-video-poster:` to the img-src Content Security Policy makes the error go away (and is potentially something we should keep), but doesn't resolve the issue of videos not playing.

With no errors, the video tag rendering, and the data url populated, I'm temped to agree with you in maybe leaving this as-is.

PR #383, opened 2018-02-16
Add android poster security policy to img-src

wwqrd:
Relates to #379

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


rebeccamadsen:
I can verify Chrome and Electron are fine with this, and that the error on Android emulator is gone. You are right that it doesn't solve the issue of viewing the video in Android emulator. Assuming this is still good on actual devices with this change, looks fine to me.

jthrilly:
I've looked into this a bit more, and it seems it is an android emulator issue. It also seems that the workarounds are not worth the effort (additional Cordova plugins), so I think we should conclude this PR fixes #379.

PR #392, opened 2018-02-16
Add 'live reload' option for running cordova

### LINKED ISSUE #350: Speed up development workflow for mobile apps

bryfox:
This adds a 'live reload' dev workflow for mobile devices & simulators. Pointing cordova at the running dev server ended up being more straightforward (and a better dev ux) than flushing a [modified] dev bundle to disk & syncing.

Testing:
- Run `npm start` as usual
- Run `npm dev:ios` or `npm dev:android` (see readme for more options)
- Change something in app source

To make this work, a few things are happening:
- Webpack dev server leaves a '.devserver' file while running, which contains the LAN URL that devices & simulators can use.
- Cordova hooks prepare a 'dev server' config.xml before run, and restore
the original once the app launches. (This is pretty quick on device or if an emulator happens to be running; slow otherwise.)

Once running, mobile apps pick up live changes just like other browsers.

This is only triggered for the `run` stage of cordova, and a `LIVE_RELOAD` env var must be set. I think that's sufficient to keep this away from dist builds.

Caveats/warnings:

The primary drawback here is that we actually have to modify config.xml on disk, as far as I could see from looking at cordova source. It's reset to the original once the `run` command completes. But, [warning] if you interrupt then there's nothing currently catching that case. And if you do something like `git commit -a` while your emulator is starting up, you could wind up with unwanted config.xml changes.

We can discuss approaches to make this more robust, but I didn't want to continue right now. It may be worth seeing how well this works for people first.

<strike>TODO: test Windows.</strike>

Fixes #350.


rebeccamadsen:
Yay for live reload! However, Windows doesn't set environment variables like that. Would something like this work for you?
`"dev:android": "node scripts/check-dev-server.js && LIVE_RELOAD=1 || set LIVE_RELOAD=1 && cordova run android",`

bryfox:
@rebeccamadsen thanks! Latest commits address that; I think an `export` is needed. Looks like windows still prints an error about a missing command, but that's hopefully just a minor annoyance.

[cross-env](https://www.npmjs.com/package/cross-env) might be worth including if there are any additional problems related to this.

I've also updated the docs with some more notes on these tasks & cordova.

jthrilly:
@rebeccamadsen - once you are happy this works well for windows folks, I think this can be merged.

@wwqrd - did you check this out yet?

rebeccamadsen:
This works for windows! The missing command message is just a minor thing, agreed. I also second the idea that if we have additional issues, cross-env is worth considering. But for a one-off, I think this is sufficient.

PR #394, opened 2018-02-21
Fix/windows protocol path test

### LINKED ISSUE #367: protocol test dislikes windows
I'm pretty sure this is a windows only phenomenon. But one of the tests fails for me locally:

```
Summary of all failing tests
 FAIL  src\utils\protocol\__tests__\protocolPath.test.js
  ● protocolPath › Electron › Generates an asset path for the file

    expect(received).toEqual(expected)

    Expected value to equal:
      "tmp/mock/user/path/protocols/foo.canvas/protocol.json"
    Received:
      "tmp\\mock\\user\\path\\protocols\\foo.canvas\\protocol.json"

      at Object.it (src/utils/protocol/__tests__/protocolPath.test.js:18:7)
          at new Promise (<anonymous>)
      at Promise.resolve.then.el (node_modules/p-map/index.js:46:16)
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:188:7)
```

wwqrd:
Use node's path.join to use correct syntax on windows.

Not actually tested on Windows, but should work in theory.

Resolves #367

### DIFF
--- a/src/utils/protocol/__tests__/protocolPath.test.js
+++ b/src/utils/protocol/__tests__/protocolPath.test.js
@@ -1,5 +1,6 @@
 /* eslint-env jest */
 
+import path from 'path';
 import environments from '../../environments';
 import { getEnvironment } from '../../Environment';
 import protocolPath from '../protocolPath';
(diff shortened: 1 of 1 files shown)


wwqrd:
Hold on, that branch has some extra history on it.

wwqrd:
Ready to review.

PR #409, opened 2018-02-28
Toggle ASAR on in build settings

wwqrd:


bryfox:
This works for me (macos 10.13.3). I tested with `npm install; npm run dist:mac`. I see app.asar in the package, and protocol assets load correctly.

@wwqrd's test build `Network Canvas-4.0.0-alpha.3-mac-with-asar.zip` also runs correctly for me.

PR #410, opened 2018-02-28
Removing instances of transparentize to use css variables.

### LINKED ISSUE #309: Use CSS variables wherever possible to allow easy customization of the app
At present we define many variables for re-use throughout the app, using SASS, JavaScript - or both at the same time. Examples of this include our colors, our animation parameters, and some component sizes. This has a few effects:

- We require a build system to synchronize shared SASS and JavaScript variable values. 
- Once compiled, SASS mixins such as used by our palette and color system are harder to override. End users will need to write a lot of custom CSS, or else use the UI repo as a dependency for their own custom SASS based stylesheet.

Instead, I propose that we move to using CSS variables wherever possible. These can be set and read within javascript, and can be overridden with a simple custom stylesheet targeting only the CSS variables.

### Implementation details
- Global (project wide) variables will be implemented in the `:root` element, allowing for easy overriding.
- Components may create and use internally scoped variables by defining them within the block scope definition for that component (using our BEM system). I've done this for the spinner component to create size variations, so take a look for an example of why you might want to do this.
- We should modify continue to modify variable values based on media queries, such as breakpoints. For example, our `--node-base-size` is initially set at 200px, but changes to 160px for smaller screens.
- Colors: we should continue to use static SASS variables to define our network canvas named colors (although these too can be come CSS variables if we wish). Examples of these named colors are "network-canvas-coral", "mustard", "cyber-grape", and so on. However, the palette definitions (which provide a mapping of a color to a specific use within the app - such as "primary node color") will be transitioned to global CSS variables, and set to our existing color definitions as a default.
- Animation parameters (duration and easing) will be moved to global CSS variables. We can use `calc()` within our CSS to create derivative properties, such as `--animation-duration-slow: calc(var(--animation-duration-standard) * 0.66)`.

### Implementation tasks
- [x] Update color system in Network-Canvas-UI, including removing existing build/compilation system.
- [x] Update Network Canvas-UI components, and Network-Canvas components to use new color system.
- [x] Update animation definitions to use CSS variables, including removing the duplication present in the Javascript side.
- [x] Update animation implementation in UI and NC to use new CSS variables. There will be more work on the React side with this, since our react transition group stuff uses animation values directly.

jamieshark:
Partially fixes #309 

While all the other instances of `palette` and `color` will compile to using `var`, we need to replace the values of colors that use SCSS functions like `transparentize` in order to use the css variables as expected.

https://github.com/codaco/Network-Canvas-UI/pull/37 should be merged in before merging this PR in.

### DIFF
--- a/src/containers/ConcentricCircles/Radar.js
+++ b/src/containers/ConcentricCircles/Radar.js
@@ -4,7 +4,6 @@ import { range, last, zipWith } from 'lodash';
 import color from 'color';
 import getCSSVariable from '../../utils/CSSVariables';
 
-
 const equalByArea = (outerRadius, n) => {
   const rsq = outerRadius ** 2;
   const b = rsq / n;
@@ -23,7 +22,7 @@ const equalByIncrement = (outerRadius, n) =>
     .map(v => (v * outerRadius) / n)
     .reverse();
 
-// Weight towards a by factor
+// Weight towards `a` by factor
 const weightedAverage = (a, b, factor = 1) =>
   zipWith(a, b, (c, d) => ((c * factor) + d) / (1 + factor));
 
--- a/src/containers/ConcentricCircles/Radar.js
+++ b/src/containers/ConcentricCircles/Radar.js
@@ -1,8 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { range, last, zipWith } from 'lodash';
-import { colorDictionary } from 'network-canvas-ui';
 import color from 'color';
+import getCSSVariable from '../../utils/CSSVariables';
+
 
 const equalByArea = (outerRadius, n) => {
   const rsq = outerRadius ** 2;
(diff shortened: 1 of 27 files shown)


jthrilly:
@wwqrd I've added some changes to show how we might do this on the JS side. Would you mind having a look? Some notes:

- Where we use the CSS variable value directly, for example by passing it as a style prop to a component, changes to the variable value are also reflected in the component 👍 
- Where we process the variable value, as with `color()`, this obviously doesn't work. I don't see it as an issue, and certainly not one worthy of implementing some kind of data binding, but I wondered what you thought.

src/containers/NodePanels.js
```diff
@@ -2,26 +2,25 @@ import React, { PureComponent } from 'react';
...
-  colorDictionary['edge-alt-5'],
+  getCSSVariable('--edge-alt-1'),
+  getCSSVariable('--edge-alt-2'),
+  getCSSVariable('--edge-alt-3'),
+  getCSSVariable('--edge-alt-4'),
+  getCSSVariable('--edge-alt-5'),
 ];
 
 const getHighlight = (highlight, panelNumber) => {
-  if (colorDictionary[highlight]) { return colorDictionary[highlight]; }
```
jthrilly: I need to fix this so that custom highlight colors work again. I will probably change this prop to just be a color value rather than a pointer to a palette.

jthrilly:
@wwqrd - another issue: the test snapshots can't have access to CSS variable colors. Obviously I could update the snapshot, but it doesn't seem to be all that useful to have a lot of `#000000` values. Is there a way I can mock them you can think of?

src/containers/ConcentricCircles/Radar.js
```diff
@@ -31,12 +32,14 @@ const Radar = ({ n, skewed }) => {
...
 
-  const colorRing = color(colorDictionary.ring);
-  const colorBackground = color(colorDictionary.background);
+  const colorRing = color(getCSSVariable('--ring'));
+  const colorBackground = color(getCSSVariable('--background'));
 
   const ringFill = (ring) => {
     const mix = (ring + 1) / n;
-    return color(colorBackground).mix(colorRing, mix).hex();
+    const colorMix = color(colorBackground).mix(colorRing, mix);
```
wwqrd: I know this was in the original but looks like: `color(color(cssVariable))`?

wwqrd:
I've added a mock for getCSSVariables. The returned colour values are arbitrary, but it's all that's needed for this test.

PR #411, opened 2018-03-02
Feature/load router protocols

### LINKED ISSUE #326: Make factory protocols loadable via router
In order to speed up development work and make better use of hot-reloading we would like to be able to load a specific factory protocol using URLs in development, for example:

`http://localhost:3000/#/protocol/<protocol-name>`

Additionally we would like to be able to load a specific stage number:

`http://localhost:3000/#/protocol/<protocol-name>/<stage-number>`

rebeccamadsen:
Resolves #326.

To test, you can use the browser's url to navigate to e.g. `/protocol`, `/protocol/development.netcanvas`, and `/protocol/development.netcanvas/3`. The next/previous stage buttons and the stage menu all update the url as well.

If you change the code so live reload kicks in, your browser should stay on the same protocol and stage you were on.


bryfox:
Just clicking around, this is great... 

I do see a couple of console errors after loading the development protocol:

> Waning: Failed prop type: Invalid prop `component` of type `function` supplied to `SetupRequiredRoute`, expected `object`.

> Warning: Failed prop type: Invalid prop `component` of type `function` supplied to `LoadParamsRoute`, expected `object`.

A couple of usability things:

- when I first load a protocol, I'm taken to `#/protocol` instead of (e.g.) `#/protocol/development.netcanvas`
- when I click "Reset Session" (left sidebar menu) while on one of the inner protocol/:id/:stage routes, I'm left on that page with a blank view since there's no data. Could that redirect to /setup?

rebeccamadsen:
@bryfox: Thanks for the feedback! I've updated this to reflect your ideas. Any other suggestions are always welcome.

bryfox:
Those changes all look good. One last question, and it may be out of scope for this PR.

If I load the dev protocol and navigate to (say) Stage 2, then reload the page, I see the Welcome (rubberduck) screen show up briefly before the state settles on the correct state. It doesn't bother me, UX-wise, but looking at it a bit, I wonder if it's an indication of something that could be simplified...

Based on redux debugger output, the protocol loads (stage ID is 0); stage 0 renders briefly; then stage N is set and things settle.

stage-state

We're persisting stage.index / stage.id in the session state (prior to this PR), but since we also now have a location that defines the stage... could the stage state be derived from the location (so there's only a single point of truth)? Or at least be resolved as soon as the route matches?

rebeccamadsen:
This is an excellent question. I looked at trying to use the location only, and deriving the stage. However, react-router-redux dictates that you [should not read the location from redux](https://github.com/reactjs/react-router-redux#how-do-i-access-router-state-in-a-container-component). Passing the stage along with props works fine for the `ProtocolScreen`, but `StageMenu` also needs the stage info and that is a sibling of the parent of the router, so higher up the tree. Maybe that structure could be rearranged, but I'm not sure that is desirable either. So I settled for just moving any change/update to protocol/stage to the router's components to handle after the url is read. If there is a better solution to this, or something I overlooked, I'm open to discussion! Maybe Context? Maybe it is worth changing the `App` structure after all?

As for resolving it when the route matches, that is a simpler solution, I think. Right now, I'm just using an existing action to load the factory protocol, and the reducer just sets stage to 0 whenever a protocol is changed. We could have some other action in place for when we already know the stage, because at that point the router knows both. I considered that briefly and debated myself whether that was worth increasing the complexity a little. I figured it was for the developers, that users won't be refreshing, so opted not to add that action, but I could add it if it is worth it to us after all.

jthrilly:
> As for resolving it when the route matches, that is a simpler solution, I think. Right now, I'm just using an existing action to load the factory protocol, and the reducer just sets stage to 0 whenever a protocol is changed. We could have some other action in place for when we already know the stage, because at that point the router knows both.

I think this is the way to go. @wwqrd?

While is is a minor annoyance to have the flash of content now, it will create really bad UX if the stage transition gets heavier (which it will when we do the new navigation system + skip logic).

wwqrd:
I'm not sure if `withRouter()` (https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/withRouter.md) might help for certain components (to avoid direct redux access). But I don't think it would hurt to change the app structure if that would help?:

```
<Provider store={store}>
  <ConnectedRouter history={history}>
    <App>
      <Switch>
        ...
      </Switch>
    </App>
  </ConnectedRouter>
</Provider>,
```

My feeling is that `http://localhost:3000/#/protocol/development.netcanvas` should be a redirect route to `http://localhost:3000/#/protocol/development.netcanvas/1`? Sorry if this was discussed above, I'm finding this tricky to grep.

rebeccamadsen:
I updated based on discussion to use withRouter and a component restructure, along with NavLinks, to give one source of truth for stage index. Hopefully this is in line with #414. I believe we should be able to use the router url to help achieve session info, though it's quite possible I've overlooked something.

bryfox:
Looks like `react-router-redux` is missing from the package-lock changes.

Once that's checked in, this looks great. Sets the stage [sorry] for #414 nicely.

rebeccamadsen:
My `package-lock` always changes all the `mac` things to `win` things, so I'm in the habit of not committing that! I can commit it, if you want. Or someone else can commit a mac version.

bryfox:
I pushed a new lock file. OK with me to merge if others are done reviewing.

PR #412, opened 2018-03-06
Fix/hide hidden form fields

### LINKED ISSUE #405: Hidden node attributes shouldn't add a step to the full screen new alter form
Currently if a node form contains a hidden property, this will create an additional blank step on a full screen new node form. Potentially this could also be adding markup to the regular node form, too.

rebeccamadsen:
Fixes #405. 

Basically this collects all the hidden fields in a form and puts them on the first step in the wizard, but this time *actually* hidden. Any information the node needed from those hidden fields still gets submitted with the form. This also populates those fields with any "value" from the protocol form, although it doesn't run or return function values, so "timeCreated", for example, is just a string.

To verify, check the "New Node" form of the development protocol when aspect ratio is low enough to prompt the full screen form wizard. It should only have the three required field, for Name, Nickname, and Age, but "timeCreated" will still be part of the node attributes.

Depends on ui's fix/hide-hidden-form-fields branch to be merged. See codaco/Network-Canvas-UI#40.

### DIFF
--- a/src/containers/FormWizard.js
+++ b/src/containers/FormWizard.js
@@ -24,26 +24,33 @@ class FormWizard extends Component {
   }
 
   nextField = () => {
-    const count = this.props.fields.length;
+    const count = this.shownFields().length;
     this.setState({
       fieldIndex: (this.state.fieldIndex + 1 + count) % count,
     });
   };
 
   previousField = () => {
-    const count = this.props.fields.length;
+    const count = this.shownFields().length;
     this.setState({
       fieldIndex: (this.state.fieldIndex - 1 + count) % count,
     });
   };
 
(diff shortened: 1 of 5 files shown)


jthrilly:
I can't figure out why the test is failing on the CI (it isn't for me locally). Any thoughts?

rebeccamadsen:
Because it depends on the ui to update the css class. Is there a way to tell CI to rerun this test when the ui branch is merged?

jthrilly:
I actually merged the UI branch prior to this because I assumed the same (and the tests were passing locally). Any other thoughts? Maybe its just cached. I'll try one more time.

jthrilly:
Had to delete the cache to fix it! 😕

PR #417, opened 2018-03-07
Fixes #320

### LINKED ISSUE #320: Redesign 'submit and new' functionality
We have a feature whereby a user is able to create a new node without closing the node form. This was implemented as a button, with the text 'submit and new'. There are several limitations with the current approach.

- The button is large, white, and prominent. This creates confusion about the primary action of the new node form. 
- It is also positioned where a cancel button might be
- The implementation is not the neatest. Currently #321 is addressing some other design issues relating to the NodeForm. One aspect of the shortcomings of the current design is the way that the 'submit and new' functionality is built.

jamieshark:
Turns "submit and new" button into a toggle at the end of the node form.


rebeccamadsen:
My styles locally are all wonky lately, so I will assume the button isn't actually some big squarish shape and that the nodes aren't actually super tiny! 

But I do think the submit/new toggle should still be optional in the form like the button was. For example, having it on the "Edit node" form makes little sense to me. And if it is going to be in the "Edit node" form, then it should either be disabled or it should actually do something. Right now it just leaves the modal open if you select it when editing a node and then click submit.

It may make sense to have it vertically stacked with the submit button instead of to the side, but I guess then the question is whether it should be above or beneath? That is purely an opinion though and I'm fine if it's preferred in that location. Just wanted to put the question out there to see if that positioning was a conscious choice or just leftover from previously.

jthrilly:
Couple of CSS issues:

- Can we make the text in line? Is this an option for toggle switches in general? If not, can it be?
- While we are tweaking this control, can we ensure that there is some bottom margin between the component and the label when the label is not displayed inline?
- There seems to be an issue when placing a button in the same row as this input (see screenshot)

screen shot 2018-03-12 at 5 50 11 pm

jthrilly:
Just to add that I also agree with @rebeccamadsen's comments above - the actual implementation of this probably needs to be more graceful. The edit node form definitely shouldn't show this, which should be easily achieved if we make the presence of this functionality a configuration property of the form itself in protocol.js. Let me know if you want more information on this.

jamieshark:
so i've tried to normalize and fix some of the styles both for full screen and non-full scree modals.
screenshot 2018-03-14 12 14 55


I think if we are to add an option to have it inline, it would be cleaner if we implement that in the UI repo. I can implement it in this repo, but it might be cleaner if we have it as an option at the component level. Thoughts @jthrilly?

jthrilly:
Option at the component level in UI sounds good to me!

jamieshark:
@jthrilly - opened a PR here! https://github.com/codaco/Network-Canvas-UI/pull/45/commits

rebeccamadsen:
The updates look good to me!

PR #418, opened 2018-03-07
Removes isNumericOnly key from input props for non-text inputs.

### LINKED ISSUE #375: "isNumericOnly" warning
With the teaching protocol, the first time you open the edit node or add node form in Chrome we get a console warning:
```
Warning: React does not recognize the `isNumericOnly` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `isnumericonly` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in input (created by ToggleInput)
```
Most likely this is coming from one of the fields we are using in the teaching protocol that we haven't used elsewhere.

jamieshark:
Fixes #375

### DIFF
--- a/src/containers/Field.js
+++ b/src/containers/Field.js
@@ -90,6 +90,10 @@ export const makeRenderInput = (componentType) => {
       InputComponent = withOptionsFromSelector(InputComponent, optionsSelector);
     }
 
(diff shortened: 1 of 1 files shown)


rebeccamadsen:
Looks good to me! Although I think it is fixing #375, not 325, right?

PR #422, opened 2018-03-13
tweak scroll/drag

### LINKED ISSUE #408: Tweaks to scroll/drag algorithm to better match user expectations.
We have been over this behavior a few times recently (#224 #204 #246 #342) and made good progress. 

However, our algorithm still has two big issues:
- Dragging too quickly consistently causes the drag action not to fire. The solution is to drag more slowly, but users find this counter-intuitive.
- We do not adequately take into account the direction of the drag, meaning that clear sideways movements do not always trigger drag regardless of speed - which they should.

Changes:
- [x] Consider the angle of the touch first, and consider everything more than a few degrees off of the vertical axis as a potential drag.
- [x] Raise the threshold for assuming a scroll event on a fast gesture to allow faster gestures to resolve to drag events. Can we also weight this assumption by the gesture angle?

rebeccamadsen:
Resolves #408.

This is obviously pretty subjective whether it improves the scroll/drag experience. It limits scroll movement to only consider initially vertical motion, or a ratio of velocity/angle for NodeLists, and leaves a horizontal scroll parameter for Ordinal Bin.

### DIFF
--- a/src/behaviours/DragAndDrop/DragManager.js
+++ b/src/behaviours/DragAndDrop/DragManager.js
@@ -38,11 +38,11 @@ function moveAngle({ dx, dy }) {
 }
 
 function axisProximityFromAngle(angle) {
-  return Math.abs(Math.cos(2 * angle)) ** 2;
+  return Math.sin(angle) ** 2;
 }
 
 function determineMoveType(movement) {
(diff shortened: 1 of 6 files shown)


jthrilly:
This looks great! I think its a big improvement. There are some small glitches that I don't think are specific to this, so I'm going to merge this in.

PR #426, opened 2018-03-16
Feature/toggle dev tools

### LINKED ISSUE #403: Implement a way to toggle the dev tools with a keyboard shortcut during development
We sometimes need to ask users for a stack-trace to help with debugging, but our compiled app disables the debug console.

We should implement either a keyboard shortcut, menu item, or both, to allow the dev tools to be opened.

rebeccamadsen:
Resolves #403. 

This gives a Session menu item to toggle devtools, or the shortcut Ctrl+Shift+I or Cmd+Option+I.


bryfox:
I tested on a mac; looks good to me. The ctrl-shift-i on mac isn't needed, but doesn't hurt. I assume whatever `Cmd+Option+I` maps to on windows doesn't conflict with anything.

rebeccamadsen:
Good point. I don't think it maps to anything on windows, but just in case, we already have a function for checking the OS so I just made use of it. Thanks!

PR #436, opened 2018-03-30
Feature/implement skip logic

### LINKED ISSUE #276: Implement skip logic
## Purpose
Skip flow logic allows the researcher to define a logical statement that is to be evaluated at interview time that determines if a given interview stage should be shown or skipped.

We devised a system for defining these logical statements (see the wiki). We need to implement their evaluation within Network Canvas.

rebeccamadsen:
Resolves #276. I've put an interface in the development protocol to test against, that currently shows only if there are exactly three nodes with either the name "Dee" or as part of an existing edge.

This adds a submodule to the project for querying against the network at codaco/networkQuery. This means when you check out this branch you need to run `git submodule init`. I think you only need to run `git submodule update` if the submodule gets changed.

Assumption I've made: the top level (network) query is asking how many nodes exist in the network after the rules have all been run. This means, for example, that when an edge query is run and nodes are pruned that don't connect to qualified edges, the ultimate question is still how many _nodes_ are left in the query result. 

Outstanding questions: 
1. Architect uses ANY/NONE at the network level query, but EXISTS/NOT_EXISTS at the rule level query. However, ANY/NONE are not part of the predicate.js list, so I'm not sure whether this difference was intentional. Either the predicate file just needs to be updated, or Architect needs to be consistent.
2. Some fields need to be quoted, and some do not. I'm using "typeof" to check for numbers, but will there be object/array/etc node/edge values that need to be unquoted to match correctly in the future?
3. I think circleCI may need to be updated to use the submodule correctly. Where would this go? https://circleci.com/docs/1.0/configuration/#checkout

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "src/utils/networkQuery"]
+	path = src/utils/networkQuery
+	url = https://github.com/codaco/networkQuery
--- a/README.md
+++ b/README.md
@@ -57,8 +57,9 @@ node-gyp rebuild --target=1.8.2 --arch=x64 --dist-url=https://atom.io/download/e
 
 0. Install the correct node and npm versions.
 1. Clone this repsitory.
-2. Enter the directory where the repository is cloned, and install the project dependencies by typing `npm install`.
-3. Refer to the development tasks section below to learn how to test and build the app.
+2. Fetch submodules by typing `git submodule update --init`.
+3. Enter the directory where the repository is cloned, and install the project dependencies by typing `npm install`.
+4. Refer to the development tasks section below to learn how to test and build the app.
 
 ## Development Tasks
 
--- a/public/protocols/development.netcanvas/protocol.json
+++ b/public/protocols/development.netcanvas/protocol.json
@@ -944,7 +944,7 @@
               "options": {
                 "type": "person",
                 "attribute": "name",
-                "operator": "EQUAL",
+                "operator": "EXACTLY",
                 "value": "Dee"
               }
             },
--- a/public/protocols/development.netcanvas/protocol.json
+++ b/public/protocols/development.netcanvas/protocol.json
@@ -705,45 +705,6 @@
       "id": "namegenroster2",
       "type": "NameGeneratorList",
       "label": "NG Classmates",
-      "skipLogic": {
-        "action": "SHOW",
-        "operator": "EQUAL",
-        "value": "3",
-        "filter": {
-          "join": "OR",
-          "rules": [
-            {
-              "type": "alter",
-              "id": "15218241538622",
-              "options": {
-                "type": "person",
-                "attribute": "name",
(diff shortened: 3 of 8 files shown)


rebeccamadsen:
The 'typeof' checks can be avoidable...if we assume that numbers do not come inside quotes in the protocol. I've committed an update that avoids Function and makes this assumption; it works so long as the "value" is appropriately quoted or unquoted (e.g. strings in quotes, everything else not). That puts the onus on Architect (and any users creating their own protocol) to get the quotes correct, I think. Currently any values I insert in Architect get quotes, whether they are string or numbers.

I've also added ANY/NONE handling at the top level.

Should I add what the expected json for skip logic should look like [here](https://github.com/codaco/Network-Canvas/wiki/Network-Query-Builder), and that the query happens on the number of nodes?

Am I correct that we just need `git submodule init` on our project readme then?

bryfox:
> Am I correct that we just need git submodule init on our project readme then?

Others will need to run update the first time (as well as init). I'd recommend documenting `git submodule update --init`, which takes care of both (and works for future updates too).

wwqrd:
Ah now I see where `new Function()` came from (wiki). That is a feature we want to support, but only for advanced users writing their own protocols. In that case the configuration would be provided as a "functionizable" string in the config, e.g:

```
# protocol.json:
{
  ...
  stages: [
    {
      "id": "skippable screen",
      "label": "Skip or Show",
      "type": "Information",
      "title": "To skip or not to skip...",
      "skipLogic": {
        "filter": "return query.and([query.edges({ type: 'friend' }), query.alter({ type: 'person', attribute: 'age', operator: 'gt', value: 29 }),]);",
      },
    },
    ...
  ]
  ...
}
```

What we've got so far definitely the intended usage for the object config, but we should probably add support for the filter function as string variant too. Sorry for the confusion!

rebeccamadsen:
The string variant was added in d466c70, with the assumption that it is just at the `filter` level and that the top level query was still present. So whether this is ready depends on whether that assumption is correct. This looks like:
```
"skipLogic": {
        "action": "SHOW",
        "operator": "EXACTLY",
        "value": "3",
        "filter": "return query.or([query.edgeRule({ type: 'friend' }), query.alterRule({ type: 'person', attribute: 'age', operator: 'GREATER_THAN', value: 29 }), ]);"
 }
```

jthrilly:
> So whether this is ready depends on whether that assumption is correct.

This is a complicated issue, and my gut feeling is that our implementation is going to evolve. It is more important that we have _some form_ of skip logic in this release than the fully formed end product. 

I think the assumption that the custom string still sits below the query is ultimately going to turn out to be incorrect, because we will want to have a completely arbitrary function that perhaps doesn't even operate on the network object (an example would be a function that returns true if the device time is after 3pm - there are other more concrete examples, too). But for now. it seems a sensible place to start.

PR #441, opened 2018-04-06
Feature/node bin sort default

### LINKED ISSUE #357: nodeBinSortOrder should be an optional parameter with a sane default
Currently the Sociogram interface fails if no `nodeBinSortOrder` parameter is provided. 

Instead, the interface should use the `subject.type` and `subject.entity` configuration properties to look up the `labelProperty` for the given node type in the variable registry, and sort using that parameter by default.

rebeccamadsen:
Resolves #357. To test, try removing the `nodeBinSortOrder` from a Sociogram interface. It should still sort by "nickname", since that is the `displayVariable` in the variableRegistry. This can be changed to other fields as well. If there is no `displayVariable`, it doesn't sort (e.g. nodes are displayed in whatever order the network nodes array is returned).


rebeccamadsen:
Good point. I agree, and updated to handle that case.

Issue #415, opened 2018-03-06
Components that consume external data as a network must generate UUIDs for nodes


jthrilly:
Only makes sense to implement this once #413 has happened.

Concerning UUID issues (#397):

- Any component using external data (JSON or CSV) interpreted as a network, and requiring a UUID for each node, is responsible for creating this UUID by hashing the the node properties. We should develop a standard helper function for this purpose.
- The specific hash method is TBD, but look at: https://github.com/puleos/object-hash
- When a node is brought into the interview, this UUID will be transferred as with any other property, allowing direct and consistent comparison to the origin node in the external data in subsequent page loads.
- This approach will work for JSON data or CSV data, since ultimately both must be converted to an array of objects to work with the node lists, or the autocomplete roster. Parsing CSV etc, is out of scope for this issue.
- If hashing each time the data is loaded creates a performance issue, we can cache the hashes in redux, and use a hash of the entire dataset to check for changes on protocol load (or something similar). Where exactly this will go in redux is not certain.
- Alternatively, we can provide a way for the user to mark an externalData source as containing a property that should be used as a UID. Suggestions welcome on an API in protocol.js for this.

wwqrd:
We could probably use the nodejs built-ins for hashing: https://nodejs.org/api/crypto.html#crypto_class_hash

The rest of this sounds good to me!

bryfox:
To be clear, it sounds like @jthrilly is talking about hashing some stringified version of a JavaScript object, rather than the source data directly (correct?). So there does need to be a well-defined serialization step before using crypto/hash. I see pros & cons to that; just want to make sure we're on the same page.

bryfox:
Can you comment more on "Any component using external data"? Would every interface, for example, be responsible for generating this? Or could the loader (#413) just take care of this centrally?

bryfox:
> Alternatively, we can provide a way for the user to mark an externalData source as containing a property that should be used as a UID. Suggestions welcome on an API in protocol.js for this.

Can it be as simple as defining a primary key?

```
"externalData": {
    "clinics": {
      "primary_key": "clinic_id"
...
```

jthrilly:
> To be clear, it sounds like @jthrilly is talking about hashing some stringified version of a JavaScript object, rather than the source data directly (correct?). So there does need to be a well-defined serialization step before using crypto/hash. I see pros & cons to that; just want to make sure we're on the same page.

[...]

Issue #413, opened 2018-03-06
Refactor externalData in protocol


jthrilly:
# New approach for loading externalData
- Mirrors the way we load protocols: data can be loaded either locally (from a file in the asset folder of the protocol), or remotely (from a URL).
- Interface configuration objects will continue to contain a pointer to an externalData item, but when accessing and loading this data, interfaces will use a special loader method.
- Requesting the data via this method will return a promise, which will resolve with the contents of the file, or an error.
- In the case of remote assets, the component should display whatever loading state (spinner etc.) is appropriate during the wait for the promise to resolve/error.
- When requested for the first time, externalData with a remote URL is fetched with the result stored in app storage, with a filename the same as the key.
- On subsequent requests, externalData with a remote URL has its lastFetched and cache properties evaluated, with data refreshed if needed. This is done by the loading function, not by the component consuming the data.

``` JavaScript

// redux protocol.externalData (same as protocol.js but with additional state properties)
{
  "externalData": {
    "clinics": {
      "friendly_name": "Available Clinics", // Used in architect to help user identify external data items
      "isLoading": false, // Used for loading state
      "isLoaded": false, // Used in conjunction with cache for skipping http request
      "url": "https://somewhere/clinics.json", 
      "cache": 3600 // Time in seconds before needing to re-fetch data
    },
    "hospitals": { // Example of remote asset that has already been fetched.
      "friendly_name": "Other available hospitals",
      "isLoading": false,
      "isLoaded": true, 
      "url": "https://somewhere/hospitals.csv", // requesting this data will resolve to asset: "appData/hospitals.json" because this has already been downloaded
      "lastFetched": 2342233, 
      "cache": 3600
    },
    "school_class": {
      "friendly_name": "School dataset",
      "asset": "class.json", // A local asset
    }
  }
}

```

# Questions
- Is the above a sensible approach? Can it be improved?
- Where should `isLoading` and `isLoaded` go? I copied the above from the protocol in redux.
- Should we use a separate folder (data?) for these assets?

wwqrd:
This looks good to me.

Might not need a lastFetched key, we could probably get that from the file system, although it could make the logic a lot simpler.

I don't think the protocol object should really be manipulated, since it's a configuration object. So these properties should probably want to like elsewhere, say in the session data, perhaps in a parallel data structure.  (maybe the same goes for the top-level `isLoading` property).

```
"isLoading": false,
"isLoaded": true,
"lastFetched": 2342233, 
```

[...]

PR #443, opened 2018-04-11
Feature/display label defaults

### LINKED ISSUE #311: displayLabel may refer to a non-existent node field
Summary: It's possible for a node to display an undefined value if the protocol does not meet assumptions (specifically, that every node contains an attribute named by `displayLabel`). A prompt may include two nodes that don't share any common attributes.

Extracted from #305:

Given that currently:

- a prompt's nodes (networkNodesForPrompt) are driven from the prompt's additionalAttributes in the protocol definition,
- a protocol dataSource is associated with a prompt, not a stage (a network can consist of different node types), and
- no business logic prevents a protocol from re-using additionalAttributes between prompts
...a prompt's displayLabel property may refer to a field that doesn't exist on one or more nodes.

35745990-4061d560-0813-11e8-9931-9f7bd823b52f

@jthrilly says:
> I'm thinking we implement logic for when a specified label is undefined such as the following: (1) attempt to look up the label for the node type as defined in the variable registry, and failing that (2) use the first property, or failing that (3) use the node id.

rebeccamadsen:
This resolves #311 and (I believe) the last case for #305. It updates node labels so they are no longer hardcoded to display `node.nickname`. Instead, this has a three-fold fallback solution to attempt to use a node's type to lookup the `displayVariable` for that type in the `variableRegistry`. If that fails, it shows the first field in the node object, and if that fails it shows `node.id`.

This should address the situation that nodes of different types can end up in a node list together, and still show something intelligent on a node specific basis. 

To verify, experiment with prompts with different node types, giving them identical `additionalAttributes` and adding nodes on each. `NodeList`s should show both node types with something intelligent for each label.

### DIFF
--- a/public/protocols/development.netcanvas/protocol.json
+++ b/public/protocols/development.netcanvas/protocol.json
@@ -79,6 +79,7 @@
       },
       "venue": {
         "label": "Venue",
+        "displayVariable": "Abbreviated Name",
         "color": "kiwi",
         "iconVariant": "add-a-place",
         "variables": {
--- a/src/containers/ConcentricCircles/LayoutNode.js
+++ b/src/containers/ConcentricCircles/LayoutNode.js
@@ -1,12 +1,12 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
 import { compose } from 'recompose';
 import { Node } from 'network-canvas-ui';
+import { getNodeLabelFunction } from '../../selectors/interface';
 import { selectable } from '../../behaviours';
 import { DragSource } from '../../behaviours/DragAndDrop';
 import { NO_SCROLL } from '../../behaviours/DragAndDrop/DragManager';
-
-const label = node => node.nickname;
 
 const EnhancedNode = compose(
   DragSource,
@@ -16,12 +16,13 @@ const EnhancedNode = compose(
 class LayoutNode extends PureComponent {
   render() {
     const {
-      node,
-      selected,
-      linking,
       allowPositioning,
       allowSelect,
+      getLabel,
       layoutVariable,
+      linking,
+      node,
+      selected,
     } = this.props;
 
     const { x, y } = node[layoutVariable];
@@ -38,7 +39,7 @@ class LayoutNode extends PureComponent {
         style={styles}
       >
         <EnhancedNode
-          label={label(node)}
+          label={getLabel(node)}
           onSelected={this.props.onSelected}
           selected={selected}
           linking={linking}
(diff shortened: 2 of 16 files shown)


wwqrd:
Using a function to create labels on the fly adds a compositional feature that we no longer need, since display label is always displayVariable/first prop/id. I'm wondering if this would make sense as an addition to the network selectors instead - that way we could remove the label function entirely.

rebeccamadsen:
This is certainly possible, to move away from the function approach. Although in discussing this with @wwqrd it raises a few questions that I'm still mulling over. For example, we could modify the nodes in the reducer, or in the selector. Doing this in the reducer would have certain implications -- for example, downloading data would then have our "label" field. Is that a plus or a minus?

Complicating this idea is that we have external data which would need a similar massaging to map our "label" on to it. We could do this either in the reducer (when parsing the protocol) or the selector (when accessing the node providers). But [async external data](https://github.com/codaco/Network-Canvas/issues/413) is in the plans and I'm not entirely sure what implications that might have. 

And finally, we add a label as a prop and then all the node attributes using the spread operator. This means if a node already has a `label` field, the value from the data is used, not our calculated label (this is true even in master, not just in this branch). Is that an advantage to use a generic prop like "label" that users may be using already, or should we have a different prop name that would be less likely to conflict (e.g. something obnoxious like networkCanvasNodeDisplayLabel)? Or if we document props we are using in the wiki is that enough?

On a slightly related note, a similar thing happens if users have a "color" field on a given node. I think that is something we intended, but there are other props we use that could similarly be clobbered by existing fields in the data (e.g. "selected", "meta", "allowDrag", etc). This prop clobbering is perhaps related to our [uid discussion](https://github.com/codaco/Network-Canvas/issues/415), though likely should be a separate issue.

jthrilly:
Good to see how much thought you've put into this @rebeccamadsen! Lots of excellent questions.

> we could modify the nodes in the reducer, or in the selector. Doing this in the reducer would have certain implications -- for example, downloading data would then have our "label" field. Is that a plus or a minus?

Here was the thought process I went through with this: 

1) technically, the display label is a derived value. It _should_ be a synthesis of other variables that are already on the node.
2) but in the case of a custom display label (say a function that concatenated a nickname variable with a random number for some sort of control grouping) could add new "data". That would mean that the researcher might want to store that data when the network is exported. 
3) However, they could achieve this result by having a variable in a new node form that sets its value via a custom function, and setting _this_ variable to the display variable. 

To summarize, I think the calculated value of a display label should stay out of the exported data because it is either a combination of other properties, or can be stored in a different way if the user wishes.

> Complicating this idea is that we have external data which would need a similar massaging to map our "label" on to it. We could do this either in the reducer (when parsing the protocol) or the selector (when accessing the node providers). But async external data is in the plans and I'm not entirely sure what implications that might have.

External data is indeed complicated, and we haven't fully thought through it yet. However, I think the current logic still stands. So basically (see my note below for why the order of the following is altered), (1) look for a label property with a set name (like `label`) and use it, failing that (2) look at the node type property and use that to try to look up the display variable, then use it. If that fails, (3) use the first useful looking property, and if that fails, 
 (4) use the id (which will be required).

> And finally, we add a label as a prop and then all the node attributes using the spread operator. This means if a node already has a label field, the value from the data is used, not our calculated label (this is true even in master, not just in this branch). Is that an advantage to use a generic prop like "label" that users may be using already, or should we have a different prop name that would be less likely to conflict (e.g. something obnoxious like networkCanvasNodeDisplayLabel)? Or if we document props we are using in the wiki is that enough?

I think you've brought up an excellent point here, which helps to refine the idea a bit. The way I see it now is that the display variable setting in the node registry should function like an override. That is, by default we will look for a `label` property and use it, but if the user wishes to use a different property they can modify this behavior using the variable registry.

This will mean that wherever a label property is present it will be used, an we will let users decide if they wish to modify this behavior manually.

As for what this property is called, I've used label above because I think it is obvious. If you think a more specific property would be better then go for it. I don't think label is a variable name that researchers will use unintentionally, so I think it is fairly appropriate.

jthrilly:
> The algorithm you describe is what is implemented here -- 

Yep. But this wasn't the original intention. We have been very clever in updating our spec to match our behavior 😉 

> first useful looking property is questionable -- right now it is just taking the first property, but maybe it should take the first property that isn't 'id' or 'uid'.

I think the idea was to have something like the first text variable. I'm not sure how feasible this is. Perhaps if a variable of type text isn't available (either because there just isn't one, or because the data is from an external source and has no variable registry) that can be when we skip to the final fallback of using ID.

jthrilly:
> And that perhaps we should still consider documenting somewhere what props we are using that nodes could inadvertently override. Some of the props we throw on a node I could see a plausible reason for allowing users to override, like label and color and possibly selected. Some, not so much, like onSelected, meta, and allowDrag. Perhaps this should be a new issue to look at which fields/props might be problematic to allow users to override?

You brought a up a great point previously about this being related to the UID issue. In that issue I talked about the idea of private properties. I think a low tech solution to implementing that idea and avoiding collisions would be to have our special net canvas properties begin with an underscore (`_label`, `_allowDrag`, `_uuid`, `_meta`, etc.). We could then prevent variable names from beginning with an underscore.

rebeccamadsen:
Alright. I updated case 3 to look at the first variable in the registry of type 'text', and added tests.

I can add an issue regarding the possible collisions to discuss further.

bryfox:
Following up from #305... this is great!

PR #447, opened 2018-04-16
Avoid having to `window.require`

bryfox:
This is a PR onto the ui branch which targeted web and added `window.require`s for electron (needed to get things working with the new markdown dependency). 

Presumably we want to keep a single output target. `react-markdown` already includes a universal build, so this resolves the module to the UMD version via an alias.

- reverts 11a967d and 7af2bd3
- updates webpack config for dev & prod

If you want to rebase the other changes away in the ui branch & just cherry-pick the config change, 60190a2 is the commit you want.

Markdown rendering seems to work as I'd expect (testing Android and Electron).

### DIFF
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -87,7 +87,9 @@ module.exports = {
     // for React Native Web.
     extensions: ['.web.js', '.js', '.json', '.web.jsx', '.jsx'],
     alias: {
-
+      // Use the UMD version of React Markdown, so that we can keep building for
+      // electron-renderer and running other browsers
+      'react-markdown': require.resolve('react-markdown/umd/react-markdown.js'),
       // Support React Native Web
       // https://www.smashingmagazine.com/2016/08/a-glimpse-into-the-future-with-react-native-for-web/
       'react-native': 'react-native-web',
--- a/config/webpack.config.prod.js
+++ b/config/webpack.config.prod.js
@@ -82,7 +82,9 @@ module.exports = {
     // for React Native Web.
     extensions: ['.web.js', '.js', '.json', '.web.jsx', '.jsx'],
     alias: {
-
+      // Use the UMD version of React Markdown, so that we can keep building for
+      // electron-renderer and running other browsers
+      'react-markdown': require.resolve('react-markdown/umd/react-markdown.js'),
       // Support React Native Web
       // https://www.smashingmagazine.com/2016/08/a-glimpse-into-the-future-with-react-native-for-web/
       'react-native': 'react-native-web',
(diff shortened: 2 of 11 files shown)


jthrilly:
Ahah! Very interesting!

While this does work, and gets us back to status quo, I think what I wanted to discuss was the principles behind what we are doing with our webpack config. I am worried that what we currently do has some issues:

- Currently, we are using `require` inside conditional statements that check the environment. This works because the require call isn't executed in environments that don't have require, and because we tell webpack to assume we have it, it will leave those statements alone leading to a nice and lean bundle, and no build issues. Neat little trick. However, I think you can see why it is a bit of a hack. Conceptually, the approach is at odds with building for the electron-renderer target, because our code is written as though `require` is not available, yet we are exploiting webpack believing it _is_ available. I suppose the question is: **should we think of ourselves as a web bundle with node enhancements, or a node bundle with web fallbacks?**
- Using `require` inside of a conditional statements might be bad for [other reasons](https://eslint.org/docs/rules/global-require) (performance, standards compliance, readability. In particular, I'm not sure how having many nested require statements might be hurting performance or memory consumption. I don't know enough about node to say if there is a difference between requiring, say, `fs` 15 times within a conditional/function context vs. once at the top of the file.
- This last one might not be an issue, but I don't think we currently have a way to use a node module in a browser environment without creating an alias to a UMD version (if available). It isn't an issue now, but as the problems we have had here show, we are only ever a new dependency away from needing something we hadn't needed up until now.

Just keen to hear others perspectives really. If we like what we are currently doing, then I think Bryan's fix can go straight in and we can move on.

bryfox:
I definitely agree with your last point. Yes, this is a hack. But it's the same hack we're already using, so it's the least effort, and (IMO) preferable to conditional `window.require`s as an interim solution to get markdown rendering. 

With a single target ('web with node enhancements' or 'electron-renderer with web fallbacks'), we're going to be using one hack or another.

Longer term, I proposed the idea of multiple output targets, providing the right dependencies to the right targets, and getting rid of internal conditionals like that. That has two drawbacks: implementation effort, and coming up with a decent dev workflow. (With the current setup, I can run a single dev server and test in desktop, mobile & browser clients simultaneously.)

Node caches `require`d modules, so I wouldn't worry about memory or performance.

wwqrd:
I like the idea of separate targets too. Perhaps we could move our own branching code into a similar setup to react-markdown, and then import it as a module?

I prefer the UMD alias approach here because the intent is clearer.

jthrilly:
I agree that multiple webpack targets would be the most correct way of handling this (and actually I did initially try to do that, until I realized how much was involved) I just worry that the complexity in the actual code structure will outweigh the benefits. We can't mirror the way it works for an external node module.

Anyway, let's merge this for now. Thanks for the feedback :)

PR #449, opened 2018-04-18
Revert additional winow.require

bryfox:
This reverts the revert of window.require, matching state that's in master & fixing the failing test.


bryfox:
This undoes the change in ui branch leftover from earlier dependency work, and matches master, so I'm going to merge. Should fix the CI failure there.

PR #451, opened 2018-04-18
Node upgrade and UI 1.13.3 Compatibility

jthrilly:
This PR upgrades network canvas to use Node 8.11.1 and npm 5.8.0. Steve and I felt this would be advantageous because it includes some fixes for the way that npm handles the package-lock that might help us get more stability. Node 8.11.1 has performance and security fixes, too.

This PR also includes support for UI 1.13.3, which required a minor change to the way the the main stylesheet was being imported (so that it came before our component imports).


bryfox:
Looks good to me; the UI fixes work.

The Readme should probably get an update with the new version numbers. If there's an important difference between npm 5.6 and 5.8, we might want to highlight that since 5.6 is the default with 8.11.1 & (at least with my setup) 5.8 requires a manual upgrade step. (The only change from a 5.6 `npm install` for me is switching a transitive url from http to https, which may not be related to npm version.)

PR #452, opened 2018-04-18
Use default cordova dependency locations; update zeroconf

### LINKED ISSUE #448: Misplaced Cordova dependencies in package?
Returning to this project for the first time in awhile, I was unable to run any of the Cordova tasks (e.g., `cordova build android`); scripts would exit with unhandled promise rejection warnings and no real useful info. I ended up removing & re-adding the platforms, which fixed my issues. However, all of the cordova plugin dependencies now exist in both `dependencies` and `devDependencies` (where they had been moved to, from `dependencies`, in a previous commit).

I'm running cordova 8.0.0, which is latest. Was the move to devDependencies intentional? I can PR a package update that works with latest cordova, but I'm not sure if those plugins need to stay in devDeps or not.

bryfox:
This moves cordova plugins back to `dependencies` in package.json.

Steps taken to get to this state (with cordova@8.0.0): `rm -rf platforms plugins; cordova prepare`. I then manually deleted the devDependency entries.

As noted in the commit, I removed dev dependencies that didn't exist in config.xml, and so weren't installed by cordova: swift-support (a transitive dep) and 'device' (which doesn't seem to be used).

I also bumped the zeroconf version, which supposedly fixes the installation/Xcode problems that I ran into when the plugin was added.

Finally, I updated the Readme: `cordova prepare` seems a better alternative than `cordova platform add...` for new & re- installs, since it follows the versions defined in the package.

Resolves #448.


bryfox:
May want to hold on this; I can re-issue this after #451 (node & npm version updates) is merged.

PR #454, opened 2018-04-18
Fix/refresh protocol

### LINKED ISSUE #445: Refreshing an in progress interview in electron or cordova fails to load the protocol
If you load a remote protocol, begin an interview, and then refresh the app, Network Canvas fails to load the protocol.

rebeccamadsen:
Resolves #445. This uses a protocol type in the url so we know whether to use `loadProtocol` or `loadFactoryProtocol`.

To test, open a remote protocol in electron or cordova, and refresh the app. The protocol should reload the protocol and stage you were on.


src/containers/SessionMenu.js
```diff
@@ -175,15 +175,22 @@ class SessionMenu extends Component {
...
-    electron.remote.getCurrentWindow().on('blur', () => {
-      electron.remote.globalShortcut.unregister(accelerator);
-    });
+    electron.remote.getCurrentWindow().on('focus', register);
+    electron.remote.getCurrentWindow().on('blur', unregister);
+
+    window.onbeforeunload = () => {
+      unregister();
+      // remote listeners since this renderer window is going away
+      electron.remote.getCurrentWindow().removeListener('ready', register);
```
jthrilly: removeAllListeners();

PR #462, opened 2018-04-20
fix can't delete edges

### LINKED ISSUE #456: deleting an edge isn't working
Deleting an edge doesn't seem to work anymore, unless we changed the interaction when I wasn't paying attention. To replicate, create an edge between two nodes in the sociogram screen, and then try to delete it by clicking both nodes. The edge stays in the network object, and still shows on the screen. `TOGGLE_EDGE` is called. I can use the "reset" button to remove edges (this calls `REMOVE_EDGE` instead of toggle).

This may be related to #437?

rebeccamadsen:
This fixes #456, reverting TOGGLE_EDGE to add an edge if it doesn't exist already and remove it if it does exist. Also adds comments and updates the test.

### DIFF
--- a/src/ducks/modules/__tests__/network.test.js
+++ b/src/ducks/modules/__tests__/network.test.js
@@ -132,7 +132,7 @@ describe('session actions', () => {
     });
   });
 
-  it('should toggle edges', () => {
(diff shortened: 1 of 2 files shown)


bryfox:
LGTM. I confirmed toggle edge removes the edge the second time.

Issue #432, opened 2018-03-26
URL in text opens in same 'window' (i.e. within app)


berniehogan:
Clicking on a URL in the app sends the user to the URL inside the same window. This means that we escape Network Canvas. There's no 'back button' meaning you must thereafter restart Network Canvas. 

For this reason I propose that all URLs pop out or use the system default browser. 

This happened in electron / browser version / compiled version up to 4.0.3.

PR #467, opened 2018-04-20
update education protocol to use markdown

### LINKED ISSUE #453: Fix outstanding markup issues
Making sure this doesn't get lost (from @rebeccamadsen in #438):

> The education protocol is still using html in the information interfaces. That should be updated. Probably the wiki should be updated [here for sure](https://github.com/codaco/Network-Canvas/wiki/Information-Interface) and possibly [here](https://github.com/codaco/Network-Canvas/wiki/Name-Generator#interface-specific-api).

- [x] Convert markup in education protocol to markdown
  - Note that markup uses lists in a couple of places, which is currently disallowed in markdown settings; need to allow lists or rewrite
- [x] Update wiki with those examples

rebeccamadsen:
Adds links and lists to markdown for Information Interfaces. Updates the education protocol to use markdown and not html. Resolves #453 since Josh updated the wiki.

### DIFF
--- a/public/protocols/education.netcanvas/protocol.json
+++ b/public/protocols/education.netcanvas/protocol.json
@@ -340,7 +340,7 @@
       "type": "Information",
       "title": "Welcome to the Teaching Protocol",
       "items": [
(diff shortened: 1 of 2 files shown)


bryfox:
Seems fine to me. I don't know if link support was explicitly excluded (for security or UX), but we can address as part of #432 if we include it.

jthrilly:
Links are included here, but the issue with the link target is unverified (likely still exists). Tagging it for the next release now we have switched to markdown.

PR #473, opened 2018-04-24
move cleartimeout to fix nodelist on quick prompt changes

### LINKED ISSUE #305: Incorrect state of Nodes on prompt change (edge cases)

rebeccamadsen:
Fixes #305. I think this does actually fix the issue with node lists displaying the previous prompt's nodes.

### DIFF
--- a/src/components/NodeList.js
+++ b/src/components/NodeList.js
@@ -34,6 +34,8 @@ class NodeList extends Component {
   }
 
   componentWillReceiveProps(newProps) {
+    if (this.refreshTimer) { clearTimeout(this.refreshTimer); }
+
     // Don't update if nodes are the same
     if (isEqual(newProps.nodes, this.state.nodes)) {
       return;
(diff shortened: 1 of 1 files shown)


bryfox:
This works from my point of view 👍. I tested pretty thoroughly back at #313, and if there's still an edge case, I can't find it.

jamieshark:
The only other thing I might add is to clear the timeout on `componentWillUnmount` as well.

rebeccamadsen:
Good point @jamieshark. I've added that in as well.

jthrilly:
Looks good to me!

Issue #377, opened 2018-02-15
"playsinline" warning


rebeccamadsen:
On Chrome, the teaching protocol "Drawing Ties" screen gives this warning:
```
proxyConsole.js:56 Warning: Invalid DOM property `playsinline`. Did you mean `playsInline`?
    in video (at Video.js:7)
    in Video (created by mapProps(Video))
    in mapProps(Video) (created by lifecycle(mapProps(Video)))
```
I don't see this on Electron or Android.

wwqrd:
I was mistaken, looks like React 16 simply passes though unknown props, so technically this is working as it should.

PR #475, opened 2018-04-25
Fixes #431. Removes video controls, adds option to loop video via pro…

### LINKED ISSUE #431: video display controls interrupt flow
On the text display screen where we show video, there's no need to have the controls. In fact, because they have a full screen and download  button, the run the risk of either a. taking the respondent out of network canvas, or obscuring the screen. 

My proposed change is to have the video without the chrome and set on a loop by default.

jamieshark:
…tocol, fixes playInline syntax.


rebeccamadsen:
Does anyone know if we have had any problems with autoplaying videos with sound? Do we need [`muted`](https://developers.google.com/web/updates/2017/09/autoplay-policy-changes) in there?

We looked at [playsinline earlier](https://github.com/codaco/Network-Canvas/issues/377) and decided it wasn't necessary to change the case. But if we can verify React handles it and passes the prop along properly (was it ios that was the concern?) I'm okay with it.

jthrilly:
> Does anyone know if we have had any problems with autoplaying videos with sound? Do we need muted in there?

We haven't had an issue with it yet, but we also haven't tried a video with sound. I'd like to see a test video that should play with sound adding to the development protocol to verify this.

> We looked at playsinline earlier and decided it wasn't necessary to change the case. But if we can verify React handles it and passes the prop along properly (was it ios that was the concern?) I'm okay with it.

There is a lot of contradictory info out there about this. The HTML spec is playsinline, and when cased that way React leaves it along and just passes it through. There is some talk that React now supports passing playsInline, but I need to test it. It might require a version bump.

jthrilly:
Okay, I've added a test video that loops and has sound to test on devices.

jthrilly:
Alright, this works in iOS and Electron, but on Android I get a blank screen (but sound still plays!). It could be a video encoding issue. It also could be an emulator issue.

rebeccamadsen:
We've had issues with [android emulator not playing videos](https://github.com/codaco/Network-Canvas/issues/379) before, so that's most likely.

jthrilly:
Thanks Rebecca!

Looks like this is good to go, then.

PR #479, opened 2018-04-27
Launch of Alpha 4 - Gresley

jthrilly:
This release is codenamed Gresley named after [Nigel Gresley](https://en.wikipedia.org/wiki/Nigel_Gresley) the famous British steam locomotive engineer. 🚂 

As well as bumping config file versions, this PR also updates a lot base packages, including React, Redux, and Jest.


jthrilly:
I'm going to merge on the basis that the above looks to be a rendering glitch in electron, which we would expect to be resolved before 2.0 is released properly. It also seems to be non-blocking in terms of functionality.

PR #480, opened 2018-04-27
Improve dev workflows on devices & sims

bryfox:
This adds native support to the 'live-reload' dev workflow. Specifically: cordova & its plugins should all work; electron should work as expected (no shims).

See information in the readme, but the gist is that you start up a frontend and then a native app (on device or emulator) by providing one of `android`, `electron`, or `ios` as a platform: `npm run start:[platform]` + `npm run [platform]:dev`.

For example: `npm run start:ios`; `npm run ios:dev -- --target="iPad-Pro, 11.3"`.

Note I renamed 'run dev:[platform]' to 'run [platform]:dev' to be more consistent.

The caveat is that you specify one platform up front, and the content served is platform-specific. (My use case is being able to quickly develop features that rely on MDNS/zeroconf.)

Standard browser development (`npm start`) is unchanged.

Production builds are unchanged.

I marked as WIP for now; I want to work with this more before a merge is considered, but it seems to be doing well on macOS with simulators & external devices.


bryfox:
I added a caveat to the readme: you can't run multiple dev servers in order to (e.g.) run native android & ios builds at the same time, because the cordova build hooks rely on the `.devserver` file.

This was already the case, but you probably wouldn't have been tempted to try it until now.

My assumption/goal is to develop on a single platform at a time; cross-platform testing & QA can still rely on a build running index.html.

bryfox:
This has been working well for me, so I've removed the WIP designation.

I tested (only) electron on Windows, I don't have android tools installed there; hopefully @rebeccamadsen can take a look at some point. Note that you'll need to re-re-build the mdns dependencies (with or without this branch) since electron version recently changed. I [updated the readme](https://github.com/codaco/Network-Canvas/pull/480/files#diff-04c6e90faac2675aa89e2176d2eec7d8R43).

README.md
```diff
@@ -93,9 +101,24 @@ See Cordova's [iOS docs](http://cordova.apache.org/docs/en/latest/guide/platform
 
 ### Local Development & Testing
 
-If you have a running webpack dev server (started with `npm start`), you can run dev cordova builds on devices & emulators with live reloading.
+Starting the web app: If you have a running webpack dev server (started with `npm start`), you can run dev cordova builds on devices & emulators with live reloading.
+
+Starting a device or simulator: Run `npm run [android|ios]:dev`. This is a thin wrapper around `cordova run [android|ios]`; you can pass arguments to the cordova script with an extra pair of dashes. For example: `npm run dev:android -- --emulator`, or `npm run dev:ios -- --target="iPad-Pro, 11.3"`. Changes will be picked up from the dev server.
```
rebeccamadsen: The "for example..." part has the order of dev and android/ios flipped

config/nc-dev-utils.js
```diff
@@ -0,0 +1,83 @@
...
+// to serve (platform-specific) cordova dependencies to the iOS or Android client.
+// (Currently, you must make your choice at server startup time.)
+const isTargetingCordova = cordovaPlatformMatch.test(process.env.NC_DEV_PLATFORM);
+
+const isTargetingElectron = process.env.NC_DEV_PLATFORM === 'electron';
+
+const logPlatformInfo = () => {
+  if (isTargetingCordova) {
+    console.log(chalk.green('Targeting Cordova platform', process.env.NC_DEV_PLATFORM));
+    console.log(chalk.green('Live mobile dev available: `npm run', process.env.NC_DEV_PLATFORM, ':dev`'));
```
rebeccamadsen: I get this on console:
```
Live mobile dev available: `npm run android :dev`
```
Is the extra space after android coming from here?

jthrilly:
> INSTALL_FAILED_VERSION_DOWNGRADE sounds like the version it's trying to install is older than whatever's on your device or emulator. I rebased this on top of master this morning, which bumped the version... if you haven't pulled the latest, I'd bet that's what is wrong. (config.xml should say version="4.0.0-alpha.4").

In addition to this, android uses the build number in config.xml, which I have also bumped for the new release. This will be the cause of @rebeccamadsen's error.

rebeccamadsen:
My config.xml file was a little funky at times (yes, references to alpha.3). Maybe a cache thing of some sort?? I can't figure it why it was funky, but after switching branches and rebuilding I think this is working for me in android with live reloading!

Although the app doesn't update with changes to the protocol.json file like electron and chrome do. Is that an expected limitation of android or extraneous windows funkiness?

bryfox:
I imagine config.xml funkiness was a combination of the merged v4 changes and the [transient state it's in while building the dev version](#392). Unfortunately, I don't know of an easy way to fix that set of problems; maybe there's a better way to document.

> the app doesn't update with changes to the protocol.json file

Aha, great catch. So, this works by adjusting the dev server's 'content base' to the cordova www dir — the react app is loaded dynamically, and all of the requested [static] cordova assets are read from disk. In the non-cordova versions, the public directory is the 'content base', so changes to the protocol are seen. I just hadn't even considered the case for updating based on public file changes.

I'll see if there's an easy way to make this work...

rebeccamadsen:
If it's easy to add in, sure, but if not I am fine with considering this a limitation of developing with cordova. The rest of live reloading is already a big win as it stands.

bryfox:
Alright, adding a watch & copy step is easy. But as things stand now, we'd need to re-build the app, which means restarting from the setup screen, which is a wash for protocol development. (It may be possible to get this working with Android's Instant Run, but we're leaving the realm of easy/fast.)

I'll at least update the docs.

Issue #390, opened 2018-02-16
Global app settings


jthrilly:
The ability for the app to set and read settings, that persist between restarts and are independent of interview protocol.

Candidates for this sort of setting:

- Window state/position.
- Animation enabled/disabled.
- Assume device has physical keyboard

bryfox:
From #474, paired server information (currently persisted via redux, and resettable) may make sense to add here as well.

jthrilly:
This is implemented in the settings menu.

PR #488, opened 2018-05-04
WIP: Pairing

bryfox:
This begins the pairing work on the client side. Opening an initial PR for feedback & questions; do not merge.

Quick recap of testing setup:
- Check out the [feature/encrypted-pairing](https://github.com/codaco/Server/tree/feature/encrypted-pairing) on Server
- Start Server UI *on port 3000* (required for now): `npm start:electron`; then `npm electron:dev`
- Start Client: `npm start:electron`; then `npm electron:dev`

Summary of changes here:
- Move the discovery process behind an "add" button on the setup screen
- Implements the pairing endpoints described at https://github.com/codaco/Server/issues/41
- Adds protocol list & download from server (unencrypted)

enter-code

list


wwqrd:
This is from my reading of the pairing process for the server side, I haven't looked through the UI/Client aspects.

# Operation of handshake

As I understand it from a read through of the server:

1. client device requests to pair
1. pairing code + salt is generated on the server
    - salt + code saved to database
    - salt is shared with device
    - pairing code displayed on screen
1. User enters code from server screen into device
    - salt + pairing code = "secret" (`pwhash`) = cipher + nonce
1. device sends message using derived secret to cipher (`crypto_secretbox_open_easy`)
1. server decrypts message
1. if entry exists in database then treated as valid request.

 # Potential MITM attack vectors

Thought experiments, not verified:

  **fake server**
  1. relay what real server says, or make up your own response
  1. accept anything as valid
  1. receive confidential information from devices?

  **fake client**
  1. relay what real client sends
  1. you are now a client
  1. access protocols?

# General observations:
- Deterministic secrets meaning 64!-52! (?) possibilities when knowing salt. Not really feasible to brute force, especially with expirations on codes.
- Nonce should only be used once for the same cipher, in this case it's used twice? I don't know if this is a negligible problem or not.
- If the above MITM vectors are confirmable, then this pairing won't validate identities. However, in real world usage we intend to send this information over https, which would then make these attacks impossible because https would be verifying identity already (so long as we have a CERT authority), but then would we still need this handshake?

bryfox:
Thanks for reviewing. I'm responding to everything below; please continue to poke holes.

Your read-through steps are correct. I would add that the server then responds with an encrypted payload that includes a device ID; if the client is unable to decrypt this response using the secret key it generated, the pairing process fails.

So, quick recap:
- the current pairing protocol generates a shared secret key on a trusted client/server pair
- Trust is established by:
    1. a user selecting the LAN server to talk to, and
    2. entering an out-of-band password.
- Neither secret nor pairing code are transmitted in the clear. 

The secret can be used to encrypt messages, and as a means to authenticate a client, though currently the /protocols endpoint has no authentication. (See HTTPS discussion below.) The primary motivations are protecting privacy & authenticity of user data (sent from client) and external data (sent from server); and the authenticity of protocols (sent from server).

Two basic assumptions: out-of-band transfer can't be intercepted (e.g., by an attacker looking through an office window), and the secret is (at worst) only as good as the pairing code entropy for its lifetime.

## Potential attacks & weaknesses

### fake server

1: MITM (relaying). Beyond the first response, payloads are encrypted and require knowledge of the pairing code (transfered out-of-band). If assumptions hold (that code is unknowable and unguessable), then anything intercepted should be resistant to both tampering and decryption.

2: You (client) trusting the server you're talking to (impostor server). This is part (1) of 'Trust' above. The user must select the server to talk to from the discovery view.

Possibility A: Client sends pair request to spoof server. In this case, real server never displays pairing code. (If request is forwarded, then see relaying above.)

Possibility B: Client sends confirmation request to spoof server. Server cannot decrypt the message, and cannot respond with a valid message, since it has no knowledge of pairing code.

Currently, there is a minor problem here, as 'name' (from MDNS) is treated as unique on the client side, and easily spoofable. ('name' is currently the placeholder 'network-canvas'; that's an outstanding TODO. Based on the mockups, 'name' will be an IP address, though hostname would work.) This should disambiguate based on advertised IP as well. It gets a little tricky since a server can advertise multiple addresses. Could restrict to IPv4 to make this easier. This has been on my internal TODO list, but I'll file an issue.

I believe the risk here is denial of service, not tampering nor eavesdropping.

### fake client

1. MITM/Relay: What you'd receive is encrypted, and would not allow you to generate new requests to server. Requests/responses should be resistant to tampering & eavesdropping.

2. Replaying: Without knowledge of pairing code (or, later, the secret), the responses should be indecipherable.

### Deterministic secrets

Yes. If we assume the initial request with salt is intercepted (and we should), this all comes down to the pairing code entropy, which should basically be treated like a password. Currently, character space of 64 ** 12 characters. Re-pairing a device can be thought of as password rotation.

And as noted, pairing codes expire on use, or after 5 minutes automatically. Requests to pairing API are also throttled.

### Nonce used twice

Nonce shouldn't be re-used. It's generated anew in `encryptedBytes()`, which is called to generate the encrypted message to send. Can you provide more info?

### HTTPS

A CA-issued cert on a server would be preferable to (and remove the need for) this. But to be installation-specific, a user would need to purchase and install/configure a certificate.

There will be no mechanism (e.g., centralized server) available to distribute certificates. The only other means of getting a CA-signed cert (I think?) on Server is to publish it openly, private key and all.

Other options:
- Self-signed cert, installed by user on server & device(s). Still requires user effort. But if we're willing to require this, then I think could replace the proposed pairing process.
- Self-signed cert, generated automatically, with above pairing used to bootstrap trust.
- Self-signed cert, clients blindly trust. Makes eavesdropping somewhat harder, but still requires app-level encryption (from pairing key exchange), at least for anything sensitive.
- TLS without certificates (TLS-SRP?); little-no device support?
- Anything I'm missing?

jthrilly:
> Currently, there is a minor problem here, as 'name' (from MDNS) is treated as unique on the client side, and easily spoofable. ('name' is currently the placeholder 'network-canvas'; that's an outstanding TODO. Based on the mockups, 'name' will be an IP address, though hostname would work.) This should disambiguate based on advertised IP as well. 

I think I had actually intended for the card to show the hostname, with IP address also visible but much smaller/less emphasized. 

Multiple instances of Server on the same machine (and therefore IP address) is not something we want to support.

> It gets a little tricky since a server can advertise multiple addresses. Could restrict to IPv4 to make this easier. This has been on my internal TODO list, but I'll file an issue.

I'd rather not restrict to IPv4 if we can help it. I know it is basically a pipe-dream, but IPv6 has the potential to help a lot with network issues in corporate settings.

The way I understand this works is that you get a service announcement message for each independent route to the host. This could be across IP versions, or even because of different NICs and subnets that are routable between one another. What I imagined is that we would aggregate all these options in the serverList component, but only actually use the first one (for the time being).

jthrilly:
Just to add also that I don't think any workflows that require installing HTTPS certificates on devices are going to be viable, nor is distributing a certificate (since it would be a revokable violation). 

Most of our users are going to need to use a VPN to get access to their university network, and this will provide a degree of "transport" security (certainly from the most high-risk context - the open internet).  Once the traffic is internal to a university network, I think we should consider that a trusted, or at least low-risk, domain. Certainly I am not aware of other applications going to extensive lengths to secure themselves against eavesdropping in that context.

Perhaps this is work talking a bit more about as it seems to be the final outstanding issue.

wwqrd:
As you point out, the `/protocols` endpoint isn't encrypted at the moment, so I believe the fake device attack would still work? I can't say for the fake server, since this endpoint doesn't exist yet (as far as I know?).

mitm

wwqrd:
Nonce used for the second time here: https://github.com/codaco/Server/blob/feature/encrypted-pairing/src/main/server/devices/DeviceAPI.js#L372 probably the simplest way fix this would be send the nonce as public information with each message that uses the cipher, rather than using the secret bytes.

bryfox:
Here's my intent/understanding with nonce & key use:

- `encrypt()` takes a message and a secret key.
- The key is reused (many times) as the symmetric key for de/encryption, but is never sent in the clear
- That function calls through to `encryptBytes()`, and that's where [the nonce is generated](https://github.com/codaco/Server/blob/feature/encrypted-pairing/src/main/utils/cipher.js#L128)
- Nonce is sent in the clear with each message (prepended a couple of lines later)
- Decrypting is basically the inverse. [Nonce can be read directly from the front each message](https://github.com/codaco/Server/blob/feature/encrypted-pairing/src/main/utils/cipher.js#L88) since it's of a known, fixed length.

I think that's what you're getting at with 'nonce as public information', but let me know if I'm missing something else.

bryfox:
re: the MITM scenarios (diagrams). You're absolutely right; there's nothing in there yet to prevent anyone from requesting `/protocols` or downloading one at `/protocols/:name`. (Same, presumably, most URLs that will be entered into the direct download field.)

I should have been clearer about that. All we have here is key exchange (via pairing: https://github.com/codaco/Server/issues/41); we're not using the keys yet. Above, I was trying to work through possible weaknesses in the pairing process itself. I still *think* we're OK there, at least by what's shown in your flows?

Going forward: we'll use the shared key as a starting point for encrypting user data, and generally having trusted communications between client & server. *How* is TBD. We could use this to bootstrap TLS, or prove client/server authenticity of calls, encrypt every API req/res body (as above)... I think this depends on the outcome of some questions in the security doc.

bryfox:
I'll take another pass through to clarify it. Thanks!

jthrilly:
I just checked this out now - wow! It is fantastic to see the process working so well.

I took this to be a first pass of the UI, both on the server and the NC side, so I will hold off giving any specific feedback about rough edges. 

One thing I wondered is about the UX of the out of band code. The biggest problem we face is having a mixture of upper and lowercase letters along with numbers. That's going to be extremely tedious on a mobile device. Can you tell me how our security would be impacted by, for example, going to only a single case and only using letters?

bryfox:
Yeah, I kept the character space to (at most) an extra tap away on standard mobile keyboards, but assumed this would change.

tl;dr: should be fine as long as length is high. But I'm going to summarize since this should/will be added to a security doc anyway.

Space of possible passwords:
- Right now, char set is 64; ** 12 chars gives a password space of ~ 4e21
- 12 lowercase letters (only): 26 ** 12: ~ 9e16
- 16 lowercase letters (only): a little better than 12 from the larger char set
- All of these are randomly-chosen, per-pairing, so brute-force is what we worry about

A brute-force attack would more likely happen offline than against a running server. The key derivation config is static/deterministic, so (given possession of the pairing salt) one could reconstruct the secret. Thinking more about this:
- It's not a substitute for password strength, but we mitigate attempts with key stretching.
- Keys are generated with a hashing function; [config here](https://github.com/codaco/Server/blob/feature/encrypted-pairing/src/main/utils/cipher.js#L54-L57); [docs here](https://download.libsodium.org/doc/password_hashing/the_argon2i_function.html).
- This could be hardened (purposefully slowed down) by increasing the ops limit. But this time depends on the system it's running on, which we have no control over.
- Theoretically, we could generate a subsequent key with dynamic, device-specific configurations, which would make attacks that much harder

As noted, the attack above assumes that someone (who has LAN access) intercepted the initial pairing request. An attacker without access to the device salt would have to attack the entire keyspace on the running Server.

Further, assuming we use the key to immediately transfer a public cert from server to client, I think the client can still trust the server when exporting data.

Disclaimer: part of the purpose of this PR is to verify everything I just wrote!

bryfox:
@jthrilly re: UI, I'll have more refinements pushed up to NC soon. However, I'd love to hear feedback on Server at https://github.com/codaco/Server/pull/77. Other than items noted there, I'm sure how to best get that ready for alpha 1.

bryfox:
I'm removing the in-progress designation. There are no transitions/tweens between components, but all pairing functionality is still hidden behind the 'add' button on Setup.

Corresponding server branch: [feature/pairing-updates](https://github.com/codaco/Server/tree/feature/pairing-updates)

Latest additions/changes:

- Display hostname on server cards (https://github.com/codaco/Server/issues/78)
- Pairing via manual host/port entry (https://github.com/codaco/Server/issues/84)
- Friendly device name saved to redux & sent with pairing (https://github.com/codaco/Server/issues/73)
- Pairing code uses lowercase, alpha-only. (https://github.com/codaco/Server/issues/88)
- Handle loading & error states during pairing
- Pairing Code inputs (https://github.com/codaco/Network-Canvas/issues/492)
- Some general styling improvements around pairing & server components

Known issues:
- Pairing code set to 16 chars; we can lower if it seems overbearing
- Will conflict with https://github.com/codaco/Network-Canvas/pull/496

bryfox:
> The UI for key entry appears on NC device before the pairing request has been acknowledged on Server.

Yeah, I wasn't sure about this. I reasoned that the pairing code input appearing immediately 'told' the user they needed to shift attention to the Server GUI (which is probably on a different screen), as opposed to looking at a loading screen.

At any rate, adjusting this (and/or adding the more granular 'cancel' communication) is doable at the cost of chattiness in the protocol. This would also involve client polling, long-lived connections, or message-passing instead of restful API. I'll open an issue on Server.

> visual separation every four characters

Ah, right — you mentioned that. Good idea. I'll update here.

> Pairing seems to be triggered each time a client clicks on a server

Nothing is persisted yet. I assume that's the piece that depends on #390. Otherwise, I was thinking I could persist the server to redux for now.

> Server issues

Both good ideas, thanks — I'll open new issues on Server.

> Other than this

I just found another bug, in that friendly device name isn't surviving session resets. I'll also push a fix to this PR, if that's OK.

jthrilly:
Sounds great Bryan. Let me know when you want me to have another look.

bryfox:
@jthrilly I've pushed five more commits here, and one to server, to resolve #497. Please review again when you have a chance.

Some new test scenarios for this functionality:
- Begin pairing, wait for server timeout (30s)
- Begin pairing, dismiss on server
- Begin pairing, then close modal on client (which cancels request)

Should work for both discovery and manually entered server info.

bryfox:
Update (5/25): added one commit to prevent double form submits, and force-pushed to resolve conflicts with master.

bryfox:
Pushed a minor styling fix (ensuring margins so as to not overlap side menus) to fix layout issues seen on Android xhdpi 2048x1536.

Issue #490, opened 2018-05-08
Add confirmation prompt to reset menu item


jthrilly:
The menu item order is not ideal. Reset session is next to download data. There isn't an easy fix for this, and we don't want to invest time into it since the menu system is being revamped (#245). 

With this in mind, we need to implement a confirmation dialog for destructive actions, to ensure they aren't used by mistake.

Issue #245, opened 2017-09-28
New Menu System


jthrilly:
## Purpose

The current animation between interview stages is a simple fade out/in. It is functional, but could be improved.

We have feedback indicating that creating a better sense of space within the app would improve comprehension. Additionally, we have developed an interview overview in the Architect software that shows the interview in a top-down timeline. This seems to have been well received. Reconciling these two points, we came to the conclusion that we should experiment with a vertical scroll animation between stages, and a timeline component to visualize progress through the interview.

## Mockups
![timeline mockup](https://user-images.githubusercontent.com/1387940/31546223-48a3acae-b01a-11e7-87e4-4b9dfabab987.png)

## Specification

- A timeline should be rendered on the left of the screen, with a circle representing each stage in the interview. The timeline scrolls automatically as the interview progresses, but cannot be scrolled manually by the user. The top of the timeline scrolls off the top of the screen, but the bottom visually originates in the "next stage" button - until the end of the interview is reached, at which point the user sees the 'terminal' of the timeline.
- If an interview stage contains sub tasks (such as multiple prompts), the timeline should represent each as a segment of the circle for the current stage. As the user progresses through the prompts/tasks, these segments should fill in with the primary action color. 
- The next button should trigger forwards navigation through the interview. In a change from previous behavior, clicking "next" on a stage with multiple tasks/prompts will advance the prompt to the last prompt, before subsequently advancing the stage.
- Tapping/clicking on the timeline itself, or swiping in from the left edge of the screen, will open a secondary menu view that will allow the researcher to move to any specific stage.

jthrilly:
Michelle asks if the prompt swiper could/should also work vertically. She is concerned about the dissonance between the vertical movement of the menu and the horizontal prompts swipers. 

Suggestion: instead of the mini-dots, a pie chart that fills as progress happens through the stage.

jthrilly:
A note that this issue depends on #276 having been completed.

rebeccamadsen:
Is the mini-dots/pie chart thought with respect to the pips for prompts up above the prompt text, or in the side timeline? Is this a principle prototype, or just the image seen here in this issue? Is there a previous button in this design still?

pfsj:
1) The mini-dots/pie chart are supposed to represent the pips for prompts. Although they are still shown in this mockup, I'm not sure if the pips will make it into the final system if we have the same functionality on the new menu via the mini-dots/pie chart.

2) This is just a mockup image for this issue.

[...]

PR #491, opened 2018-05-08
Support copy/paste in Electron

### LINKED ISSUE #489: Copy and Paste Protocol URL
Copy and paste within the text box for the remote protocol URL doesn't work for Windows and Mac when running the compiled app (It does work on the iPad). This would be useful for testing custom protocols.

bryfox:
Mostly just pulling in the related change from Server here.

Including the built-in copy/paste menu items automatically adds keyboard shortcuts. This enables copy/paste across the entire app, which I assume is OK. The mac-specific handling is for the app menu item, as in [the electron docs](https://electronjs.org/docs/api/menu#main-process).

(Note that this also overrides Electron's default menu during development; if we were purposefully using different menus for dev/prod, then this may not be what we want.)

Fixes #489.

### DIFF
--- a/public/electron-starter.js
+++ b/public/electron-starter.js
@@ -1,4 +1,4 @@
-const electron = require('electron');
+const { app, BrowserWindow, Menu } = require('electron');
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
@@ -6,11 +6,6 @@ const url = require('url');
 const log = require('electron-log');
 const registerAssetsProtocol = require('./components/assetsProtocol').registerAssetsProtocol;
 require('./components/updater');
-
-// Module to control application life.
-const app = electron.app;
-// Module to create native browser window.
-const BrowserWindow = electron.BrowserWindow;
 
 const isMacOS = () => os.platform() === 'darwin';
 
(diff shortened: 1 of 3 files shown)


jthrilly:
Overriding the menu is fine, as we will need to start adding our own menu items soon anyway (potentially as part of a proper fix for #490 and when implementing #245).

I haven't checked this out, but can you tell me if "allowing copying and pasting anywhere" includes arbitrary text (paragraphs, headings, etc.)? One thing we specifically disallowed through CSS was arbitrary selection, due to UX concerns. As long as this respects that, I think you can merge it.

bryfox:
Yes, this respects `user-select: none`. I hadn't noticed that was in there. (Should we also set `cursor: default` so it doesn't look selectable?)

bryfox:
I just pushed another commit that does add reload & dev tools in the menu. Dev tools is already available from the sidebar menu, but this restores default behavior during development.

jthrilly:
Perfect.



> (Should we also set cursor: default so it doesn't look selectable?)

Yes, definitely. This would be a good improvement.

bryfox:
Sorry for piling on to this, but it's all kind of related:

- updated to `cursor: default` as noted above
- removed client setup of devtools shortcuts, since the shortcuts are now automatically provided by the native menu item

rebeccamadsen:
I thought we don't show the electron menu for the production version, which is why we had a devtools shortcut in our menu? Or did I miss something? I haven't tried this branch out yet, so may just not understand exactly.

bryfox:
Sorry, I think I was unclear in the description. This *will* now show the same menus in both dev & release builds. If that's not what we want I'll close this PR. ('Automatic' keyboard shortcuts won't work without the native menu items.) 

We could likely implement keyboard-shortcut-only paste with the [clipboard API](https://electronjs.org/docs/api/clipboard) and :focus selectors.

jthrilly:
I think for the time being, this is fine.

PR #496, opened 2018-05-15
Feature/multiple interviews

### LINKED ISSUE #414: Support multiple in-progress interviews
We need to be able to support resuming interviews, as well as starting a new interview while another one is in progress (irrespective of protocol). 

To facilitate this, we will need to change the structure of the redux store to look like the following:

``` JavaScript
{
  errors: {},
  menu: {},
  modals: {},
  protocol: {},
  search: {},
  sessions: {
    235235: { // an automatically generated session UID
      protocol: "demo.canvas", // A pointer to the protocol to be loaded
      network: [], 
      prompt: {}, 
      stage: {}
    } 
  },
  session:  235235, // A pointer to the currently loaded session to be used by other parts of the app.
}


```

The UI for resuming in progress interviews is present in #388.

rebeccamadsen:
Resolves #414.

This changes the structure of the redux store to:

```
{
  errors: {},
  menu: {},
  modals: {},
  protocol: {},
  search: {},
  sessions: {
    235235: { // an automatically generated session UID
      path: '/path/to/protocol',
      network: {},
      promptIndex: 0,
    } 
  },
  session:  235235, // A pointer to the currently loaded session to be used by other parts of the app.
}
```

Currently the buttons on the start screen just add a new session with that protocol  and a new network. To change to a previous session, you can use the session's path from redux to navigate. It will restore prompt in addition to the protocol and stage.

The url is the "truth", so if you use a session id and change the protocol or stage, redux will update the path that is saved for that session.

I commented out some experimental code in `SetupScreen` to switch sessions programatically. I will remove that shortly.

### DIFF
--- a/src/behaviours/withPrompt.js
+++ b/src/behaviours/withPrompt.js
@@ -21,13 +21,13 @@ export default function withPrompt(WrappedComponent) {
     }
 
     promptForward = () => {
-      this.props.updatePrompt(this.props.sessionId,
+      this.props.updatePrompt(
         (this.promptsCount() + this.props.promptIndex + 1) % this.promptsCount(),
       );
     }
 
     promptBackward = () => {
-      this.props.updatePrompt(this.props.sessionId,
+      this.props.updatePrompt(
         (this.promptsCount() + this.props.promptIndex - 1) % this.promptsCount(),
       );
     }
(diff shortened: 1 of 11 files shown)


wwqrd:
I wonder if the session action creators would make better thunks?

I'm not sure we need the flexibility of being able to specify session on a per-component basis, and I think this would be easier to maintain as `sessionId` can be kept in the sessions module:

e.g.
```jsx
const updateNode = (node, full = false) =>
  (dispatch, getState) => {
    const { session } = getState();

    dispatch({
      type: UPDATE_NODE,
      sessionId: session,
      node,
      full,
    });
  };
```

rebeccamadsen:
Updated to use thunks! Let me know any other feedback.

wwqrd:
Looks good to me 👍

PR #498, opened 2018-05-19
Render from LoadParamsRoute if resetting state

bryfox:
This updates the LoadParamsRoute to render if `shouldResetState` is true.

I'm not sure this is the correct fix, but it seems to solve the following problem on `master`. If I'm viewing a session/protocol and then "Reset Session" (from side menu), the app renders a blank screen (and empty menus), and I'm forced to quit & restart the app.

### DIFF
--- a/src/containers/LoadParamsRoute.js
+++ b/src/containers/LoadParamsRoute.js
@@ -84,8 +84,8 @@ class LoadParamsRoute extends Component {
       sessionId,
     } = this.props.computedMatch.params;
 
-    // not finished loading
-    if (!isProtocolLoaded || this.props.sessionId !== sessionId) { return null; }
+    const finishedLoading = isProtocolLoaded && this.props.sessionId === sessionId;
+    if (!shouldReset && !finishedLoading) { return null; }
 
     return (
       isSkipped ?


jthrilly:
This is odd. I didn't catch this, but presumably it snuck in with #414. I think @rebeccamadsen is implementing something that will fix this with the new start screen, where the "reset session" button will actually be removed in favor of a flow, whereby:

- The user has a menu option to "Return to start screen"
- Clicking this will prompt the user if they want to _discard_ the current in-progress interview data. Default action will be to retain.
- From the start screen, the user will either _resume_ an in-progress interview, or begin a fresh interview, which will create a fresh session.
- If the app pages out of memory, or crashes, or is otherwise restarted, I believe this system will let the app resume at the start screen, with the previous in-progress interview available to resume.

bryfox:
Side note: we don't yet have a specific 'unpair' flow. If 'reset session' is removed for the Greeley release, re-testing pairing will be more difficult (once paired servers are stored in Redux; part of #488). Do we need to add something else to cover this?

jthrilly:
Added #501 to cover this.

bryfox:
#501 is to make sure we have a way to reset the paired servers during this next round of alpha testing. If the "Reset Session" menu item is staying put, then we don't need to do anything.

However, @jthrilly mentioned above that the button would be removed. If that *is* going to happen in this next release, then I'll self-assign #501 and implement something similar, but only for the server state (e.g., "Reset Servers"). I'll let @jthrilly confirm.

jthrilly:
> However, @jthrilly mentioned above that the button would be removed. If that is going to happen in this next release, then I'll self-assign #501 and implement something similar, but only for the server state (e.g., "Reset Servers"). I'll let @jthrilly confirm.

Yes, this is more of a "global app reset", meaning that we hose all session data but crucially also all "meta" data concerning devices and app settings. This will be needed throughout alpha, and likely beta. @wwqrd actually found an issue where the redux state had changed so substantially between runs, that his app was unable to even open, which really emphasizes the need for this while things are so in flux. 

Come to think of it, the above means that a "reset app data" button is best implemented as a native menu item, to ensure it will still work if the main app fails to render.

bryfox:
OK, it sounds like #501 involves renaming to "Reset App Data", then. (I think we'll want it in the on-screen menu for mobile clients.) I can update #501 to add it to the native electron menu as well. I'll merge this in the meantime.

PR #502, opened 2018-05-23
Fix app reset

### LINKED ISSUE #501: Reset all app data menu item ("Global reset").
As discussed in #498, we need a global reset option for Network Canvas so that server pairing data, and global app settings, can be reset once the "reset session" menu item is removed.

This needs to be implemented within the current menu system until the new one is ready.

bryfox:
- Renames reset menu item in web UI
- Adds native reset (redirects instead of reloading to clear path info)
- Uses "Develop" Menu to match Architect

Resolves #501.

### DIFF
--- a/public/electron-starter.js
+++ b/public/electron-starter.js
@@ -1,4 +1,4 @@
-const { app, BrowserWindow, Menu, shell } = require('electron');
+const { app, BrowserWindow, dialog, Menu, shell } = require('electron');
 const fs = require('fs');
 const os = require('os');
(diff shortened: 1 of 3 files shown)


bryfox:
Note: I force-pushed with trivial conflict resolution.

@jthrilly are you OK with the label "Reset All Data"?

jthrilly:
Yep, seems reasonable. Thanks also for adding a dialog prompt for the native menu. Before we merge this, I am going to add the same for the in-app menu.

PR #504, opened 2018-05-23
Open content URLs in default browser

### LINKED ISSUE #432: URL in text opens in same 'window' (i.e. within app)

bryfox:
This adds:

1. a markdown link renderer to open anchor URLs in new windows. ReactMarkdown's link transformer already sanitizes things like `javascript:` URLs.
2. `new-window` handling so that URLs are opened in the user's default browser instead of a new electron window

Fixes #432.

### DIFF
--- a/public/electron-starter.js
+++ b/public/electron-starter.js
@@ -1,4 +1,4 @@
-const { app, BrowserWindow, Menu } = require('electron');
+const { app, BrowserWindow, Menu, shell } = require('electron');
 const fs = require('fs');
 const os = require('os');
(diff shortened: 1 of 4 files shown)


rebeccamadsen:
Looks okay to me!

PR #508, opened 2018-05-24
Interface/ordinal bin

jthrilly:
Description incoming tomorrow morning. Feel free to have a look in the meantime.


src/components/Node.js
```diff
@@ -0,0 +1,44 @@
...
+  /* eslint-disable react/no-unused-prop-types */
+  type: PropTypes.string.isRequired,
+  color: PropTypes.string,
+};
+
+Node.defaultProps = {
+  color: 'node-color-seq-1',
+};
+
+export default connect(mapStateToProps)(Node);
```
wwqrd: Since the component only serves to add defaultProps, you could refactor to:

```js
import React, { PureComponent } from 'react';
import { connect } from 'react-redux';
import { compose } from 'recompose';
import { makeGetNodeColor } from '../selectors/protocol';
import { Node } from '../ui/components';

const getNodeColor = makeGetNodeColor();

function mapStateToProps(state, props) {
  return {
    color: getNodeColor(state, props),
  };
}

export default compose(
  defaultProps({
    color: 'node-color-seq-1',
  }),
  connect(mapStateToProps),
)(Node);
```

src/components/Prompt.js
```diff
@@ -27,12 +27,11 @@ const Prompt = (props) => {
...
-      <h1 className="prompts__prompt-header">
-        <ReactMarkdown
-          source={label}
-          allowedTypes={TAGS}
-        />
-      </h1>
+      <ReactMarkdown
+        className="prompts__prompt-header"
+        source={label}
+        allowedTypes={TAGS}
```
wwqrd: Whilst we're changing this, what about the unusual case?

src/containers/ConcentricCircles/EdgeLayout.js
```diff
@@ -15,24 +13,11 @@ export class EdgeLayout extends PureComponent {
     displayEdges: [],
   };
 
-  renderEdge = ({ key, from, to }) => {
-    if (!from || !to) { return null; }
+  renderEdge = edge => (<Edge {...edge} />);
```
wwqrd: I'd be inclined to make `key` explicit, since it's this component (`EdgeLayout`) that makes it a requirement.

src/containers/Interfaces/OrdinalBin.js
```diff
@@ -0,0 +1,95 @@
...
+            prompt={prompt}
+            prompts={prompts}
+          />
+        </div>
+        <div className="ordinal-bin-interface__bucket">
+          <OrdinalBinBucket
+            nodes={nodesForPrompt}
+            listId={`${stage.id}_${prompt.id}_NODE_BUCKET`}
+            label={getLabel}
+            id={'NODE_BUCKET'}
```
wwqrd: `id="NODE_BUCKET"`

src/containers/OrdinalBins.js
```diff
@@ -0,0 +1,140 @@
...
+    toggleNodeAttributes: PropTypes.func.isRequired,
+  };
+
+  static defaultProps = {
+    isDragging: false,
+    meta: {},
+  };
+
+  renderOrdinalBin = (bin, index) => {
+    const stageId = this.props.stage.id;
```
wwqrd: Destructuring might make this function easier to read?

src/containers/OrdinalBins.js
```diff
@@ -0,0 +1,140 @@
...
+    const promptId = this.props.prompt.id;
+    const promptColor = getCSSVariableAsString(`--${this.props.prompt.color}`) ? color(getCSSVariableAsString(`--${this.props.prompt.color}`)) : color(getCSSVariableAsString('--ord-color-seq-1'));
+    const backgroundColor = color(getCSSVariableAsString('--background'));
+    const missingValue = bin.value < 0;
+    const accepts = () => true;
+
+    const calculateAccentColor = () => {
```
wwqrd: These methods should probably be moved outside of this function, currently they will be defined every time the render method is run.

I might be tempted to break this into it's own component actually.

wwqrd:
It looks great!

Maybe it could do with a way of removing the nodes from the categories?

(Required https://github.com/codaco/Network-Canvas/pull/509 to run it locally)

src/components/Node.js
jthrilly: I actually tried the recompose way, but couldn't get it to work. Your code here doesn't work either (it makes `selectable` crap out, along with other errors that I don't really understand. Perhaps we could take a look at this together at a later date?

wwqrd: Is it just `defaultProps` that's causing the problem? Sure we can look at that next week.

PR #518, opened 2018-05-30
Integrate server setup

bryfox:
This merges `master` (with conflicts resolved) to get the relevant server setup changes:

- SetupScreen is now in a Setup subdirectory; the pairing/import flows have several related components & containers
- ...but that's all behind the button and "protocol-import" route, so SetupScreen shouldn't need to know anything about ServerList.
- Looks like there's more we can clean up here (e.g., onClickLoadFactoryProtocol), but this gets the conflicts out of the way


rebeccamadsen:
This looks fine to me. Agreed that there is still cleanup to do. I think the ProtocolList and SessionList could also go into the Setup directory eventually too.

Issue #524, opened 2018-05-31
nodelist order keeps changing on scroll


rebeccamadsen:
Add a bunch of nodes (e.g. mock nodes) to a list. Note the visible names. Try to scroll that list a bit. When you stop scrolling, note the visible names. Often, they are the same as before, or maybe a few names are updated but many are the same.

Likely related, you could infinitely scroll the list and not reach the end.

wwqrd:
Sounds like something is going wonky with react-transition-group? keys?

PR #525, opened 2018-05-31
Skip sort if no criteria given

bryfox:
Fixes common case of #524.

If underlying sort isn't stable, and this isn't just a library issue, then this will need further refinement to prevent edge cases from cropping up.

### DIFF
--- a/src/components/NodeList.js
+++ b/src/components/NodeList.js
@@ -104,7 +104,10 @@ class NodeList extends Component {
 
     const styles = isHovering ? { backgroundColor: hoverColor } : {};
 
-    sorty(sortOrder, nodes);
+    if (sortOrder && sortOrder.length) {
+      // TODO: review; may need to use a stable sort
+      sorty(sortOrder, nodes);
+    }
 
     return (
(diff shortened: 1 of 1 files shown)


jthrilly:
What do you mean here by "stable" sort? As in deterministic? I think it might gave been simpler to provide a default empty object, which gets extended by the configuration object that is passed over?

bryfox:
> What do you mean here by "stable" sort? As in deterministic?

Yes; if two elements are seen as equal by the compare function, would they always appear in the same order? Looks like chrome doesn't use a stable sort.

I assume that's what was going on here; `sortOrder` provides empty criteria by default, so the resulting array can be ordered differently on each sort.

So there's still likely an issue for any 'equal' data. Example: user sorts on "age", and two nodes share the same age.

The larger question might be why the data is re-sorted continually while scrolling. Assuming render() needs to run that frequently, can the sort be done (once) outside render?

jthrilly:
Yeah, you're absolutely right. I think it has to do with the scroll container. I'll create an issue for review.

PR #526, opened 2018-06-01
Feature/new start screen

rebeccamadsen:
This is the new start screen, adding the ability to view and load previous sessions, and to rotate through available protocols. Downloaded protocols are added to the available list.

Note: this is the list of yet to be implemented items. We could hold off on this PR, or file separate issues to address these:
- animation for switching tabs
- drag sessions to trash instead of an "x" button (assuming this is still a desired feature)
- load initial protocol info from a file or some other sort of meta data -- Education and Development protocols are mocked in redux for now.

I also need to add some new tests, though existing tests all pass right now.


bryfox:
@rebeccamadsen FYI, I pushed a trivial change just to fix the lint error.

jthrilly:
I've created some new issues to handle implementing these in the next release.

> animation for switching tabs
#528

> drag sessions to trash instead of an "x" button (assuming this is still a desired feature)
#529

> load initial protocol info from a file or some other sort of meta data -- Education and Development protocols are mocked in redux for now.

PR #527, opened 2018-06-01
skip ordinal bin bucket sorting if no configuration object is provided

jthrilly:
Mirrors the fix provided here: #525 

This component should perhaps be refactored as a HOC of node list, but for now this will suffice.

### DIFF
--- a/src/containers/Interfaces/OrdinalBin.js
+++ b/src/containers/Interfaces/OrdinalBin.js
@@ -45,7 +45,7 @@ class OrdinalBin extends Component {
             id={'NODE_BUCKET'}
             itemType="EXISTING_NODE"
             onDrop={this.onDrop}
-            sortOOrder={prompt.bucketSortOrder}
+            sortOrder={prompt.bucketSortOrder}
           />
         </div>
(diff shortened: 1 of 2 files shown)


src/containers/Interfaces/OrdinalBin.js
```diff
@@ -45,7 +45,7 @@ class OrdinalBin extends Component {
             id={'NODE_BUCKET'}
             itemType="EXISTING_NODE"
             onDrop={this.onDrop}
-            sortOptions={prompt.bucketSortOrder}
+            sortOOrder={prompt.bucketSortOrder}
```
bryfox: Typo here

bryfox:
Got my branches mixed up; UI ref update matches the new-start-screen changes. 👍

src/containers/Interfaces/OrdinalBin.js
jthrilly: Ha! I think this is literally down to the train ride being bumpy.

bryfox:
I'm going to close this & update ordinal bin as part of #530.

PR #530, opened 2018-06-01
Sort nodes in list before render

### LINKED ISSUE #524: nodelist order keeps changing on scroll

bryfox:
This change to NodeList sorts the nodes every time the props change, rather than on each render pass. I think this still covers the intended use case, and arguably sort no longer needs to be stable.

If this seems OK, we could do something similar with the ordinal bin (see #527).

Follow-up to #525. Fixes #524.

### DIFF
--- a/src/components/NodeList.js
+++ b/src/components/NodeList.js
@@ -42,10 +42,16 @@ class NodeList extends Component {
       return;
     }
 
+    const newSortedNodes = newProps.nodes;
+
+    if (newProps.sortOrder && newProps.sortOrder.length) {
+      sorty(newProps.sortOrder, newSortedNodes);
+    }
+
     // if we provided the same id, then just update normally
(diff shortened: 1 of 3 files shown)


jthrilly:
Perfect! This is way better. I can't imagine how bad the performance could have been with a large node list.

Could you possibly add this change to the ordinal bin bucket component? I know it is bad that I didn't base that component off of node list. I'll rectify it later.

bryfox:
@jthrilly I updated ordinal bin sorting as well.

PR #534, opened 2018-06-04
Decode URI component for download urls for local paths

wwqrd:
Tiny fix to ensure downloaded urls generate URI decoded paths.

Is marked as WIP because this needs to be tested on devices,
which is waiting on PR https://github.com/codaco/Network-Canvas/pull/533

### DIFF
--- a/src/__mocks__/request-promise-native.js
+++ b/src/__mocks__/request-promise-native.js
@@ -1,0 +1,3 @@
+const request = ({ uri }) => Promise.resolve(`DUMMY DATA//${uri}`);
+
+module.exports = request;
--- a/src/utils/__mocks__/filesystem.js
+++ b/src/utils/__mocks__/filesystem.js
@@ -2,12 +2,14 @@
 /* eslint-env jest */
 
 const readFile = jest.fn(console.log);
+const writeFile = jest.fn(() => Promise.resolve());
 const ensurePathExists = jest.fn(console.log);
 const writeStream = jest.fn(console.log);
 const userDataPath = jest.fn(() => 'tmp/mock/user/path');
 
 export {
   readFile,
+  writeFile,
   ensurePathExists,
   writeStream,
   userDataPath,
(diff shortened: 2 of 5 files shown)


bryfox:
I tested "Import by URL" (with filename spaces) successfully on macOS & Android. However, iOS errors.

The plugin [converts target to an escaped URL](https://github.com/apache/cordova-plugin-file-transfer/blob/master/src/ios/CDVFileTransfer.m#L442-L449), but then later [fails this check against the unescaped string](https://github.com/apache/cordova-plugin-file-transfer/blob/master/src/ios/CDVFileTransfer.m#L461).

I suppose there are a couple of paths forward. Since this plugin is [deprecated in favor of xhr](https://github.com/apache/cordova-plugin-file-transfer#deprecated), might it make sense to switch to that now?

wwqrd:
From what you've described it sounds like an issue with the file system for cordova? Which probably means that we need to change how protocols are loaded too. I'm tempted to use %20 thoughtout.

jthrilly:
Another way of looking at this is that we don't actually have to preserve the original filename when storing the protocol internally. If this is causing a headache, we could rename them to something every platform supports? Not a plan of action, just a thought.

jthrilly:
Could even use the new UUID util to generate names?

wwqrd:
Generating local names would make a lot of sense, and help to avoid collisions between protocols. I'll update it to do that instead.

wwqrd:
Using a hash type file name is simple enough, but incorporating the actual protocol name into sessions seems complex.

I've found a way using a thunk, but it feels quick and dirty, as it depends on the protocol being loaded into state and updates the name each time session is updated - if anyone has an idea of a more appropriate placement I'd appreciate the feedback!

wwqrd:
Okay, having looked at this some more I think adding protocol titles to sessons probably falls within the scope of https://github.com/codaco/Network-Canvas/issues/545. So I'm going to request a merge with the file names still set as the protocol names.

Issue #472, opened 2018-04-24
Sanitize and/or custom-parse all user input


bryfox:
App should never evaluate user input (e.g., from a protocol file). Any input should be checked against a whitelist; any dynamic functionality should be come from validated & parsed input rather than evaluated directly.

This is especially true on desktop, where the current implementation requires node integration in the browser view.

jthrilly:
# Areas of particular concern
- Skip logic processing allows a stringified function to be evaluated. In a node environment. This is bad.
- Custom react interfaces can be built and imported. These ultimately get evaluated, which is also bad.

# Proposed solutions to make protocol files safer

- Remove node code to main thread, and communicate with node modules only via IPC calls. this is the priority, because arbitrary execution in a node environment could be very bad.
- Remove string eval for skip logic entirely, and depend on network querying. This functionality is somewhat speculative, and we can reintroduce it if/when we get more design input.

bryfox:
`webPreferences.nodeIntegration` in [browserWindow](https://electronjs.org/docs/api/browser-window) is the main mechanism for proposed solution 1 above.

There's another challenge with implementing this that I ran into on Server: Webpack's `electron-renderer` target assumes the presence of `require`. So in addition to removing node APIs from renderer code, I think we'd need to customize the webpack config further.

wwqrd:
Also worth noting from a conversation @jthrilly and I had: if we decided to drop support for Electron that would really limit this scope of this issue.

bryfox:
For some of the 'arbitrary functionality' use cases like skip logic, I also wonder if delegating to a web worker (without nodeIntegration) would solve most of the problems. Input data is contained; output data is limited to a well-formed object (or boolean...); the worker itself has limited exposure, but has the flexibility to do anything else a web worker can do. (For skip logic that needs tighter integration with the environment, we have the object interface in the protocol, which already sidesteps the eval problem.)

jthrilly:
> For some of the 'arbitrary functionality' use cases like skip logic, I also wonder if delegating to a web worker (without nodeIntegration) would solve most of the problems.

I have looked into this idea too, and I think it has a lot of potential. We can use the messaging mechanism to provide a boolean response value, and a timeout to avoid bad UX. We could also post limited interview data (the interview network?), although careful consideration would need to be made of potential leakages.

[...]

PR #535, opened 2018-06-04
Fix distribution build targets

### LINKED ISSUE #471: Exclude node_modules from Cordova builds
We need to find a way to exclude the node_modules directory from Cordova builds, because the mdns module contains binaries that are signed for MacOS, and this makes iTunes connect reject our bundle.

Might be another nudge toward separate build processes for Cordova and electron.

bryfox:
**Note: you may want to disable the "prepublish" setting when testing electron.**

This sets the production webpack build target appropriately (to 'web' or 'electron-renderer') based on environment, and updated build & dist scripts to npm to set this appropriately. https://github.com/codaco/Network-Canvas/commit/3cd13666384649c61215ec27e9d78c834d87040c has the meat of the changes.

I've updated the `dist` tasks to match what I understand is used currently to produce dist builds. *Distribution for any platform should now be done with* `npm run dist:[platform]`; for example, `npm run dist:mac` and `npm run dist:android` — this ensures the frontend is compiled appropriately. There's no top-level "dist" script since, as I understand it, there's no cross-signing support (e.g. signing a windows build on macOS).

Question: If I don't pass `--device` to the cordova build tasks, then I appear to get simulator builds. Is this intended? Are there other manual build steps taken in Xcode, for example?

Fixes #471. Also removes the need for #533.


bryfox:
> changing the webpack build target changed the way that require was treated, which meant that we needed code changes that were incompatible with one build target or another. 

It does change `require`, but this should mean we need fewer workarounds. I've been using the same thing in dev builds over the past month, but please do let me know if you see any issues.

> Are you saying when you do that, the simulator launches?

No, and I'm not sure if there's any difference on Android. But with iOS, at least, it targets a different platform:

```
$ cordova build ios
Building for iPhone X Simulator
Building project: /Users/bryfox/dev/network-canvas/Network-Canvas/platforms/ios/NetworkCanvas.xcworkspace
	Configuration: Debug
	Platform: emulator
```

```
$ cordova build ios --device
Building project: /Users/bryfox/dev/network-canvas/Network-Canvas/platforms/ios/NetworkCanvas.xcworkspace
	Configuration: Debug
	Platform: device
```

---

> such as the availability of a better alternative fo Math.random in the web crypto API, we should prefer a single function over an external dependency.

uuidv4 uses crypto as long as it's available, but I have no complaints either way.

jthrilly:
> It does change require, but this should mean we need fewer workarounds. I've been using the same thing in dev builds over the past month, but please do let me know if you see any issues.

It looks to be working brilliantly. So can you say a bit about how we should use require from now on?

For example:

- How does this impact our environment detection? Do `isElectron` `isCordova` etc still work?
- Do we need to use window.require to access the electron version of node at runtime still?
- Does this mean that redundant codepaths will be included in the bundle? For example, will the Cordova builds have browserified versions of node modules bundled, that are behind conditional checks (and so won't ever be called)? Or will webpack sort that out. 

> No, and I'm not sure if there's any difference on Android. But with iOS, at least, it targets a different platform:

Ah, yeah. It needs to be built for the generic device profile to be a releasable build.

> uuidv4 uses crypto as long as it's available, but I have no complaints either way.

Right, but now we have the browser crypto API, and as long as we don't _need_ strong randomness, this isn't actually a benefit.

bryfox:
> How does this impact our environment detection? Do `isElectron` `isCordova` etc still work?

Still works; webpack shouldn't be touching anything on window.

> Do we need to use window.require to access the electron version of node at runtime still?

Nothing really changes here:
- the `isElectron` fn obviously needs it since that path is executed by all environments
- Elsewhere, when [inside](https://github.com/codaco/Network-Canvas/blob/v4.0.0-alpha.4/src/utils/filesystem.js#L27) an env check, the reason to use it seems to be [our eslint config](https://github.com/codaco/Network-Canvas/blob/v4.0.0-alpha.4/src/utils/filesystem.js#L1). i.e., we never *needed* to use window (inside an env check), unless there's a scenario I'm not aware of.

We talked briefly at one point about (for example) providing separate filesystem abstractions for Electron & Cordova. We'd use webpack to target one and exclude the other. In that case, top-level requires/imports would be fine.

> Does this mean that redundant codepaths will be included in the bundle? For example, will the Cordova builds have browserified versions of node modules bundled, that are behind conditional checks (and so won't ever be called)? Or will webpack sort that out. 

Yes, I think you're right. If we're requiring libs and not using them in any code path, they're still in the bundle. Looks like 1.94MB gzipped here vs 1.79MB in master (though there are some other commits there).

- We could direct webpack to exclude modules we know won't be used (which is basically what we're doing by targeting electron)
- In the future, we could use webpack (as mentioned with "filesystem abstractions" above) to target the correct environment and separate both app & lib code
- We may actually want a custom target in the future anyway — but to swing us towards "web" and away from "electron-renderer" — for the reasons mentioned in https://github.com/codaco/Network-Canvas/issues/472. i.e., it would be good to isolate web views from node entirely.

jthrilly:
Thanks for the clarification about require, and the other stuff. I'm not sure I completely understand it, but this is definitely working well, so after updating the readme I think you can go ahead and merge.

> So should I add --device to the two cordova build tasks? Given the dev workflow, I don't think there's a reason to build for emulators, is there? Also, for that reason, I'm thinking I should remove the bare build task entirely.

No, I think having `--release` is enough. removing the bare build task is fine.

bryfox:
I decided to leave the default build task in (seems cleaner), but updated the Readme instructions.

jthrilly:
Just restarted the CI job as it failed due to networking issues. Assuming everything is good when it finishes, feel free to merge :)

PR #546, opened 2018-06-06
Feature/start screen tests

rebeccamadsen:
This fleshes out tests for new and updated code relating to the new start screen. It also fixes a bug relating to the prompt not reloading correctly when revisiting a previous session.


src/ducks/modules/__tests__/sessions.test.js
```diff
@@ -43,6 +44,7 @@ describe('sessions reducer', () => {
         path: 'path/to/session',
         network: { ego: {}, nodes: [], edges: [] },
         promptIndex: 0,
+        updatedAt: 1487076708000,
```
bryfox: This assertion could be reverted to `toEqual` if you want that specificity.

(I switched to `MatchObject` and ignored timestamps, thinking at some point we'd add assertions that timestamps are modified under expected conditions.)

The other ones below would match entirely as well except for the network nodes.

rebeccamadsen: Ah, I can switch that back to ignoring the timestamps if you prefer. I'm not particular about it.

rebeccamadsen:
Repo for the prompt bug:
Start a session (or return to a previous one). Navigate to a stage with multiple prompts. Navigate to a different prompt, other than the initial one. Return to the start screen. Resume that same session, and the stage *and prompt* should be restored. 

The bug was that the prompt was just going to the initial prompt instead of the saved prompt index, because of overeager updates to sessions.

src/ducks/modules/__tests__/sessions.test.js
jthrilly: I think its worth changing, and perhaps ignoring timestamps.

bryfox:
Thanks, I was able to repro on master & confirm the fix here. 👍

jthrilly:
Looks good to go whenever you're ready :)

PR #552, opened 2018-06-08
Feature/fifo sociogram bucket

### LINKED ISSUE #401: Teaching protocol: node bin sort order should be first in first out.
The sort order for the node bin on the sociogram interface should be changed to "first in, first out".

rebeccamadsen:
Resolves #401. This allows the sociogram nodebin bucket to have a first in, first out sort order, using the "*" to specify it and the nodes' positions in the array to sort on. 

This also updates the education protocol to use this sort.

If this is alright, I could also update the wiki.


jthrilly:
Do you think we could generalize this functionality so that it works for every node container that is sortable? So that would include node lists, too. The reason I ask is that I would prefer we not have multiple different approaches to sorting. 

@wwqrd - I think you had an idea about how to achieve this using loadash rather than a library. Do you mind saying more about what it is here?

jthrilly:
It just occurred to me that Steve is also changing the API for sorting a little bit (I believe). This could be an opportunity to update that, too.

rebeccamadsen:
It sounded in slack like Steve was in favor of merging this, and then he would build off this with some additional updates. But I could have the timing off on which comment was earlier. I'm happy to take a look at generalizing this, but don't want to overlap too heavily with ongoing efforts by Steve.

wwqrd:
Yep I think merge this in, I haven't started work on this yet and I think it'll be easier to manage with less moving pieces.

jthrilly:
Okay, let's merge it then. But we must make sure that when we update the other components with node sorting that they inherit this functionality.

PR #554, opened 2018-06-15
Fix/graphml export updates

### LINKED ISSUE #406: GraphML exporting doesn't work for the teaching protocol
Numerous reports that exporting as graphML doesn't work when using the teaching protocol. This functionality needs a review, since it has likely become out of sync since we implemented the variable registry and changed several APIs.

rebeccamadsen:
Fixes #406. This adds a check for variable type in the variableRegistry first. Variables with type 'layout' are given two keys with type 'double' -- in the form [variable]X and [variable]Y. The types use a typeof check if there are no specifications in the variableRegistry.

To verify, before this, any location variables were stored as a string. Now they should be two separate variables in the graphML, one for "x" and one for "y" locations. Also, if you have a numeric value in a text field, prior to this it would have been classified as a number (either "double" or "int"). But here, if the type in the variableRegistry is "text" then its key will be a "string" regardless of typeof values. Conversely, some form data is stored as "string"s even if they are numeric. This checks values of the fields for numeric content.

Other known types may need to be added in the future.


src/containers/ConcentricCircles/NodeLayout.js
```diff
@@ -30,7 +30,7 @@ const dropHandlers = compose(
     onDrop: props => (item) => {
       props.updateNode({
         uid: item.meta.uid,
-        [props.layoutVariable]: relativeCoords(props, item),
+        [props.layoutVariable]: { type: 'layout', ...relativeCoords(props, item) },
```
jthrilly: I think there might have been some confusion about this part. Basically, I don't want to "pollute" the node properties with variable type info. The ExportData methods shouldn't need additional properties on the nodes. It should just be able to look at the variable name, look up its type in the variable registry, and when it finds that it is a layout variable parse it in the appropriate way.

src/utils/ExportData.js
```diff
@@ -17,7 +19,45 @@ const setUpXml = () => {
...
 
-const generateKeys = (graph, graphML, elements, type, excludeList) => {
+const getVariableInfo = (variableRegistry, type, element, key) => (
+  variableRegistry[type] && variableRegistry[type][element.type] &&
+  variableRegistry[type][element.type].variables &&
+  variableRegistry[type][element.type].variables[key]
+);
+
+const isVariableRegistryExists = (variableRegistry, type, element, key) => {
+  const knownTypes = ['boolean', 'text', 'number', 'categorical', 'ordinal', 'checkbox',
```
jthrilly: I think we need to be careful with this list. I realize a lot of these 'types' were present previously, but the list needs to be revised now that we have a new system that allows us to differentiate between variable types and input components. So, specifically:

- checkbox is an input type - should be removed. It is covered by the boolean type
- enumerable is another word for ordinal
- options is another word for categorical (at least, they refer to the same thing)

src/utils/ExportData.js
```diff
@@ -53,6 +104,10 @@ const generateKeys = (graph, graphML, elements, type, excludeList) => {
               keyElement.setAttribute('attr.type', 'string');
             }
             break;
+          case 'text':
+          case 'categorical':
+          case 'enumerable':
```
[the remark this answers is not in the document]
jthrilly: See comment above about removing this.

src/utils/ExportData.js
```diff
@@ -179,19 +235,19 @@ const saveFile = (data, openErrorDialog) => {
...
+  generateKeys(graph, graphML, networkData.nodes, 'node', ['id'], variableRegistry);
+  generateKeys(graph, graphML, networkData.edges, 'edge', ['from', 'to', 'id'], variableRegistry);
 
   // add nodes and edges to graph
-  addElements(graph, graphML.namespaceURI, networkData.nodes, 'node', ['id']);
-  addElements(graph, graphML.namespaceURI, networkData.edges, 'edge', ['from', 'to', 'id'], true);
+  addElements(graph, graphML.namespaceURI, networkData.nodes, 'node', ['id'], variableRegistry);
+  addElements(graph, graphML.namespaceURI, networkData.edges, 'edge', ['from', 'to', 'id'], variableRegistry, true);
 
   saveFile(xmlToString(xml), openErrorDialog);
```
jthrilly: I know this is out of scope for this PR, but while we are updating this, I would really like it if you could move the file handling code out of this method and into a utility function elsewhere. This would be in preparation for other export functionality being able to reuse the file handling/writing.

So perhaps create methods in `./utils` for writing files that continue to use system dialogs on electron and share on Cordova. The interface for these can then handle, via callbacks or whatever, integration with the dialog system.

Then the actual file encoding methods here would just return strings that would then be passed to these file handling methods.

src/containers/ConcentricCircles/NodeLayout.js
rebeccamadsen: So an object is only a 'layout' object then if it is in the variableRegistry? Or should variables with type "object" be checked for an "x" and a "y"? Or certain wording in the variable name? 

This was an attempt to catch layout variables we were creating for the sociogram interface based on the protocol, but that weren't in the registry. I knew at that point in the code those were layout variables, but I could also just update the variableRegistry for both protocols.

src/utils/ExportData.js
rebeccamadsen: I can update the existing protocols. So a "checkbox" variable that isn't in the registry is just handled as a string? Or should the registry be updated to have the "boolean" type for that variable? Enumerable and options were lingering in the development protocol, but I can update those.

src/utils/ExportData.js
rebeccamadsen: I'll take a look at this and some testing as well.

src/containers/ConcentricCircles/NodeLayout.js
wwqrd: Sorry! It wasn't made clear before, `{ type: 'layout' }` is what has been added to the variable registry, layout variables can cross referenced against that:

```
variableRegistry: {
  node:
    person:
      variables: {
        "closeness": {
            "label": "Closeness layout",
            "description": "More detailed description",
            "type": "layout"
          },
        }
     }
  }
}
```

src/utils/ExportData.js
jthrilly: In the case of the factory protocols, I've updated them in a forthcoming PR addressing inputs. I have also created the following wiki page which attempts to map the basic variable types and input components that cover them: https://github.com/codaco/Network-Canvas/wiki/Variable-Types

jthrilly: If a variable is encountered that doesn't have a corresponding definition in the variable registry, I think that should probably be a hard failure, since it indicates an invalid protocol. When we are closer to a final release we might be able to relax that, with a fallback of just treating it as text.

rebeccamadsen:
This adds a test, refines the expected types, and separates the code a bit.

jthrilly:
> Layout variables: I tested by adding "closenessLayout": { "type": "layout" } to the development protocol. I don't know exactly what the requirements are, but it generates two double attrs instead of a string. Should we add that definition to the factory protocols now?

Yes, the factory protocols will need to have layout variables updated to `type: "layout"`. Perhaps do that here quickly, before merging.

@pfsj - can you verify that the exported data loads the layout correctly in Gephi?

rebeccamadsen:
I can update the factory protocols here; I thought that was happening in Josh's inputs branch. I commented out the hard fail until the factory protocols were "correct". They currently don't have all the variable in the variable registry, so would both fail.

Any "layout" will generate two variable of type "double" with the labels appending "X" and "Y" after the variable name (e.g. closenessLayout -> closenessLayoutX and closenessLayoutY). I'm not familiar enough with Gephi to know how to make those variables load as positions, but I can see the values in the data tables in Gephi.

bryfox:
Latest changes look good. I tested development & teaching protocols, see the closenessLayout additions, and confirmed the export error (when editing to have invalid variable types). I poked around a little in gephi; with the [GeoLayout plugin](https://gephi.org/plugins/#/plugin/geolayout-plugin), I was able to mimic my sociogram layout:

gephi

I think this is good to merge; if there are more specific requirements, we should follow up with additional issues.

jthrilly:
The Geo layout plugin is designed for lat/long style geospatial coords, which aren't appropriate for this use case I'm afraid. The layout should actually be picked up by default, without plugins. If that's not the case I'm afraid we need to investigate further.

rebeccamadsen:
I'm curious about the "picked up by default". We had originally come up with the layoutX/layoutY designation because more than one variable of type "layout" may exist. If there are more than one variables of type "layout", are we just to pick the first? Or is that an error?

jthrilly:
I think we need to pick the first by default, yes. Even duplicating this variable and relabelling as X, Y labels. I know this seems like an anti pattern, but we want any additional processing in terms of layout to he explicitly done by users.

rebeccamadsen:
The latest adds two fields, x and y, cooresponding to the first layout variable listed in the registry. Gephi reverses the high "y" value from javascript, so I'm using `1-[layout].y` here, and multiplying by the screen size to make the layout match better in Gephi. This leaves the layoutX and layoutY variables as we had them before for users to have the original data.

jthrilly:
One last change - can we add some verbosity to the error messages? I was trying to debug a protocol yesterday, and was just getting the generic "Error exporting data" response. It looks like it should be possible to add descriptive text to the callback which opens the dialog (and using the `additionalInfo` prop, or whatever it is called, to show this in the dialog itself). 

Nothing too detailed - just something like, "Variable 'X' does not seem to exist in the variable registry", and so on.

rebeccamadsen:
I added detailed information in the addionalInformation field. Let me know if there's any other request!

Issue #108, opened 2016-05-19
v4 tracker


jthrilly:
- Variable "registry"
  - Means to define and track variables. Contrain types, provide data validation functions, etc.
  - Could also facilitate generation of dummy data
- Node and edge types fully abstracted
  - verbs for edges
  - clearly defined distinctive styles for each type, with sane defaults.

jthrilly:
- Digital signing
- Iconography

jthrilly:
- JS Docs

jthrilly:
- keyboard (or external clicker) session navigation, and alternative user input more generally.

jthrilly:
- protocol custom modules
- class based module extension

jthrilly:
- Self updating. How feasible? How should this work with mobile apps?
- Mobile app publishing. Through app stores? Sideloaded? How will mobile apps load protocol files (specify protocol as git repo?)

jthrilly:
- Where should skip logic live? In the stage (stage.skipMe())? In the protocol? Proposal: each module contains its own skip logic. No need for confusing additional section in protocol file. Can simply be another parameter of the stage initialisation properties.

jthrilly:
- Nodes as 'reactive components', that include data and event bindings to underlying model. For example, change in underlying data model for a node (name field updated) results in UI update for any visible component.

jthrilly:
- Protocols should be able to define data 'keys' (related to variable schema?) that can be queried by other parts of the app. For example, skip logic could compare node names to a list of pre-populated names, or a series of name generators could have their autocomplete populated by a single list.

jthrilly:
- How do we handle dependency between variables? For example, Interface A creates a variable, which makes a given node meet the criteria for being shown on interface B. Participant then returns to interface A, and changes the value of the variable. This should also trigger the invalidation of the variable set by interface B.

jthrilly:
- Standardise interactions for specific types of action within the app. For example when do you tap vs double tap vs long press, vs, drag. (also relates to establishing variables for easing functions and timings)

jthrilly:
- Optional branding
- Unassisted interviewing and tutorial mode for all interfaces (intro.js)

jthrilly:
- Video tutorials/splash screens to introduce research questions or tasks.

jthrilly:
- how should we handle platform specific node modules? Do we need an installer?

jthrilly:
[...]

PR #555, opened 2018-06-19
Update Network Canvas to handle new orderBy syntax in protocols

wwqrd:
This adds a new sort generator in utils/sortBy, usage:

```
const sorter = sortBy([{property: 'name', direction: 'asc'}]);
const sortedItems = sorter(items);
```

The app has been updated to use this sort object everywhere orderBy-type
configuration is currently defined, with the exception of
NameGeneratorList which appears to sort by a single property and has
only been updated to accept the new object syntax: `{ property, direction }`.

Partially resolves: #108

### DIFF
--- a/public/protocols/development.netcanvas/protocol.json
+++ b/public/protocols/development.netcanvas/protocol.json
@@ -947,12 +947,10 @@
             ]
           },
           "sortOptions": {
-            "sortOrder": [
-              {
-                "property": "name",
-                "direction": "asc"
-              }
-            ],
+            "sortOrder": {
+              "property": "name",
+              "direction": "asc"
+            },
             "sortableProperties": [
               {
                 "label": "Pupil Name",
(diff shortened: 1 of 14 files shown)


wwqrd:
@jthrilly Also fixed what seemed like a bug in NodeList et alia? https://github.com/codaco/Network-Canvas/blob/91e95297f908c6fdf97da818ed63cd06aa20aac7/src/components/NodeList.js#L28,L35

jthrilly:
@rebeccamadsen  - any input on NameGeneratorList?

> edit: The nodes didn’t appear to be sorted when reloading the app. I suspect it would also have been the case for directly navigating to an interface from the menu.

Excellent spot.

> The app has been updated to use this sort object everywhere orderBy-type
> configuration is currently defined, with the exception of
> NameGeneratorList which appears to sort by a single property and has
> only been updated to accept the new object syntax: { property, direction }.

The issue with NamGeneratorList is a bit tricky. Basically, we have a few places where we have unintentionally developed overlapping components, and this is one of them. I'm going to create an issue now to address this.

Issue #545, opened 2018-06-05
Better definition and handling of protocol metadata


bryfox:
Currently:
- redux protocols are initialized from mocks.
- there's a loose association between session & protocol (URL-based)
- an updated protocol (new version) overwrites the existing version (I think)

Requirements:
~~- Sessions need to be associated with a protocol version~~

Questions:
~~- Should protocols with multiple versions be shown separately in the 'start session' interface? (Once download, are old versions available for starting new sessions?)~~

**JRM**: I updated the above to avoid confusion about multiple concurrently installed versions of the same protocol. This is no longer something we want to support.

New Requirements:
- Updating a protocol "version" (now defined as a "last modified" property in protocol.js) in a client will prompt the user that continuing will delete all existing interview sessions using this protocol.

Relates to https://github.com/codaco/Server/issues/107.

wwqrd:
I also think that maybe the protocol / session modules could do with a bit of revision, it seems like there are multiple overlaps which could be streamlined so that behaviour is better isolated between the modules. Something like:

# loading a protocol (new session)
1. [open] createSession 
1. loadProtocol / factoryProtocol
1. updateSessionMeta
1. [on change] updateSession

# loading a session
1. [open] setSession
1. loadProtocol / factoryProtocol
1. updateSessionMeta
1. [on change] updateSession

jthrilly:
In addition to agreeing with Steve about restructuring the behavior for protocol loading, I updated the above to be consistent with our current understanding.

bryfox:
@wwqrd can you provide some more detail on how you see the redux state evolving? I think your comments above are describing user workflows; how would those responsibilities be separated between the modules?

Here are some questions I had after looking at it:

- Should we lean more on epics to handle composed flows like download/import/load/set?
  - Conceptually (to me), starting a session would mean loading the protocol, then creating a session. (Currently, if I import an invalid protocol, the app creates a 'zombie' session that can't be resumed.)
  - ...and resuming a session would mean setting the session, then loading the needed protocol.
  - ...but thinking of it like this probably hinders the desired isolation
- When downloading from the Server API, there's additional metadata that makes the above more complicated; one example is remote protocol ID in #556. I assume that's where the `protocols` module will fit in? i.e., `addProtocol()`?
- Does factory protocol handling need to be separate? There seem to be a lot of differences in actions and reducers, but I wonder if that's one of the places we can simplify once we're handling metadata appropriately.

wwqrd:
[...]

Issue #544, opened 2018-06-05
Implement "Finish Interview" UI


bryfox:
This needs more definition to begin. Notes from jthrilly:

> This again has no fixed design at present. Some starting points for an implementation are a menu item or a full blown interface, which is configurable in architect. Right now, I’m excited by the possibilities of the latter.

#543 covers data export; that's out of scope for this issue even if it ends up being part of this interface.

bryfox:
For the next release, we can use a simple version of a Finish Interview interface as a natural point to export (single session) data.

- this can be automatically added (as last stage) without needing configuration
  - though maybe with a simple flag to disable it (or can we use skip logic?)
- interface should contain an 'export' button (assuming there's a server to export to) which exports and closes session.
- interface should contain a 'finish' button which closes session

jthrilly:
> this can be automatically added (as last stage) without needing configuration
> though maybe with a simple flag to disable it (or can we use skip logic?)

I don't think we will need to disable it.

> interface should contain an 'export' button (assuming there's a server to export to) which exports and closes session.

Worth pointing out that "export" could be either (1) uploading to a server, or (2) writing a file in a network file format. Perhaps with a choice between either.

> interface should contain a 'finish' button which closes session

We also had an idea of setting a session flag (`exported`?) to indicate that the session had finished and data had been exported.

bryfox:
>  "export" could be either (1) uploading to a server, or (2) writing a file in a network file format

This could be (one or more of) multiple servers as well (#543)... for this release, I've assumed a single (default) server.

Is file export I should integrate into this interface right now? (Should the Download Data menu item be moved here?)

> setting a session flag

I've added it `lastExportedAt` property to the session.

Here's what it looks like now ([feature/finish-interface](https://github.com/codaco/Network-Canvas/tree/feature/finish-interface)):

![export](https://user-images.githubusercontent.com/39674/41785007-51e09b7c-760f-11e8-9986-dd974dcb9215.png)

jthrilly:
> Is file export I should integrate into this interface right now? (Should the Download Data menu item be moved here?)

Yes, that's actually a really good suggestion.

Looking great!

Issue #543, opened 2018-06-05
Implement data export to Server


bryfox:
Needs definition. This may be an epic that we can open new issues from. Possible requirements & subtasks:

- Data export is always user-initiated (there is nothing like automatic 'streaming' of data)
- UI: Design & implement a view to see finished (but unexported) interviews
- UI: Select a server to export to
- UI: Select which interviews are exported?
- Make API call to server (see https://github.com/codaco/Server/issues/109)
- Clear out stored sessions that have been exported
  - Or make this user-initiated? May a user want to export a session to multiple servers?
- "Session" is likely the unit of export, but export should probably support batches
  - However, see https://github.com/codaco/Server/issues/107. Will there be data other than sessions exported?
- [update] associate sessions with server protocols (see also #529, #545)

bryfox:
Basic exporting is done; follow-on issues referenced above.

Issue #529, opened 2018-06-01
Ability to delete protocols and in progress interviews


jthrilly:
Users should be able to delete protocols from the device, and they should be able to discard/delete in-progress interviews.

- Deleting protocols could be managed by long-pressing the card representing the protocol. Care will need to be taken to handle in-progress sessions using this protocol. We will need to warn that user that the protocol the session refers to cannot be found, and perhaps think about ways to allow them to export data. This could be linked to functionality for exporting data en masse from the start screen.
- Deleting in progress interviews should use the "drag to bin" mechanic from the node list.

bryfox:
Superseded  by #550.

PR #556, opened 2018-06-19
Initial data export

bryfox:
This implements the session export API call, part of #543.

We still need to design UI around export (which is #544, or possibly part of #543 if the 'resume' screen is migrating to a 'management' screen, and needs to coordinate with #529). For now, I've just added this to the existing 'remove' button as an example. If you've paired a server, and the protocol has a remote identifier, then the button will prompt you to export before removing.

To make this work for now, I've added a 'remoteId' to protocol actions, which gets picked up by the sessions reducer. As discussed in #545, sessions & protocols could be simplified where there are overlaps. We can tackle that here, but it may conflict with other related work.

To test this: use with the latest server branch `feature/protocol-versions`.

### DIFF
--- a/src/components/Setup/ProtocolCard.js
+++ b/src/components/Setup/ProtocolCard.js
@@ -20,7 +20,7 @@ const ProtocolCard = ({ protocol, selectProtocol, className, size }) => {
       <Icon className="protocol-card__icon" name="add-a-screen" />
       <div className="protocol-card__labels">
         <h2 className={sized('protocol-card__name')}>{protocol.name}</h2>
-        <p className="protocol-card__version">{protocol.version}</p>
+        <p className="protocol-card__description">{protocol.description}</p>
       </div>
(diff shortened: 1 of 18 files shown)


bryfox:
I removed the in-progress designation here. UI will be handled with #544. Further redux changes discussed at #545.

bryfox:
I pushed an additional commit:
- Fix `type` prop on protocols
- Add remote ID to protocols for now, so that export still works after reload

rebeccamadsen:
If I create a new session from the "Start New Interview" screen, instead of the Server list, it looks like `protocolIdentifier` is undefined so it doesn't offer to export sessions created that way. Possibly this is part of a future issue and you already know.

Otherwise this seems to work for me!

bryfox:
Thanks — I have that fixed in my UI branch; I'll pull it over here.

Issue #550, opened 2018-06-07
Resume Interview 2.0


jthrilly:
Enhancements to the "resume interview" screen:

- [x] Proper styling of cards
- [x] Sorting and filtering by useful criteria (protocol, last modified, network size?). This work should be compatible with #549
- [x] Long press to select multiple interviews, as a means to export.
- [x] Drag to bin in order to delete

bryfox:
Depending how we want to handle export of multiple interviews, we may need an additional issue to support large file uploads over https.

jthrilly:
> Depending how we want to handle export of multiple interviews, we may need an additional issue to support large file uploads over https.

What do you see as "large"? Given that interview data will be limited to JSON (no binary data), and that we are also probably facing quite aggressive RAM and app storage limits (which are as yet uncharted), I think we should expect 100% of interview payloads to be < 1mb, and 99% to be smaller than 500kb, with a mean of around 200kb.

If this is large, what special support do we need to add? Might be easiest if you go ahead and create the issue and link it here so it is ready for someone to pick up.

bryfox:
If uploading via a cordova plugin, I think the bottleneck is serialization into the native side. The hard limit seemed to be on the order of megabytes, but soft limits for UX are possibly lower. I don't expect a single export to be a problem. If we want to upload a single larger batch, we could do so via disk, or use an auth mechanism other than client-side certs so that we can upload using XHR.

PR #559, opened 2018-06-23
Add 'Finish' interface

### LINKED ISSUE #544: Implement "Finish Interview" UI

bryfox:
Resolves #544.

- Adds a built-in 'Finish' interface to the end of all sessions
- Move 'export' functionality this interface
- Move 'download data' from menu to this interface
- Update session cards to display 'exported at' times

Notes:
- 'Download' always appears
- there are variants of Export display, but button only appears if a Server is paired & the session is exportable (has a remote protocol ID)

finish

session-cards

### DIFF
--- a/src/selectors/__tests__/skip-logic.test.js
+++ b/src/selectors/__tests__/skip-logic.test.js
@@ -2,6 +2,10 @@
 
 import { getNextIndex, isStageSkipped } from '../skip-logic';
 
+import { stages as getStages } from '../session';
+
+jest.mock('../session');
+
 const mockState = {
   session: 'a',
   sessions: {
@@ -172,6 +176,10 @@ const mockState = {
 };
 
 describe('skip-logic selector', () => {
+  beforeAll(() => {
+    getStages.mockImplementation(() => mockState.protocol.stages);
+  });
+
   describe('getNextIndex()', () => {
     it('returns the index if valid', () => {
       const index = getNextIndex(1)(mockState);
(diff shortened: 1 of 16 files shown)


src/selectors/__tests__/skip-logic.test.js
```diff
@@ -2,6 +2,10 @@
 
 import { getNextIndex, isStageSkipped } from '../skip-logic';
 
+import { stages as getStages } from '../session';
+
+jest.mock('../session');
```
bryfox: The tests here make assertions about stages length (hardcoded). I think it's appropriate to mock the session selector here, but if we want integration tests between session & skip-logic, then I can revise.

bryfox:
> "reached the end of the interview"

That's much better, thanks. I'll update.

> do we expect users to ever want to export a session again?

Good question. Server won't import a session it already has (https://github.com/codaco/Server/issues/107), and there's not yet a mechanism to select a different export server. On the other hand, I expect the the answer to be yes before too long (notably via the session management screen — #550), so it probably makes sense to not remove the session? I did consider adding a 'remove once finished' checkbox, but left it alone for now.

Flip side: I could leave the Export button there, keeping a "last exported" label, since technically a user can delete the session from Server. But that seems more confusing than just saying export was successful (for now).

rebeccamadsen:
That makes sense. As a "fake" user I deleted a session in server and then wondered how to get it back. But that's definitely not a real use case. :) I'm fine with this for now.

PR #562, opened 2018-06-25
Fix/app menu items

### LINKED ISSUE #486: iOS should not have a "Quit App" item in session menu
The Session Menu includes a "Quit App" item. On iOS, tapping this will throw; it should probably just not be rendered at all.

### LINKED ISSUE #487: Electron app without a window
In Electron (macOS, at least), if I close the window, the app is still running but there's no way to (re-)open a window.

The app should probably either have a menu item to open the window or quit automatically when the window closes.

bryfox:
- Fixes #486: [iOS] Remove "Quit" menu item causing crash
- Resolves #487: [macOS] Add window menu with standard controls + item to re-open main window

### DIFF
--- a/public/electron-starter.js
+++ b/public/electron-starter.js
@@ -67,6 +67,12 @@ function createWindow() {
   });
 }
 
+function reopenMainWindow() {
+  if (mainWindow === null) {
+    createWindow();
+  }
+}
+
 function createMenu() {
   const template = [
     {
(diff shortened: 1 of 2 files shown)


rebeccamadsen:
I can't check the iOS and macOS, but this removes "Quit" from the Chrome version. Is that intentional? I know users can just close the browser or tab, just curious.

Otherwise, Android and electron on windows work fine.

bryfox:
No, that wasn't intentional. I wasn't thinking about browser dev.

Looks like the functionality doesn't work there (`Scripts may close only the windows that were opened by it.`); do we want it anyway?

bryfox:
I added the item back in for web; we probably want to at least see the menu item during dev.

rebeccamadsen:
Looks fine to me! Although, again, I did not test on iOS. :)

bryfox:
I double-checked iOS. If I'm mistaken, the author of #486 can hold me responsible.

PR #563, opened 2018-06-25
Remove looping on swiper

### LINKED ISSUE #557: Cannot start new interview after looped swipe navigation
Steps:
1. Load start screen
2. Click back arrow
3. Click protocol

Expected: loads protocol. Observed: nothing happens.

(Same behavior after navigating from last to first.)

With `loop` mode on the swiper, there are duplicated protocol nodes in the DOM (but not React's virtual DOM). We can either disable looping, or manage events outside of React for the duplicated nodes.

bryfox:
Looped nodes are cloned in DOM (outside of React). Fixes #557.

This removes looping from the protocol list. I experimented unsuccessfully with other settings in [swiper](http://idangero.us/swiper/api/). I'd like to avoid additional event handling outside of React, but if looping is required, then I can add that as described in the issue.

### DIFF
--- a/src/containers/Setup/ProtocolList.js
+++ b/src/containers/Setup/ProtocolList.js
@@ -48,7 +48,7 @@ class ProtocolList extends Component {
         nextEl: '.swiper-button-next.swiper-button-white',
         prevEl: '.swiper-button-prev.swiper-button-white',
       },
-      loop: true,
+      loop: false,
       shouldSwiperUpdate: true,
       rebuildOnUpdate: true,
     };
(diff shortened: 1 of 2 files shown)


rebeccamadsen:
This was fine for me in electron and android, but in chrome on windows, the swiper initialize like this: 

untitled

The swiper centers things again if I add this to protocol-list.scss:

```
.protocol-list {

  &.swiper-container {
    width: 100%;
  }

```

bryfox:
You're right; I also see it on electron when wide enough  — thanks for the catch & the fix!

rebeccamadsen:
Looks good to me!

PR #564, opened 2018-06-25
Use larger viewBox for Safari

### LINKED ISSUE #398: Disappearing edges on iPad
Edges between two connected nodes sometimes disappear when dragging at least one of those nodes to a new angle. The edges generally return when the angle is further adjusted. The two pictures below have the same edges with nodes (Dee, Eu, Ryan, and Quinn) moved around.

968e2b08-8cef-4abf-9987-3b51511d7ac9
89c5e3f0-8c5b-4d10-b037-0cff89995e28

bryfox:
Fixes #398.

I'm not sure exactly what the issue was here — seems like a loss of precision until a full repaint is done — but using a larger viewBox seems to make iOS happy. I've tested iOS 11.4 sim [edit: and device (iPad Air)].

### DIFF
--- a/src/components/Edge.js
+++ b/src/components/Edge.js
@@ -7,15 +7,15 @@ const getEdgeColor = makeGetEdgeColor();
 
 export class Edge extends PureComponent {
   render() {
-    const { from, to, color } = this.props;
+    const { from, to, color, viewBoxScale } = this.props;
     if (!from || !to) { return null; }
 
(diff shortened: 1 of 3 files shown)


jthrilly:
Fantastic stuff Bryan. I was dreading this issue, but seems like a logical fix. I can't test right now, but if this sim shows this as fixed, please merge.

Issue #551, opened 2018-06-07
Layout churn/animation choppiness when opening full screen node form


jthrilly:
When opening the full screen new node form on the pixel C, the experience is poor. 

This seems to be because the initial open animation conflicts with the keyboard opening. The animation has not completely finished when the keyboard resizes the viewport, causing further render thrashing.

Potential fixes (one or more of these):

- Delay focusing the first input until after the open modal animation has finished. This should prevent the keyboard opening too soon.
- Disable the open modal animation entirely for full screen modal. Sliding from the top doesn't make a great deal of sense here anyway, since there is so little space.
- Be much more specific about which elements animate, and when. For example, use an `opening` class to enable transitions for properties such as height or position, and remove it once the animation is finished.

PR #565, opened 2018-06-26
Fix animation choppiness on fullscreen modal form

bryfox:
Attempt to fix animation choppiness for #551 (using approach 2).

This will need testing on Pixel C hardware.

### DIFF
--- a/src/styles/components/_modal.scss
+++ b/src/styles/components/_modal.scss
@@ -33,7 +33,6 @@ $component-name: modal;
 
     &-content {
       height: inherit;
-      max-height: 80vh;
     }
 
     &-title {
@@ -42,7 +41,6 @@ $component-name: modal;
       flex-direction: column;
       align-items: center;
       justify-content: center;
-      max-height: 20vh;
       padding: 65px 0 0;
     }
   }
@@ -68,17 +66,30 @@ $component-name: modal;
   }
 
   &--fullscreen {
-    .#{$component-name} {
+    display: block;
 
+    .#{$component-name} {
       &__background {
         background-color: palette('background');
       }
 
       &__window {
-        display: flex;
-        flex: 1 1 auto;
         height: 100vh;
       }
+
+      &__layout {
+        display: block;
+
+        &-title {
+          display: block;
+          padding-top: 5rem;
+          text-align: center;
+        }
+      }
+    }
+
+    .input__container--hidden {
+      position: absolute;
     }
 
     .form {
--- a/src/containers/SettingsMenu.js
+++ b/src/containers/SettingsMenu.js
@@ -229,7 +229,7 @@ class SettingsMenu extends Component {
         title="Settings"
         toggleMenu={toggleMenu}
       >
-        <div style={{ position: 'fixed', top: 0, right: 0, display: 'inline', padding: '10px', zIndex: 1000 }}>{ version }</div>
+        <div style={{ position: 'fixed', top: 0, right: 0, display: 'inline', padding: '10px', whiteSpace: 'nowrap', zIndex: 1000 }}>{ version }</div>
         <Dialog
           name="CONFIRM_DELETE_DATA"
           title="Delete ALL data?"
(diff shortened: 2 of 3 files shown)


rebeccamadsen:
I don't have the hardware, but the emulator looks okay to me.

bryfox:
I'll assign to @jthrilly for additional review when he's available; hopefully he has access to a Pixel C since he filed the issue.

jthrilly:
Afraid I don't have the Pixel C, but I have the Chromebook plus running the android app. I'll give it a whirl, and if it is okay I suggest we merge this since it is relatively low impact and easily reverted.

bryfox:
Using "Add Person" from the dev protocol as my example, I've gone through the form & modal contents looking for elements with transitions set. The only positioning-related transition I see set is on `.text__label`, and that isn't transitioning on the Android emulator (it does if I change it to transition all, but I'm guessing that's not the cause of what you're seeing).

We could manually control (and delay) the focus event using refs. Since the modal itself isn't transitioning, I suspect that would just delay the transition problems you're seeing, but it might improve things if the view is having trouble rendering the initial layout in time. A rough test for that: does the animation look OK when you tap to blur & focus the field after the modal has loaded?

I'm not sure if there's a way to prevent the animation triggered by the soft keyboard appearing. We can be pretty sure that the field will be in view when a modal first appears, and we shouldn't need to support scrolling for the wizard version of the form.

One other thing we might consider is to stop using flex layout for the various components of the wizard. It seems like some of the jank is coming from the relative re-positioning of elements while the viewport height changes (even though we're not adding transitions). The alternative would probably be to have everything aligned from the top of the view.

jthrilly:
> One other thing we might consider is to stop using flex layout for the various components of the wizard. It seems like some of the jank is coming from the relative re-positioning of elements while the viewport height changes (even though we're not adding transitions). The alternative would probably be to have everything aligned from the top of the view.

I think this is the crux of the issue, actually. The question is: why does the recalculation seem to be "transitioning" as the keyboard opens? I wonder if it is something in the web view that someone thought would be nice, but is actually a pain.

I wonder if we shouldn't try absolutely positioning the arrows and content to a place that works for when the keyboard is open, and have a breakpoint for device height so we can put the alignment back to the middle for if/when the keyboard is closed. Thinking being that proportional units might be the cause of the recalculations as the keyboard opens. 

Thoughts?

bryfox:
> why does the recalculation seem to be "transitioning" as the keyboard opens?

On a 'normal' page (with statically-positioned elements), the viewport transitions up along with the keyboard (iOS & Android). Maybe the renderer is recalculating flex layout at various points during that transition?

> breakpoint for device height 

Interesting... I'll give this a try.

bryfox:
@jthrilly I've pushed a completely different fix (css-only), which removes the flex layout from the fullscreen modal form. The previous effort still exists in [fix/static-modal](https://github.com/codaco/Network-Canvas/compare/fix/static-modal).

Notes:
- I didn't use breakpoints to target the shorter height only. The flex layout only has two visible children with very little variation in layout; the new version matches the previous flex layout for typical tablet dimensions; other devices would only differ very slightly, and this implementation is cleaner.
- This version includes the standard transition modal (unlike the previous code that removed the transition). I don't think we need a special case for that any longer, but I could add it back in if needed.
- next/prev buttons are still positioned halfway down the screen, and so keep the automatic sliding behavior.

src/styles/components/_modal.scss
```diff
@@ -42,7 +41,6 @@ $component-name: modal;
       flex-direction: column;
       align-items: center;
       justify-content: center;
-      max-height: 20vh;
```
bryfox: I removed these max-heights from both versions; overflow is visible, so if the title is that long, it obscures form text anyway. We could limit to just the fullscreen version, however.

bryfox:
I tacked on an additional fix for the version string, which sometimes wraps and renders the cancel button un-clickable (Android).

wrap

jthrilly:
Thanks Bryan. Reviewing this today. Will reply shortly.

jthrilly:
This looks great, Bryan! Fixes the issue for me on the devices. Happy for this to be merged whenever you like.

PR #566, opened 2018-06-26
Update .eslintrc to .json

### LINKED ISSUE #396: Update .eslintrc to one of the new formats
According to [this](https://eslint.org/docs/user-guide/configuring#using-configuration-files) eslint is moving away from `.eslintrc` as a config file towards other formats.

At some point, we will want to move our configuration to one of these.

The reason this came up is that I wanted to implement a custom rule that extends `no-redeclare` to also catch functions that are redefined in the format `object.property = () => {}`. There's currently no easy way to do this without creating a custom package when using `.eslintrc`

bryfox:
Resolves #396.

@jthrilly is this all you needed for #396? Or did you actually want to be able to write custom functions in the config file (in which case .js is the only option)?

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -8,7 +8,7 @@
     },
     "parser": "babel-eslint",
     "parserOptions": {
-        ecmaVersion: 7,
+        "ecmaVersion": 7,
         "ecmaFeatures": {
             "jsx": true,
             "spread": true,


jthrilly:
Not right now, no. But this change will make that possible when I eventually do. I worked around it in the meantime.

bryfox:
Would you like this merged, then? Or should we just hold until it's needed?

Issue #570, opened 2018-06-27
Migrate 'factory' protocols to standard protocols


bryfox:
If factory protocols were treated as standard protocols (albeit ones that are automatically imported to the user data directory), then we no longer need special handling for factory protocol types (in redux state, actions, file handling, and app routing).

To bootstrap the factory protocols, files can be copied (from app dir) at startup and the corresponding redux state updated.

- Because these are actively developed, they should be restored easily/automatically on reload (at least during development).
- [TBD] Users shouldn't be prevented from overwriting those protocols (e.g., we don't need to make 'Teaching Protocol' a reserved name)

jthrilly:
@wwqrd - any thoughts on this?

jthrilly:
I've taken care of this, by removing the concept of factory protocols.

Issue #569, opened 2018-06-27
Prevent or clean up invalid sessions


bryfox:
Currently, 'zombie' sessions are created when importing an invalid protocol; they can never be resumed since there's no associated session. Ideally, a user would not have to manually clean up.

Update: With updates to redux actions, need to also ensure this is fixed for failed downloads (as mentioned in #571).

Related to #568: if session creation is decoupled from protocol loading, then these should be cleaned up. If sessions are created after their protocols are validated, this will be prevented.

rebeccamadsen:
Another side effect of the zombie sessions is that you can add mock nodes to a zombie session just after creation; the menu item shows up. I don't think it's worth making this a separate issue, though, because not being able to create zombie sessions would eliminate the issue of adding mock nodes to a zombie session.

jthrilly:
This was resolved by changes to protocol import.

PR #571, opened 2018-06-27
Update redux for protocols & sessions

### LINKED ISSUE #545: Better definition and handling of protocol metadata

### LINKED ISSUE #558: Duplicate protocols appear in setup list
Downloaded protocols are saved with unique (uuid) filenames, so we can't rely on path to determine whether a protocol already exists in state.

#556 contains an interim fix for protocols imported from Server (based on remote ID), but this is still an issue with protocols downloaded by URL.

This is probably relevant to the discussion of redux state shape in #545.

bryfox:
Resolves #545. Resolves #558.

This updates redux actions (with some minor changes to state shape) to better support protocol exporting and simplify some of the import/load/set flows with sessions. This builds on the FinishSession interface branch (a merge should wait on #559), but the two can be reviewed in parallel.

This also adds support for loading devtools extensions in electron (as redux-devtools was very useful for this effort), but installation left as a manual exercise (see Readme).

### Notes on changes
- Removes some of the differences between `loadFactoryProtocol` and `loadProtocol`
  + In particular, both now make use of `addSession()`
  + ...and session IDs are no longer generated by protocol actions
  + This reduces some of the coupling, but interfaces must call addSession in both cases now. (Unwanted side-effects from programming error should be solved by #570.)
- Removes the `remoteId` that was passed through actions (to provide initial support for export). Remote IDs are now calculated locally as they are not server-specific.

### Manual testing

We don't have much integration testing yet, and there's still integration between sessions and protocols. I've gone through combinations of {protocolType} + {loadAction|exportAction}:

- Factory protocols vs Downloaded (Server- and URL-sourced) protocols
- Loading
  - existing sessions
    - via app reload
    - via 'resume interview'
  - new Sessions
    - via protocol cards (startup screen)
    - via download flows
- Exporting (or not) as expected

### DIFF
--- a/public/electron-starter.js
+++ b/public/electron-starter.js
@@ -144,12 +144,21 @@ function createMenu() {
   Menu.setApplicationMenu(appMenu);
 }
 
+function loadDevToolsExtensions() {
+  const extensions = process.env.NC_DEVTOOLS_EXENSION_PATH;
+  if (process.env.NODE_ENV !== 'development' || !extensions) {
+    return;
+  }
+  extensions.split(';').forEach(filepath => BrowserWindow.addDevToolsExtension(filepath));
+}
+
 // This method will be called when Electron has finished
 // initialization and is ready to create browser windows.
 // Some APIs can only be used after this event occurs.
 app.on('ready', () => {
   createMenu();
   createWindow();
+  loadDevToolsExtensions();
 });
 
 // Quit when all windows are closed.
--- a/README.md
+++ b/README.md
@@ -132,6 +132,17 @@ Troubleshooting:
 - If a dev cordova build is interrupted, you may find that `config.xml` has changes, and that there's a `config.xml.original` file (ignored by git). You may restore the contents of `config.xml` from git or the original, and delete the '.original' file.
 - The webpack dev server writes a `.devserver` file on startup, which is removed when it exits. The file is used by the cordova dev build; it should contain the LAN URL of the running dev server.
 
+### Dev tools
+
+Electron supports [extensions to Chrome devtools](https://electronjs.org/docs/tutorial/devtools-extension) such as Redux DevTools.
+
+In the development environment, these will be loaded if you provide one or more paths to your extensions (semicolon-separated) in the `NC_DEVTOOLS_EXENSION_PATH` environment variable. The electron docs describe how to find the filepath for an extension once installed.
+
+Example: enabling Redux Devtools on macOS:
+```bash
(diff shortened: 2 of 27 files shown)


rebeccamadsen:
If I try to "import from url" in chrome (which shouldn't work, but I can click on it, so), a "fake" session gets created.

Some protocol names disappear on the "resume interview" screen. I think when I add a protocol over again from the server list, or from the "import from url" button.

Otherwise the iterations I tried worked well. I like the list for manual testing you've started.

public/electron-starter.js
```diff
@@ -125,12 +125,21 @@ function createMenu() {
   Menu.setApplicationMenu(appMenu);
 }
 
+function loadDevToolsExtensions() {
```
wwqrd: Cool! I'll add this to Architect.

bryfox:
> If I try to "import from url" in chrome (which shouldn't work, but I can click on it, so), a "fake" session gets created.

Good catch! This is related to #569, where fake sessions are created if 'load' fails. But in this branch, they'll also be created if 'download' fails. I think the same problem would occur if, for example, you entered an import URL that 404'd.

This was a side-effect of my using `addSession()` for download protocols (not just factory). I'll need to do a little more thinking on a fix.

bryfox:
I've rebased this onto master, which now contains the FinishSession interface.

I addressed the two issues @rebeccamadsen found as follows:

1. 'Fake' Sessions with no corresponding protocol are removed from the SessionList. For now, I'm letting them be created in the store; they're a rare case. Removing them entirely will take more integration between protocols & sessions, which should be cleaner once we get rid of the 'factory' special case. We can use #569 to track.

2. Sessions whose protocolPath ref is outdated (indicating the protocol was updated) display '[version out of date]' instead of the name. Displaying the name for these would again require more coupling between the modules, and (in the near future) protocol updates will invalidate/remove old sessions anyway.

PR #572, opened 2018-06-27
Add basic error boundary for stages

bryfox:
In the face of errors caused by stage loading or interaction (which is commonly happening as we update the protocol format), this allows user to continue or restart, rather than having to quit app or reset all data.

I think we've discussed using boundaries in places, but I don't see an issue filed. Apologies if this isn't what others had in mind — but it's making development with all of my [now outdated] protocol files easier.

### DIFF
--- a/src/components/StageErrorBoundary.js
+++ b/src/components/StageErrorBoundary.js
@@ -1,0 +1,37 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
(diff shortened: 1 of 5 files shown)


rebeccamadsen:
This seems pretty great to me! I tested with the demo protocol in electron (because: sort order changed) and by removing all the prompts from a stage in the education protocol in chrome. It's nice not to have to quit the app.

wwqrd:
:+1:

PR #576, opened 2018-06-29
[WIP] card styles

jthrilly:
Changing the card style to better reflect other elements of the overall style guide, and be more consistent throughout the app.


wwqrd:
I like the transitions!

I don't think the text should move horizontally between states, keeping text lined up makes it easier to skim/read.

When inactive the cards seem a little "unanchored".

rebeccamadsen:
I ditto the "unanchored" comment, especially when you look at the cards in the Auto Complete list.

jthrilly:
Agree on that point, definitely. I am probably going to absolutely position the checkbox and take it out of the flow of text entirely. Thanks for the input.

PR #579, opened 2018-07-02
Fix/protocol loading

bryfox:
Fixes loading of factory & demo protocols generated from a prior NC version.

### DIFF
--- a/src/containers/Setup/ProtocolList.js
+++ b/src/containers/Setup/ProtocolList.js
@@ -13,12 +13,14 @@ import { actionCreators as sessionsActions } from '../../ducks/modules/sessions'
   */
 class ProtocolList extends Component {
   onClickNewProtocol = (protocol) => {
-    if (!(protocol.isFactoryProtocol || protocol.path)) {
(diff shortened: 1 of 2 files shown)


jthrilly:
I'm a bit conflicted on this one. Am I right in thinking that as long as the factory protocols are updated to the new format, and the app data is reset, there is no need for this compatibility layer? If so, I think we should leave it out.

Referring back to something you said on slack, Bryan, I think changing the protocol format between releases is a given for the alpha phase at least, so we don't need to worry about breaking things. There are no current users outside a very select group.

Is there something more here that I am not thinking of?

bryfox:
Correct; if app data is reset (testing a fresh install), then there's no need for this compatibility code. 

Resetting of the app between alpha releases was my assumption with the initial PR; that seemed to take people by surprise, so I added the checks here. I can close without merging if you prefer.

jthrilly:
> Resetting of the app between alpha releases was my assumption with the initial PR; that seemed to take people by surprise, so I added the checks here. I can close without merging if you prefer.

Ideally we wouldn't need to, and at some point it would be good to actual deal with this deliberately (i.e transition an old data structure to a new one). But for now, let's just keep breaking things. 

Closing, but leaving around just in case.

PR #580, opened 2018-07-03
Send token with API requests

### LINKED ISSUE #541: Send token (obtained during pairing) with API requests
Client-side equivalent of https://github.com/codaco/Server/issues/106

bryfox:
Resolves #541. This stores device credentials (ID & Secret, associated with a server) once pairing is complete.

For subsequent API requests, the client sends a token (ID). This uses the username part of HTTP Basic Auth (at least for now); this is only for identification. We'll eventually be using another mechanism to authenticate, but this lays some of the groundwork.

Because of CORS headers, this requires server-side changes (or API calls will fail); see [feature/device-tokens](https://github.com/codaco/Server/compare/feature/device-tokens).


rebeccamadsen:
As far as I can tell this looks fine!

PR #590, opened 2018-07-10
Update protocols to use new sort prop

bryfox:
Make examples match #589.

Will need to update [wiki](https://github.com/codaco/Network-Canvas/wiki/Sociogram) upon release.


jthrilly:
Thank you Bryan. I shouldn't have missed this in the first PR - must learn to stop assuming things!

PR #594, opened 2018-07-10
add missing values

rebeccamadsen:
This is a small change to change "values" to "options" for the education protocol. Without "options", the Ordinal Bin for the education protocol was not loading. Development protocol was working, so I assumed "options" was the correct word here.

This also adds a missing variable type for "special_category" to the development protocol, so graphml will download correctly.

### DIFF
--- a/public/protocols/development.netcanvas/protocol.json
+++ b/public/protocols/development.netcanvas/protocol.json
@@ -154,6 +154,11 @@
             "label": "Prompt id",
             "description": "Prompt id",
             "type": "text"
+          },
+          "special_category": {
+            "label": "Special Category",
+            "description": "Special Category",
+            "type": "categorical"
           }
         }
       },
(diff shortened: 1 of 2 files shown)


bryfox:
This looks good to me.

public/protocols/development.netcanvas/protocol.json
```diff
@@ -154,6 +154,11 @@
             "label": "Prompt id",
             "description": "Prompt id",
             "type": "text"
+          },
+          "special_category": {
+            "label": "Special Category",
+            "description": "Special Category",
+            "type": "categorical"
```
jthrilly: Shouldn't the categorical variable define its valid categories here, same as an ordinal variable?

jthrilly: Bryans PR adds this so just merging them in sequence.

PR #595, opened 2018-07-10
Fix standard protocol card styling

bryfox:
This fixes the positioning of the protocol card icon when used on the server protocol list.

before:
card

after:
after

Styling on the startup screen (the 'large' variant) should remain the same within a couple of pixels.


rebeccamadsen:
Looks good to me!

PR #598, opened 2018-07-10
Remove unsupported hidden inputs

bryfox:
This removes the hidden inputs from the development protocol, which are no longer [documented](https://github.com/codaco/Network-Canvas/wiki/Variable-Types) as supported, and render as text inputs instead.

development

### DIFF
--- a/public/protocols/development.netcanvas/protocol.json
+++ b/public/protocols/development.netcanvas/protocol.json
@@ -963,11 +963,6 @@
         {
           "variable": "age",
           "component": "TextInput"
-        },
-        {
-          "variable": "timeCreated",
-          "component": "hidden",
-          "value": "return Date.now().toString();"
         }
       ]
     },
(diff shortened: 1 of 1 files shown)


bryfox:
Update: I restored the `timeCreated` definition as it's needed to export mock data. (We could also remove that field from the mock factory.)

jthrilly:
So I think this is a slight misunderstanding. The _variable type_ hidden is indeed not a thing. However, any variable can be mapped to a hidden _input component_, which will cause it not to display. This seems to be what has been removed here, which is an error.

I think what has happened is that in the process of cleaning up the field mapping of variables to input components in the form generator, we have removed the mapping to the hidden component. So the fix for this should be to reinstate that.

jthrilly:
Documentation for input components: https://github.com/codaco/Network-Canvas/wiki/Input-Types

bryfox:
That makes sense. See https://github.com/codaco/Network-Canvas/issues/606.

PR #600, opened 2018-07-10
Update import URL for working demo

bryfox:
This points the import URL at a working demo protocol.


jthrilly:
I think for this, we should remove the repo this points to, and replace the initial text with `https://`

Issue #596, opened 2018-07-10
Possible unexpected attributes when rendering nodes & inputs


bryfox:
It looks like we now render arbitrary attrs on Node divs. This causes PropType validation errors, even with the factory protocols, such as:
- protocol attrs (e.g., `stageId`) is not a valid DOM attr
- non-boolean attrs (`school_important`) are passed with boolean values
- `dispatch` functions are passed

PropType warnings are only in development, so this may not matter in the short term. I don't see any other unwanted behavior.

This could theoretically cause unexpected behavior with certain protocols (say, with a "hidden" attribute in external data), but again this probably doesn't matter short-term.  In general, it quickly gets hard to reason about the coordination of protocol.json and UI components — both of which are changing regularly.

This affects:
- [ ] Node component
- [ ] Field component (hidden input); see #608

bryfox:
This is really the same as #446, it's just gotten worse since then. I'll close this as a duplicate.

PR #608, opened 2018-07-11
do not override type for hidden input

### LINKED ISSUE #606: Make hidden inputs type=hidden
> in the process of cleaning up the field mapping of variables to input components in the form generator, we have removed the mapping to the hidden component. So the fix for this should be to reinstate that.

hidden-inputs

https://github.com/codaco/Network-Canvas/wiki/Input-Types

rebeccamadsen:
Fixes #606.

To verify, look at the new node form in the development protocol, in both mobile and regular view. The hidden "timeCreated" field should not be visible, but present in the dom.

### DIFF
--- a/src/containers/Field.js
+++ b/src/containers/Field.js
@@ -26,7 +26,7 @@ const fieldTypes = {
   RadioGroup,
   Text,
   Toggle,
-  hidden: props => <input type="hidden" {...props} />,
+  hidden: props => <input {...props} type="hidden" />,
 };
 
 export const getInputComponent = componentType =>


bryfox:
This fixes the reported issue.

However, I think the issue is indicative of a deeper problem: that we're passing arbitrary props into the <input> element. As another side effect, I see react warnings ("React does not recognize the `fieldLabel` prop on a DOM element.").

That's the same basic problem in #596, so I can update that issue to cover this case as well; I think we can merge this one.

PR #611, opened 2018-07-17
Fix/node positioning false

### LINKED ISSUE #591: In NameGenerator nodes can be positioned from NodeBin even when `allowPositioning` is set to false
```
layout: {
  allowPositioning: false,
}
```

Correctly prevents already placed nodes from being positioned, but still allows them to be place from the NodeBin.

If this is the intended behaviour it is questionable because mis-placed nodes can't be repositioned.

Suggest hiding the NodeBin for this setting.

rebeccamadsen:
Fixes #591. This fixes the sociogram's NodeBucket to only display if `allowPositioning` is `true`. To test, set `allowPositioning` to false in a sociogram prompt's `layout`. The NodeBucket should not display.

### DIFF
--- a/src/containers/__tests__/__snapshots__/NodeBucket.test.js.snap
+++ b/src/containers/__tests__/__snapshots__/NodeBucket.test.js.snap
@@ -1,11 +1,59 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`<NodeBucket /> renders ok 1`] = `
+exports[`<NodeBucket /> does not display bucket when positioning is disabled 1`] = `
 ShallowWrapper {
   "length": 1,
   Symbol(enzyme.__root__): [Circular],
   Symbol(enzyme.__unrendered__): <NodeBucket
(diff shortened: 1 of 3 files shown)


src/containers/__tests__/__snapshots__/NodeBucket.test.js.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`<NodeBucket /> renders ok 1`] = `
+exports[`<NodeBucket /> does not display bucket when positioning is disabled 1`] = `
```
bryfox: Just personal preference, but looking at this snapshot (compared the other), it doesn't immediately jump out that there's no inner content rendered.

An assertion like `expect(component.children()).toHaveLength(0)` might be more clear.

rebeccamadsen: That works for me. I've updated here.

PR #612, opened 2018-07-18
Allow electron to start even if dev extensions fail

bryfox:
In development, if you've set a path to a load devtools extension that
includes a version number, and the extension is upgraded, the call to
addDevToolsExtension throws an error we should catch.


rebeccamadsen:
This looks fine to me, but I'm not positive how to verify it.

PR #614, opened 2018-07-24
Feature/collapse instructions

### LINKED ISSUE #430: Feature request: collapsing instructions
The node occlusion issue on the sociogram screen presents a challenge. do we prevent people from dropping on the left hand side? The box of instructions seems to be collapsible if you click the green bar (as the name generator collapses, e.g., 'Your classmates' in the teaching protocol). But this bar neither closes nor is movable. This means that one can drop a node in behind that square and it cannot be retrieved. There's no 'undo'. The reset button resets everything and not just the layout on that page. 

Putting the text at the top of the sociogram really warps the layout and pushes everything down, which is not idea on standard sized sccreens  for laptop and mobile. I propose that the instructions are altered so that they say 'instructions' at the top, and have the green bar at the bottom, then they can be collapsed just like the other panels. However, in this case, they should collapse downwards so that it is a tab at the bottom and not upwards so that it is a rectangle roughly 25% from the bottom of the screen.

rebeccamadsen:
Resolves #430. Adds a minimize button to the instructions in sociogram:
minimizable

and collapses to a small button when clicked:
minimized

### DIFF
--- a/src/containers/Interfaces/Sociogram.js
+++ b/src/containers/Interfaces/Sociogram.js
@@ -27,6 +27,7 @@ const Sociogram = ({
         prompts={stage.prompts}
         prompt={prompt}
         floating
+        minimizable
       />
     </div>
     <div className="sociogram-interface__sociogram">
--- a/src/containers/PromptSwiper.js
+++ b/src/containers/PromptSwiper.js
@@ -81,13 +81,22 @@ class PromptSwiper extends Component {
         'prompts--minimized': this.state.minimized },
     );
 
+    const minimizeButton = (
+      <span className="prompts__minimizer" onClick={this.handleMinimize}>
+        {this.state.minimized ? '?' : '—'}
+      </span>
+    );
+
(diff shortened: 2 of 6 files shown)


bryfox:
This looks good to me! The CI error seems like it might just be transient; maybe force-push to retry it?

wwqrd:
We could also connect this up to `MonitorDragSource` and auto hide this panel when a node is being dragged?

rebeccamadsen:
> We could also connect this up to MonitorDragSource and auto hide this panel when a node is being dragged?

Interesting. Would that be in addition to a button option? If they place a node in that "blocked" space, they need a way to access the node other times beyond just dragging (e.g. if they want to move it again or link an edge or highlight).

wwqrd:
I wonder if we could hide the origin "bucket" too - though that would have the same issue of moving once placed.

It's almost like we need a "show desktop" button or a "placement mode".

Issue #587, opened 2018-07-09
download graphml errors: missing variables from registry for default node form


rebeccamadsen:
Creating a protocol with the default form in Architect, and adding a node in NC using that protocol, means that downloading graphml will fail with this error message:

`The variable registry seems to be missing "type" of: "uid" in node.person, "type" in node.person, "stageId" in node.person, "promptId" in node.person.`

I think these are all variables we create on nodes in NC. To address this, either (1) Architect defines these in the variable registry (which doesn't seem right), or (2) we have a separate method to define system generated variables (not by checking the registry), or (3) we don't export system generated variables.

wwqrd:
I think that these shouldn't really be in the protocol - that's for user defined variables. Since they are required by the app we can assume that they will always be there as fixed types.

bryfox:
From #597, some of the fields to consider (which aren't documented as public/exportable):
- stageId
- promptId
- createdIndex
- timeCreated

Issue #446, opened 2018-04-13
node field/prop collisions


rebeccamadsen:
Since we use the spread operator to include all the fields in a node or `EnhancedNode` _after_ our props (in most or all cases?), it is possible we could have unintended collisions and clobbering of network canvas functionality. Some overriding of props is likely intentional (e.g. `label`, `color`, `selected`). Some is decidely not (e.g. `meta`, `onSelected`, `allowDrag`). Which is which is up for discussion.

This is related to our [uid discussion](https://github.com/codaco/Network-Canvas/issues/415). See the conversation [here](https://github.com/codaco/Network-Canvas/pull/443#issuecomment-381155177), with one idea being to prepend special props with an underscore.

wwqrd:
This configurations allows you to override default properties. In cases where we don't want to do that we can change the position of the spread operator:

```jsx
<EnhancedNode
  {...props}
  label="Non overridable field"
/>
```

Or in cases where we want to exclude specific properties from the component:

```jsx
const foo = ({ dontIncludeThis, orThis, ...rest }) => (
  <EnhancedNode
    {...rest}
  />
);
```

jthrilly:
I think reordering will work. @rebeccamadsen ?

rebeccamadsen:
I think reordering is part of the solution. It would ensure the user can't override our system created props and cause a crash or otherwise adversely affect functionality of the app; it protects us. But without the documentation to warn a user which fields we are using, wouldn't we still run the risk of clobbering their data? It's a two way clobbering. Maybe they have their own use for a 'meta' field just as much as we do?

jthrilly:
I think I am still in favor of our "special" properties having some kind of pattern to them. Either nc_prop or just _prop.

@wwqrd - this seems like the sort of suggestion you would hate. Can you think of a better way?

bryfox:
I'm trying to understand the reason for spreading external data props into view components, rather than keeping things contained in something like a `sourceData` prop. With the node example:

- What view properties are we intending users to customize directly via externalData? Labels, for example, have a different mechanism (`displayLabel`).
- Should 'special' props (including uid) be exported with data? Currently, the export merges view information (like stageId) into the Node data (which I assume is intentional); put another way, is the merging of app & user data required?
- Can a user export data from the app, and source that as external data for another interview? (If so, does that make special naming (`nc_*`) an incomplete solution for this problem?)

[...]

Issue #397, opened 2018-02-22
Sticky Baz


Defelt:
when i reached the end of the list of classmates and dragged "Baz" over, rather than disappearing from the column on the left, the name stayed and I ended up dragging it in multiple times - because I had to scroll down in the right half of the screen to see names being added at the bottom, I'd already added it several times before I realized there was a problem. After that I just did it a few more times to see if it would resolve itself but it didn't seem to

![image 3](https://user-images.githubusercontent.com/36740723/36553260-7b19da6e-17c1-11e8-8fa5-5b44a7501fa2.png)

![image 3](https://user-images.githubusercontent.com/36740723/36553245-71dd186c-17c1-11e8-9530-2a82fd2de1b0.png)

rebeccamadsen:
This reminds me of an issue I saw [here](https://github.com/codaco/Network-Canvas/pull/293#issuecomment-360851048). Basically, it is a bug related to Baz not having a uid in the dataset. The "quick fix" is to make sure Baz has a uid in the dataset, but the bug is because the uid is not generated until you drag it to the main list. And since the generated uid doesn't match the Baz node, Baz stays in the side panel as well. Dragging it again just creates a new uid over and over.

Interestingly, I also notice the "Annie" is in the dataset but not appearing in the side panel. Maybe that is related?

wwqrd:
Perhaps we should make the following changes @rebeccamadsen @bryfox?:

1. External data *must* be provided with a `uid` (instead of an `id`).
1. Randomly generate `uids` for new nodes using something like this [uuid](https://www.npmjs.com/package/uuid), that way we can assume they won't clash with the user provided ones.
1. Remove the `id` property from nodes

rebeccamadsen:
I think this greatly simplifies other issues we've seen in trying to solve uniqueness!

My only concern is from the user facing side. Is it acceptable from the user perspective to require `uid`s to be unique? I think within one dataset, sure. So how often will they use multiple data sets in the same interview? Does/will the Architect massage user data, or is that all on the user?

I suppose the interviewers have to be prepared to work with data at the end, when they export it, so it could be fair enough to expect them to present clean data at the beginning as well. I just want to make sure we've thought through the side effects of making `uid`s a user issue.

wwqrd:
[...]

PR #615, opened 2018-07-25
Ensure uids exist for all nodes in app

### LINKED ISSUE #415: Components that consume external data as a network must generate UUIDs for nodes

### LINKED ISSUE #605: Forms: user input should overwrite additional attrs
If a form contains an input and an additional attribute that share a name, the user-entered value should be used in the resulting form output. Additional attributes could be used to provide defaults in this case.

bryfox:
Resolves #415.

This standardizes the behavior of primary keys for nodes in the app:

- Each node imported from external data are given a unique identifier, unless it already has one
  + UIDs are stable across imports as long as input data is stable (UID == digest of contents)
  + This is done up front, at least for now (may need to change with #413)
- If external data defines PK values, they are assumed to be unique.
- A protocol author may define a different key for each data source, using `"nodes_primary_key"`. See the example of `studentId` in the dev protocol.
- The convention (currently; see definition in modules/network) is to use the `uid` prop as an identifier. Nodes always have this property defined internally, regardless of the external data's primary key.
- Search interface can now rely on UIDs for result display

### Things to test
- Add nodes from different sources, e.g.
  + `previousInterview` has existing `uid`s
  + `schoolPupils` uses the same IDs, but uses a custom PK ("studentId") 
  + `HIVServices` and `venue_list` do not define any primary keys
- In the dev protocol, delete UIDs from "previousInterview". The first node generator should behave well (e.g., not contain an infinite number of the same node as you drag nodes out).
- Changing the PK prop ("uid") used internally by the app shouldn't matter
- Reloading the app should not change results shown in Search or other node generators
- View generated data by downloading the export file at the end

Edit: this also fixes #605, as UIDs should override anything in additionalAttributes (which cannot be unique).


public/protocols/development.netcanvas/protocol.json
```diff
@@ -737,212 +742,213 @@
       ]
     },
     "schoolPupils": {
+      "nodes_primary_key": "studentId",
```
wwqrd: Should we stick with the camel case convention?

wwqrd:
I see that we have chosen to make separate uids the responsibility of the user, but I think it's too opaque.

If:

`{ studentId: 1, primaryKey: 'studentId' } == { teacherId: 1, primaryKey: 'teacherId' }`

Then the user simply won't see the other node once one is added to the network.

bryfox:
Good point; that would certainly be unexpected.

I thought the primary intent for specifying external PKs was to be able to identify objects that should be treated as equal between different data sets. In which case we do still want to support something like: `{ subjectId: 1, primaryKey: 'subjectId' } == { id: 1, primaryKey: 'id' }`, where subjectId=1 and id=1 should be treated as the same.

Perhaps the API for describing this is confusing? Would something like this be better?

```
mergeDataOnPrimaryKeys: { previousInterview: "uid", students: "studentId" },
externalData: { previousInterview: {}, students: {} }
```

bryfox:
Also, based on your example, I think I should no longer default to any PK. i.e., we won't automatically assume "uid" is unique.

rebeccamadsen:
I think `ExportData` needs to use the primary key. It's looking for `id` right now and falling back to index because it's not there.

lines 149-150 should change to
```
    if (dataElement[NodePK]) {
     domElement.setAttribute('id', dataElement[NodePK]);
```

I haven't tried all the scenarios yet. I think probably the `excludeList` sent to `generateKeys` and `addElements` should include the NodePK value instead of `id`.

If I find anything else I'll comment again. So far it's working well though!

bryfox:
@rebeccamadsen thanks, I pushed those fixes. I agree `excludeList` needs to include UID now; I added a specific test for that. I also removed 'id' from the edges, which is never generated by the app; that's the cause of the other test's snapshot diff.

@wwqrd what is the panels/comparison function?

jthrilly:
> We could update the panels/comparison function to use an arbitrary property on the node? - That might even mean that PKs aren't needed?

PKs are definitely needed, particularly for external data which has a lot of functionality we haven't made yet.

> I haven't tried all the scenarios yet. I think probably the excludeList sent to generateKeys and addElements should include the NodePK value instead of id.

This is reminding me very much of the discussion we had about private vs public properties. At the time (https://github.com/codaco/Network-Canvas/pull/443#issuecomment-381155177) I suggested  putting all "private" nc properties inside something like `_private` on the node. This would include things like the uuid, the property for creation order, and any additional properties that are convenient for the app/interfaces but that we don't need to export and that aren't _actual_ properties of the nodes in a graph theoretic sense.  Can anyone think of a reason not to do this?

bryfox:
Based on a conversation with @jthrilly I've removed support for defining existing PKs in external data ("separate uids", above), and no longer assume that something labeled `id` or `uid` is a PK. We may add that convenience once we've better defined requirements. (Original use case was [mentioned in #397](https://github.com/codaco/Network-Canvas/issues/397#issuecomment-369609115)).

When using the development protocol, you'll once again see seemingly-duplicated nodes between NG Closeness and NG Classmates.

I've also renamed the internal PK to `_uid` for now (and updated the Action props for clarity). Isolating model data can happen as part of related issues (#446, #587, #596).

rebeccamadsen:
These updates look okay to me.

wwqrd:
RE:

> @wwqrd what is the panels/comparison function?

In hindsight I think I may have been making the same point as @rebeccamadsen?

jthrilly:
I created #620 to discuss the idea of `_private` on nodes and edges.

PR #617, opened 2018-07-25
Re-order dev protocol

bryfox:
This moves the initial static screens to the end of the interview. I spend a lot of time clicking past them or opening the stage menu to be able to work on other frequently-changing screens. (I've added "click the thing" to my music library in case I miss hearing it all the time.)

### DIFF
--- a/public/protocols/development.netcanvas/protocol.json
+++ b/public/protocols/development.netcanvas/protocol.json
@@ -1042,57 +1042,6 @@
   },
   "stages": [
     {
-      "id": "welcome",
-      "label": "Welcome",
-      "type": "Information",
-      "title": "Title for the screen",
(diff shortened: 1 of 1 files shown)


rebeccamadsen:
I'm okay with this change. Some in progress sessions may have incorrect stage/prompt indexes with this change. This is not a problem, just something to point out if anyone has issues and hasn't cleared redux lately!

wwqrd:
Shouldn't the routing features make this unnecessary? I would expect hot reload to stay on the same stage?

@bryfox If you want any other soundbites let me know.

bryfox:
Reloading stays on the same stage & is working well. I guess I've been doing a lot of "reset app data" this week in particular.

PR #622, opened 2018-08-01
When a node is removed, remove any connected edges

### LINKED ISSUE #437: Edge still exists when a node is deleted
When a node is deleted, existing edges with that node are not deleted. This means that when a node with that same "id" is added again (which can happen if all the nodes are deleted), the edge just shows up in the sociogram. It also means that any queries against the network can use edges the user can't even see as part of a matching query for skip logic.

I think deleting any edges with the deleted node would probably make sense for both of these situations.

bryfox:
Fixes #437.

When a node is removed from the network, remove any edges that connected it.

### DIFF
--- a/src/ducks/modules/__tests__/network.test.js
+++ b/src/ducks/modules/__tests__/network.test.js
@@ -99,6 +99,17 @@ describe('network reducer', () => {
     });
   });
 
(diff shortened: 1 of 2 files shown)


rebeccamadsen:
Looks good to me!

Issue #549, opened 2018-06-07
Harmonize NodeList and CardList components


jthrilly:
We should harmonize/combine these components, since they provide very similar functionality. In fact, they differ only in the entities that they render, which could conceivably be passed to the component as props.

Key areas of functionality:

- Sorting. The (optional) ability to sort by one or more properties. Implemented using the `sorty` library.
- Filtering. (optional) ability to filter the list using fuzzy matching of one or more properties of each entity.
- Scrolling. Scroll behaviors and styles associated with a list that might overflow the screen. Potentially also supporting #548
- Transitions. NodeList uses an animated transition to reflect changes in data. CardList should benefit from this, too. Could the in and out animations be specified as props somehow?

Areas of divergence: 

- Drag and drop. NodeLists use it, but CardList does not (although this feature is required for the resume interview screen).
- Color. NodeList provides the ability to specify the color of nodes contained within. I think this should be deprecated in favor of using the node type to detect color. Overriding is still possible through CSS.

jthrilly:
This was discussed in #555 in relation to sorting.

The NameGeneratorList interface uses `<CardList />` rather than `<NodeList />` and implements filtering and sorting on top of that. The [basic api](https://github.com/codaco/Network-Canvas/blob/cfd0fa93fc6ad33fb97b3b72231912966182e3af/public/protocols/development.netcanvas/protocol.json#L950-L964) looks like this:

```
          "sortOptions": {
            "sortOrder": {
              "name": "ASC"
            },
            "sortableProperties": [
              {
                "label": "Pupil Name",
                "variable": "name"
              },
              {
                "label": "Pupil Age",
                "variable": "age"
              }
            ]
          }
```

What I think we should do, is refactor this interface to use

wwqrd:
It looks like `<NodeList />` and `<CardList />` are both simple views for a list of nodes.

In the case of `<NameGeneratorList />`, sorting an filtering is handled by a parent component called `<ListSelect />` which contains UI controls and `<CardList />, which receives the managed node array:

```
// pseudocode
const ListSelect = () => (
  <div>
    <SortControls />
    <CardList nodes={sortedNodes}/>
  </div>
);
```

So I think what we want here is:

1. Generalise `<ListSelect />`, to maybe take a component argument:

`<ListSelect component={NodeList} />`

Which will require harmonising `<NodeList />` and `<CardList />` to have the same interface where possible, perhaps passing down any specific props:

[...]

PR #623, opened 2018-08-01
Reduce node re-rendering calls in sorted lists

bryfox:
This PR includes a few optimizations for node rendering:

- returning equivalent node objects from the sorter (removing the internal `createdIndex` prop)
- some minor adjustments to the sorter to avoid two extra mappings when they're not needed (I believe lodash already uses indexes to implement a stable sort internally). It's not a big deal, but could matter with external data sets.
- Update the comparison check in NodeList (https://github.com/codaco/Network-Canvas/commit/350dd01683faa41b02718c150caa570203ee7204). When nodes were sorted, this check always returned false.
  + with the above change to keeping node props unchanged, this isn't as big a deal, but it still seems cleaner
- I also updated tests to clarify existing LIFO/FIFO behavior

I think these changes are still equivalent to what we want, but please review carefully!

Here are some behaviors on master I was trying to solve:

- during list scrolling, the list continually re-sorts, and all nodes re-render
- When there are multiple lists (e.g., with an ordinal bin), changing the props on one node (say, by moving bins) causes all nodes to re-render
  + this actually happens 3+ times, including on drag start and drag end.

To test, you can log calls to a Node's render(), and/or measure FPS on a heavily-populated list. The ordinal bins in the development protocol are a good place to start.

### DIFF
--- a/src/utils/sortOrder.js
+++ b/src/utils/sortOrder.js
@@ -1,10 +1,10 @@
-import { omit, orderBy } from 'lodash';
+import { orderBy } from 'lodash';
 
 /* Maps a `createdIndex` index value to all items in an array */
 const withCreatedIndex = items => items.map((item, createdIndex) => ({ ...item, createdIndex }));
 
 /* all items without the 'createdIndex' prop */
-const withoutCreatedIndex = items => items.map(item => omit(item, 'createdIndex'));
+const withoutCreatedIndex = items => items.map(({ createdIndex, ...originalItem }) => originalItem);
 
 /* property iteratee for the special case "*" property, which sorts by `createdIndex` */
 const fifo = ({ createdIndex }) => createdIndex;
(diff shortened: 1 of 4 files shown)


src/utils/sortOrder.js
```diff
@@ -1,34 +1,47 @@
-import { orderBy } from 'lodash';
+import { omit, orderBy } from 'lodash';
 
 /* Maps a `createdIndex` index value to all items in an array */
 const withCreatedIndex = items => items.map((item, createdIndex) => ({ ...item, createdIndex }));
 
-/* property iteratee for the special case "*" property, which sorteds by `createdIndex` */
+/* all items without the 'createdIndex' prop */
+const withoutCreatedIndex = items => items.map(item => omit(item, 'createdIndex'));
```
wwqrd: Could use destructuring:
`const withoutCreatedIndex = items => items.map({ createdIndex, ...withoutIndex  } => withoutIndex);`

src/utils/sortOrder.js
```diff
@@ -1,34 +1,47 @@
...
+
+/* property iteratee for the special case "*" property, which sorts by `createdIndex` */
 const fifo = ({ createdIndex }) => createdIndex;
 
 /**
  * Returns a configured sorting function
- * @param {array} sortConfiguration - The title of the book.
- * @param {object} variableRegistry - an object containing variables for the node type as specified
+ * @param {Array} sortConfig - list of rules to sort by
```
wwqrd: 😳

jthrilly:
This is massive, Bryan. I knew that re-rendering was an issue from the console output, but had no idea how to approach fixing it. Those performance gains look huge. 

Now we have this in place, homogenization with card list (#549) should be more efficient. 

> Update the comparison check in NodeList (350dd01). When nodes were sorted, this check always returned false.

Just checking: does this comparison exist in other sorted lists? The bucket components, and the card list for example?

bryfox:
> does this comparison exist in other sorted lists? The bucket components, and the card list for example?

Looks like it does in the Ordinal Bin Bucket. I don't think it matters either way there. Theoretically, we may only need to sort once, since you can only remove from the bucket. But it doesn't cost much; I'll match the NodeList check for consistency, and future-proofing.

rebeccamadsen:
Looks good as far as I can tell!

PR #624, opened 2018-08-03
Fix CSP for server pairing

bryfox:
This reverts an overly-aggressive CSP (connect-src) change I made in #613. The intent there was to prevent workers from making arbitrary network requests; worker scripts inherit the CSP of the app.

Of course, the app needs that flexibility in order to pair with arbitrary Servers, so I'm reverting for now. I'd like to investigate a way to keep that limitation on workers in the future; the CSP needs a review & update in general, so I'll open a separate issue for that.


rebeccamadsen:
Makes sense to me.

PR #626, opened 2018-08-03
New menu system, first iteration

rebeccamadsen:
This is a step towards a new menu system (see #245). It moves stage navigation to the left side of the screen, and shifts a few other things to make room (sociogram instructions, node trash bins, menu buttons). 

This changes the forward navigation button to advance through prompts until the final prompt of a stage, and then advance to the next stage. Backward navigation returns to the previous stage, prompt 1. Prompt "pips" were not changed here, and so still work for cycling prompts in this iteration.


jthrilly:
All the content being off-centre was really bugging me, so I removed the physical size of the container for the menu buttons. The buttons themselves are absolutely positioned, and when we refactor the structure of this will be totally different anyway.

I then manually added some margins to the interfaces to cover this. Again, we can factor this out at a later date. 

The only two issues remaining are:

- Styling for the autocomplete list.
- Button positioning on the start screen (looks hideous, and the stages menu probably shouldn't even be shown.

PR #636, opened 2018-08-07
initialize checkbox value

### LINKED ISSUE #634: checkbox not initiating with value
A checkbox in a form doesn't start with the correct value. `value` is set, but `checked` does not reflect that value. For example, if you have a form with a `checked` variable, check the box and save the node. When you reopen the edit node form, the box is not checked but css shows a `true` value.

rebeccamadsen:
Fixes #634. To verify, check and uncheck the checkbox on the development protocol's "NG Inputs" screen. It should be able to reopen the form and have the correct value for the checkbox.

### DIFF
--- a/public/protocols/development.netcanvas/protocol.json
+++ b/public/protocols/development.netcanvas/protocol.json
@@ -207,6 +207,11 @@
             "description": "Human readable description",
             "type": "boolean"
           },
+          "checkboxVariable": {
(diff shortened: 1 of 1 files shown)


rebeccamadsen:
This points to master instead of the ui branch, which solves the checkbox issue. I don't notice any other issues with this update.

PR #639, opened 2018-08-08
update networkQuery to use _uids

rebeccamadsen:
This updates networkQuery to use _uid instead of id. You would see this most with edges. The development protocol has a skip logic rule to match against any edge of type "friend" that exists, and it fails in master because edges use _uid now, but should work here.


bryfox:
Looks good! I tested with 3 connected nodes.

PR #640, opened 2018-08-08
Handle invalid input types to Radar

### LINKED ISSUE #632: Sociogram: Radii set incorrectly
If a Sociogram's "concentricCircles" is a string (which is currently the default in Architect), then the background won't render correctly. (e.g., will render "10" + 1 == "101" circles.)

- [x] parse value to int
- [x] Probably add some NaN & bounds checking (what's a sensible limit?)
- export number from Architect (see https://github.com/codaco/Architect/issues/204)

bryfox:
This fixes rendering of the concentric circles when input type isn't a number.

Coerce strings to numbers when possible, but check for validity.

If NaN, the propTypes warning should still be logged to indicate why the circles are missing.

Fixes #632

### DIFF
--- a/src/containers/ConcentricCircles/Radar.js
+++ b/src/containers/ConcentricCircles/Radar.js
@@ -27,15 +27,20 @@ const weightedAverage = (a, b, factor = 1) =>
   zipWith(a, b, (c, d) => ((c * factor) + d) / (1 + factor));
 
 const Radar = ({ n, skewed }) => {
+  const num = parseInt(n, 10);
+  if (isNaN(num) || !num) {
+    return null;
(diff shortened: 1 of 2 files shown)


rebeccamadsen:
This works for me! Although, is this not an Architect issue, that Architect should be outputting an integer here and not a string?

bryfox:
Thanks for the reminder — I opened https://github.com/codaco/Architect/issues/204 to cover the Architect side of this. Since we support protocols edited outside of Architect, I think it makes sense to handle here too.

rebeccamadsen:
Sounds good to me. :)

PR #641, opened 2018-08-08
Better handling of syntax errors in user scripts

### LINKED ISSUE #627: Node labels in chrome
The node labels on the education protocol are blank on Chrome, but are working fine on Android and electron. Node labels for the development protocol seem fine in Chrome.

bryfox:
Fixes #627.

The reported bug is with Chrome development, but this would be a problem with syntax errors in a user script as well.

If there's any syntax error in a worker script, then all clients making use of that worker will be notified of the failure, and the worker agent will reject any future requests. In the case of node labeling, this allows the Node container to fall back to the built-in label function.

Additionally, to make development mimic production a little better, I've disabled `historyApiFallback` in the dev server, which is what causes it to return index.html instead of 404s. This was the immediate source of the problem in #627. (This also uncovered a reference to a missing favicon, so I deleted that.)

### Testing

1. Browser: `npm start` and view the education protocol in a browser. See that nodes are no longer blank.

2. Any client: add a syntax error to the development worker script, and start a session with it. See that node labels use the original function (no emoji are visible).

### DIFF
--- a/config/webpackDevServer.config.js
+++ b/config/webpackDevServer.config.js
@@ -70,11 +70,9 @@ module.exports = function(proxy, allowedHost) {
     https: protocol === 'https',
     host: host,
     overlay: false,
-    historyApiFallback: {
-      // Paths with dots should still use the history fallback.
-      // See https://github.com/facebookincubator/create-react-app/issues/387.
-      disableDotRule: true,
-    },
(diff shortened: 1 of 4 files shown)


rebeccamadsen:
Works for me! I tested chrome, electron, and android.

PR #643, opened 2018-08-10
Make builds & CI faster

bryfox:
- Combines Travis jobs: We don't have the benefit of parallelism, and jobs are dominated by
boot & install overhead. (Contrast the `chore/travis-stages` run time [with others](https://travis-ci.org/codaco/Network-Canvas/builds).)
- Use [thread-loader](https://webpack.js.org/loaders/thread-loader/) for some loaders. This is used in more recent versions of create-react-app, and adopted in Server & Architect. On my machine, this speeds up dev & prod builds by 20-25%. I haven't seen any issues with it yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,5 @@ install:
   - npm install
 jobs:
   include:
-    - stage: prebuild
-      script: npm run test
-    - script: npm run lint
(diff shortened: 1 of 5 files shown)


rebeccamadsen:
This works for me.

PR #644, opened 2018-08-10
WIP: protocol schema & script

bryfox:
Please do not merge.

Right now, this is just an example of how we might start using JSON Schema.

My first desired use case is to be able to quickly validate protocols that I create in Architect, to make manual QA faster.

There are two new npm scripts:

- `npm run generate-schema` creates a JSON schema for protocols
- `npm run validate-protocol` runs validations using the above schema against a given protocol (or the development protocol by default)

By running both of the above, you can see that the development protocol validates. If you then modify the development protocol and re-run the validation step, you can cause it to fail. Examples:

Remove name & version from development protocol:
```
development.netcanvas has errors:
- dataProtocol should have required property 'name'
- dataProtocol should have required property 'version'
```

Remove id from the first stage:
```
development.netcanvas has errors:
- dataProtocol.stages[0] should have required property 'id'
```

### Notes

The intent here is to validate the general format of a protocol; which is also specified in the wiki. Protocols can still be internally inconsistent (e.g., lacking a variable registry definition), but that's out of scope here, at least for now.

Eventually, we can run validations in the apps at runtime to catch some classes of errors.

We may also want to generate the protocol.json documentation automatically from these schemas to avoid inconsistencies from manually editing the wiki.

### Next steps

Right now, this uses the development protocol to auto-create a schema, but the schema is overly-specific. (You may have noticed that running the above on the education protocol finds spurious errors.) I like the idea of an auto-generated schema from an example (at least for now), so I'd like to create a generic `protocol.json` which captures the format requirements. This could be used as a singular example in the wiki as well.

We also discussed moving this into a submodule to be shared, perhaps along with the factory protocols.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@ Icon^M^M
 
 .devserver
 config.xml.original
+
+# Ignore auto-generated schema
+protocol.schema
--- a/package-lock.json
+++ b/package-lock.json
@@ -133,15 +133,15 @@
             "dev": true
         },
         "ajv": {
-            "version": "5.5.2",
-            "resolved": "https://registry.npmjs.org/ajv/-/ajv-5.5.2.tgz",
-            "integrity": "sha1-c7Xuyj+rZT49P5Qis0GtQiBdyWU=",
+            "version": "6.5.2",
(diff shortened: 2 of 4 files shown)


bryfox:
I'll continue work in a long-running feature branch here; #697 captures the next steps.

PR #646, opened 2018-08-23
https clients for NC Server

### LINKED ISSUE #537: Store Server public key during pairing process
This is the client-side equivalent of https://github.com/codaco/Server/issues/105.

Client should store/trust the server's public cert delivered during the trusted pairing process.

### LINKED ISSUE #538: iOS: Validate SSL cert during API handshake
Use the stored server cert from #537 to enable https requests to the Server API. Needs research; likely bundled with Android support in a plugin.

### LINKED ISSUE #539: Android: Validate SSL cert during API handshake
Use the stored server cert from #537 to enable https requests to the Server API. Needs research; likely bundled with iOS support in a plugin (see #538).

Some relevant docs:
- https://developer.android.com/training/articles/security-ssl#SelfSigned
- https://developer.android.com/reference/javax/net/ssl/X509TrustManager

### LINKED ISSUE #540: Electron: Validate SSL cert during API handshake
Use the stored server cert from #537 to enable https requests to the Server API. Needs research; [see cert error docs](https://electronjs.org/docs/api/app#event-certificate-error).

bryfox:
This adds new clients for communicating with https endpoints on a paired Server. Because we have no central infrastructure, Server uses a self-signed cert which must be trusted by a client during pairing. Once paired, https calls ensure that this certificate is used in trust evaluation; users should not see a self-signed warning. Clients do not blindly trust any https endpoint.

### TODO
- [x] Cordova: SSL cert evaluation on file downloads
- [x] Device auth on file downloads

### Testing

The shape of servers is redux has changed, and is not backward-compatible. To use this branch, you'll need to remove any existing paired servers (i.e., Reset Data).

Test against the [feature/support-https-clients](https://github.com/codaco/Server/tree/feature/support-https-clients) branch of Server. The only difference on this branch is that all API calls to a paired Server (get protocols list, get protocol file, & post session data) should use https transparently to the user.

Resolves #537. Resolves #538. Resolves #539. Resolves #540. (I think it will be easier to test & review all platforms at once.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,8 +16,12 @@ yarn.lock
 build.json
 
 *.zip
-*.pbxproj
+plugins/**/*.pbxproj
 Icon^M^M
 
 .devserver
 config.xml.original
+
+# Xcode (Cordova plugin)
+xcuserdata/
+IDEWorkspaceChecks.plist
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ React Scripts
 
 This project currently requires version `8.11.1` of node, and version `5.8.0` of npm. These are the only supported versions for this project.
 
-** NOTE: ** npm 5.8.0 is not installed by default with node 8.11.1. You will need to use `npm install -g npm` to do this. Test which version of npm you are using by typing `npm --version`.
+** NOTE: ** npm 5.8.0 is not installed by default with node 8.11.1. You will need to use `npm install -g npm@5.8.0` to do this. Test which version of npm you are using by typing `npm --version`.
 
 As a convenience, the repository contains a `.node-version` file that enables convinient use of a node environment manager.
 
--- a/config.xml
+++ b/config.xml
@@ -68,4 +68,5 @@
     <plugin name="cordova-plugin-device" spec="^2.0.2" />
     <engine name="android" spec="^7.1.0" />
     <engine name="ios" spec="^4.5.4" />
+    <plugin name="cordova-plugin-network-canvas-client" spec="./cordova-plugin-network-canvas-client" />
 </widget>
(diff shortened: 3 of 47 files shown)


wwqrd:
This ran fine for me on Electron, but not ios/android.

I'm wondering if it's my system / a problem with using emulators?

iOS:
screen shot 2018-08-30 at 11 44 53

Android:
screen shot 2018-08-30 at 12 08 16

Interesting it seems to still think it's paired? (android):
screen shot 2018-08-30 at 12 08 41

Server:
screen shot 2018-08-30 at 12 01 32

bryfox:
Emulators/simulators should work fine.

- Did you reset the NC app data (to remove any existing paired servers)?
- "Server could not be reached" is an unusual error for Cordova; are you running in dev mode (`npm run ios:dev`)? If so, did you restart your frontend server (`npm run start:ios`)? Android would need to be restarted in the same way.

If neither of those are issues, then I'll need to dig a little deeper...

wwqrd:
I got that working 👍 
- Needed to reset Android app data
- Also was using the following to run: `npm run build; cordova run ios/android`, which seems to have contributed.

bryfox:
`cordova run ios` should work as well (I just confirmed); that tests the production build. However, the build step for that was recently changed to `npm run build:ios` (need to add the platform target).

I'll see if there's a low-touch way to keep backwards compatibility with existing app data. Since we're going to move away from supporting multiple Servers, I didn't worry much about it, but that seems like an easy way to cause confusion in the next release.

wwqrd:
I don't think there's any need to add backwards compatibility until we start doing beta releases?

Resetting app data isn't a big ask - I'd just thought I hadn't paired already and it would be okay without it.

bryfox:
Sorry, just saw your follow-up comment. I agree, but the [change](https://github.com/codaco/Network-Canvas/pull/646/commits/7309e68e4f48812b49e7bcb6097b62ac85f58183) turned out to be pretty trivial anyway and didn't add any cruft.

Issue #642, opened 2018-08-10
node form 'add another' does not return to top


berniehogan:
when adding another node using the 'add another' check box (which I still think is weirder than something like 'add another' as a separate button rather than a toggle), and the 'finish' button is depressed then:
- the original node is added (good)
- the form is reset to default (good)
- the form does not return to the top of the pane (bad). 

I would both request that 'add another' is a button rather than a toggle and that when pressed, the focus returns to the top of the node form.

bryfox:
- [x] focus returns to the top of the node form (#647)
- [ ] make 'add another' a button rather than a toggle

jthrilly:
We actually had this as a button previously, and it didn't work with the full screen form. Let's keep the checkbox for now.

PR #647, opened 2018-08-29
Fix node form scroll position after submit

bryfox:
This resolves the buggy behavior described in #642. When an "add another" form is scrolled down and submitted, its scroll top should be reset to 0. (It does not change the checkbox to a button; also described in that issue.)

### Testing

You can test this on the "NG Inputs" stage of the development protocol: fill out the form, check "add another", and submit the form. When submitted, the form should be cleared and return to the top.

### Notes

This uses a [forwardRef](https://reactjs.org/docs/forwarding-refs.html) to provide scrolling support, which required updating enzyme to get support in tests. Side effect: all snapshots in the app have changed. Those changes are encapsulated in https://github.com/codaco/Network-Canvas/commit/70ae45775693617a19f6284f4b300239209f0a86; it may be easier to review the other commits separately.


src/components/Modal.js
```diff
@@ -15,7 +17,11 @@ class Modal extends Component {
 
   scrollContentsToTop() {
     if (this.contentRef.current) {
-      this.contentRef.current.scrollTop = 0;
+      animejs({
+        targets: this.contentRef.current,
+        scrollTop: 0,
+        duration: getCSSVariableAsNumber('--animation-duration-slow-ms'),
```
wwqrd: Adding an `easing: getCSSVariableAsObject('--animation-easing-js'),` removes the 'bounce' that animejs defaults to.

bryfox: Thank you — fixed.

PR #649, opened 2018-09-05
Use shared API package

bryfox:
This replaces 'shared-api' source and the dependency on libsodium-wrappers with the [secure-comms-api](https://github.com/codaco/secure-comms-api) package. libsodium will be installed as a transitive dependency by npm. Make sure you run `npm install` to get the new dependency.


rebeccamadsen:
This works for me.

PR #650, opened 2018-09-06
New menu system, second iteration.

rebeccamadsen:
This is the next step for #245. This updates the navigation system to use a progress bar as a visual indicator throughout an interview, adds animation for stage changes, and fixes the "flashing" to the first prompt on a stage change.

Note that this does not include new menus for stages and settings, which will be iteration three, but it does remove the buttons. Instead, click on the timeline to open the stages menu. Settings menu is the last item in the stages menu.


wwqrd:
This is looking fantastic!

src/containers/App.js
```diff
@@ -25,8 +26,8 @@ const App = props => (
   })}
   >
     <div className="electron-titlebar" />
-    <SettingsMenu hideButton={props.isMenuOpen} />
-    <StageMenu hideButton={props.isMenuOpen} />
+    <SettingsMenu hideButton />
```
wwqrd: Is this attribute redundant?

rebeccamadsen: Yes, this was just trying to get the menu back in with minimal changes. But it wouldn't be tricky to clean that up a bit.

bryfox:
I haven't reviewed code (it looks like wwqrd did), but this is working well for me.

PR #651, opened 2018-09-07
Remove fallback form behavior from protocol

### LINKED ISSUE #592: Node form should be optional on name generator
A protocol may intend to not show the “add new node” button on a name generator.

- [x] Make 'form' an optional prop
- [x] Update [docs](https://github.com/codaco/Network-Canvas/wiki/Name-Generator#interface-specific-api)
- [x] Document a way in protocol to specify overriding the default form with "no form"
- [x] Remove fallback behavior in implementation and documentation

bryfox:
Resolves #592.

This removes the feature that defaulted form to a form definition matching the node type. If the form prop is falsy, then no form should be displayed. (See https://github.com/codaco/Architect/issues/136). The Component prop is still optional, since `null` is treated as missing by PropTypes.

I've [updated the documentation](https://github.com/codaco/Network-Canvas/wiki/Name-Generator/_compare/6433ed7b9b012597e37a228f3a5ed1a9fc299ce8...a039a973b26847dd283b887ba2cd82ecb6320730) to match.

### DIFF
--- a/src/selectors/__tests__/forms.test.js
+++ b/src/selectors/__tests__/forms.test.js
@@ -1,0 +1,33 @@
+/* eslint-env jest */
+
+import { makeRehydrateForm } from '../forms';
+
+describe('forms selector', () => {
+  describe('makeRehydrateForm', () => {
+    const state = {
+      protocol: {
+        forms: { person: {} },
+      },
+    };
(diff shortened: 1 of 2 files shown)


rebeccamadsen:
This looks good to me.

PR #654, opened 2018-09-10
fix modal positioning

rebeccamadsen:
This fixes the positioning and leverages portals, so Modal can effectively be higher in the dom and take z-index precedence over the timeline.


bryfox:
Looks good — I didn't expect the real solution to be simpler than the CSS workaround :)

PR #655, opened 2018-09-10
Fix model data issues

bryfox:
This fixes remaining test failures, the CardList issue with Resume Interview, and some other issues with the secondary Name Generators (List & Autocomplete). I tried to stick to the new direction as best I could, but had to revert part of addNodes() to support those other name generators.

It's still not immediately obvious where the attributes sub-object needs to be dealt with in all cases, but I tried to add some test coverage for the problem spots I saw.

Summary by commit; they should be individually pickable:

- Fix session & node card lists
    - CardList should provide an element of the supplied data collection for formatters.
- Fix Search (no results found)
    - Search lib must be told where to find the nested attributes
- Fix node creation on NGList & NGAutocomplete
    - Nodes already exist with model attrs inside attributes; no need to pass as a separate arg. The second arg to addNodes() should be the static set of "additional attributes" so that the batch NG interfaces can create without first inserting that information into each node.
- Fix filtering on NGList (was not wired up)
- Fix ListSelect state
    - Unconditionally deriving props cleared the sort too often; for example, when selecting an item, the list would re-sort and potentially leave the just-selected item out of view if it was scrolled.
- Re-add null check to label func
- Fix mock data

### DIFF
--- a/src/components/CardList.js
+++ b/src/components/CardList.js
@@ -6,7 +6,7 @@ import cx from 'classnames';
 import { scrollable, selectable } from '../behaviours';
 import { Card } from '.';
 import { Icon } from '../ui/components';
-import { nodePrimaryKeyProperty, getNodeAttributes } from '../ducks/modules/network';
+import { nodePrimaryKeyProperty } from '../ducks/modules/network';
 
 const EnhancedCard = selectable(Card);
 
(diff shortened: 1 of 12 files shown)


rebeccamadsen:
This looks good to me.

PR #658, opened 2018-09-11
Fix node updating during drag

### LINKED ISSUE #657: Sociogram edges do not update (re-render) when dragging
Steps:
1. Go to Sociogram interface
2. Add two nodes
3. Connect with an edge
4. Drag one node

Expected: edge remains connected while dragging

bryfox:
Fixes #657.

"meta" refers to node information; this checks the `.attributes` sub-object for layout info.

I don't see a straightforward way to add a regression test without a somewhat substantial refactor, so I've skipped that for now.


rebeccamadsen:
Looks good!

Issue #371, opened 2018-02-14
Node occlusion on the sociogram screen 


berniehogan:
The layout screen currently permits one to place a node anywhere on the canvas. However, some UI elements are floating on top. This means that when you drop a node over the panel in the lower left hand corner it will stay there and can no longer be clicked or retrieved without "resetting" all of the node places. 

Some options: 
1. The panel in the lower left can itself be dragged around. 
2. There's a hit window such that if the node is dropped within that window it goes back to the original coordinates from which it arrived. 
3. It's laid on top of the dialog window. 

I believe Josh was promoting the second behaviour. He asked me to submit this as a bug. 

![screenshot 2018-02-14 17 59 49](https://user-images.githubusercontent.com/1711620/36219952-f2a61dde-11b0-11e8-81f3-adf308461ae1.png)

pfsj:
The area where Nodes originate from in the top right corner should also probably be ineligible for dropping nodes until there are no remaining nodes to be added. They currently get placed behind the remaining nodes and cannot be selected until all nodes are removed which seems awkward. 

<img width="1440" alt="screen shot 2018-02-22 at 2 37 04 pm" src="https://user-images.githubusercontent.com/10948594/36562854-e01bb024-17dd-11e8-854c-a1f5071dd2cd.png">

rebeccamadsen:
Knowing that the collapsing instructions issue was also proposed, is this the preferred solution? That the prompt window and the node bin are hit zones that can't be dropped on? Is the reset button also a hit zone, or is it too small to be of concern?

jthrilly:
I think collapsing instructions are also desired, but node occlusion does need to be looked at as well. However, given that we haven't specified this very well I am going to bump the feature to the next release.

wwqrd:
Relates to: https://github.com/codaco/Network-Canvas/issues/469

wwqrd:
Just some technical background first: drop areas are defined as x, y, width, height, and have no concept of occlusion.

I think probably the best way to handle this would be to add the concept of *Obstacles*, this could be added to components we want to block movement.

Before any move or drag events are fired, we could check if they fall within any of these regions, if so they won't set off the drop region events.

rebeccamadsen:
Related ideas from #614:

> We could also connect this up to MonitorDragSource and auto hide [the instruction panel and origin bucket]

> It's almost like we need a "show desktop" button or a "placement mode".

bryfox:
[...]

PR #660, opened 2018-09-11
Fix sociogram centering

### LINKED ISSUE #659: Sociogram: Cannot pick up nodes dropped around timeline
Two versions ago, the sociogram was offset from the left edge of the viewport; now it extends to the left edge and overlaps with the timeline. If a node is dropped there, it cannot be picked up again; the timeline (including transparent area) is used as the side menu trigger.

bryfox:
Fixes #659.

This cleans up centering by moving style definitions to a mixin, and prevents sociogram nodes from begin dropped on top of the timeline/menu area. The Sociogram also gets a little bit of T/R/B margin so that node labels aren't cut off when dropped at the edges.


rebeccamadsen:
I'm okay with this, but I think @jthrilly wants sociogram to take the whole screen space and use "exclusion zones" as the fix for #371 eventually, so I defer to him.

bryfox:
Ah, I didn't realize this was intentional. If we want to keep the occlusion for now, I can update this to revert the Sociogram centering but keep the other cleanup; that's going to make my life easier in some future work.

jthrilly:
So in general I did want to keep the interfaces centred, but I can see the logic of having them offset against the timeline. I think we should try it the way it is implemented here and see if others find it as disturbing as I do! 😅 

The main "fix" here (preventing node drop on the timeline) works nicely, with the exception that returning the node to its last location when dropped on an invalid location is a bit jarring. Probably we will fix the underlying behaviour with #371, but just to document here what the ideal behaviour should be:

- Nodes hovering over an exclusion zone should enter an invalid state
- If dropped on an exclusion zone, they should animate back to their last valid position

bryfox:
For centering, I guess I followed the lead of the Name Generator. If it's disturbing, we could adjust [the right margin](https://github.com/codaco/Network-Canvas/pull/660/files#diff-02ad984811a3111f39d328f15ebb70f5R15) (and `interface-prompt` if needed), then special case the NG. Or revert this :) ...in part, I was just trying to avoid coupling every interface individually to the timeline.

PR #662, opened 2018-09-11
Fix prompt setting when index/ID is 0

### LINKED ISSUE #656: Prompt errors when advancing screens too quickly
If I advance through prompts unreasonably quickly, sooner or later I'll get errors when loading some screens. (This might be reproducible by double-clicking the next button on the timeline; seems to be the speed of clicking, not the number of times I advance.)

This happens in the prior release (alpha.6) as well.

```
TypeError: Cannot read property 'additionalAttributes' of undefined
Function.mapStateToProps [as mapToProps]
src/containers/Interfaces/NameGenerator.js:186
```

```
TypeError: Cannot read property 'variable' of undefined
(anonymous function)
src/selectors/interface.js:97
```

bryfox:
Fixes #656.

One way to reproduce #656 (on master):
1. Start a session with the development protocol
2. Go to the third stage (ordinal bin with orange headers)
3. Double-click the 'next' timeline button (bottom left)

The promptIndex should be selected for the currentSession only when undefined (not 0). During the transition, the stage has already changed, and promptIndex may be out of range.

### DIFF
--- a/src/behaviours/__tests__/withPrompt.test.js
+++ b/src/behaviours/__tests__/withPrompt.test.js
@@ -1,0 +1,37 @@
+/* eslint-env jest */
+import React from 'react';
+import { shallow } from 'enzyme';
+import { createStore } from 'redux';
+
+import withPrompt from '../withPrompt';
+
+jest.mock('../../selectors/session', () => ({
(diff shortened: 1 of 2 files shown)


rebeccamadsen:
Good find! This looks great to me!

PR #665, opened 2018-09-11
Fix node updating

### LINKED ISSUE #664: Node updating broken in places
Some call sites for `updateNode` still use the old signature; it [was updated to take an attributes object](https://github.com/codaco/Network-Canvas/pull/652/files#diff-84d3293378051539622cb080f8dd1997R110).

Symptoms:
- Ordinal bin updates with wrong values
- Sociogram not reset

bryfox:
Fixes #664. See test cases there.

This fixes nested attribute updating in a few places to match behavior from alpha.6:
- Sociogram reset
- NameGenerator drag/drop
- Ordinal Bin


rebeccamadsen:
I haven't had a chance yet to look to see if there are any other cases we're missing, but these cases look resolved to me.

PR #666, opened 2018-09-12
Fix coordinates & inferred types in downloads

### LINKED ISSUE #663: attr.type missing for attributes with number values
When exporting a graphml file, "type" is missing for numbers. Also, "x" and "y" values aren't being added to the file. Gephi gives these warnings when importing this graphml file:
```
Issue{message=Attribute type not found for attribute 'special_category'. Set to default string., level=SEVERE}	SEVERE	
Issue{message=Attribute type not found for attribute 'age'. Set to default string., level=SEVERE}	SEVERE	
Issue{message=Attribute type not found for attribute 'discussFreq'. Set to default string., level=SEVERE}	SEVERE	
Issue{message=Attribute type not found for attribute 'contactFreq'. Set to default string., level=SEVERE}	SEVERE	
Issue{message=Attribute type not found for attribute 'numberVariable'. Set to default string., level=SEVERE}	SEVERE	
Issue{message=Attribute type not found for attribute 'ordinalVariable'. Set to default string., level=SEVERE}	SEVERE	
Issue{message=Attribute type not found for attribute 'Latitude'. Set to default string., level=SEVERE}	SEVERE	
Issue{message=Attribute type not found for attribute 'Longitude'. Set to default string., level=SEVERE}	SEVERE	
```

bryfox:
This fixes:
- where `getTypeForKey` looks for node attributes during export
- adding x/y (and closenessLayout) coords to the graphml download

I think this fixes #663. I've validated the generated XML but haven't been able to check using Gephi.


rebeccamadsen:
You were faster than me. :) I started a branch to try and fix this by using map on the network nodes, to pull out just the attribute and the "id"s, but your approach works as well. I think part of the complication is that nodes now have `attributes`, but edges do not, and most of these functions were written thinking that edges and nodes were pretty similar.

Anyway, I can verify this works in Gephi.

PR #678, opened 2018-09-12
Supply node UIDs to user workers

### LINKED ISSUE #668: Support network queries in node label worker?
Extracted from https://github.com/codaco/Network-Canvas/issues/620#issuecomment-409717568

A nodeLabel worker is given a copy of the network as one input for custom labels. On the development protocol (in alpha.6), the existing nodeLabelWorker illustrates this by appending "😎" when two nodes are connected on the sociogram.

This no longer works in alpha.7; edges associate nodes using UIDs, but those aren't included in the node data provided.

The [Counter use case](https://github.com/codaco/Network-Canvas/blob/v4.0.0-alpha.6/public/protocols/development.netcanvas/nodeLabelWorker.js#L27-L29) is another place this might crop up; we should not require a deep compare to find the node in a collection.

bryfox:
Fixes #668. Makes the sociogram cool again.

cool

- the ID is added at the same level as user props
- to minimize conflicts and self-document, uses "networkCanvasId" as key

The node info provided to the worker will look something like:
```
{
    "networkCanvasId":"2f92b04bbd7842d4ae90da7f2a1a178101f6ca51",
    "support_friend":true,
    "name":"Eustace",
    "age":"24"
}
```

I've updated [the Node Labeling wiki](https://github.com/codaco/Network-Canvas/wiki/Node-Labeling).

### DIFF
--- a/public/protocols/development.netcanvas/nodeLabelWorker.js
+++ b/public/protocols/development.netcanvas/nodeLabelWorker.js
@@ -11,9 +11,14 @@
  *
  * @param  {Object} data
  * @param  {Object} data.node All props for the node requiring a label
+ * @param  {string} data.node.networkCanvasId Unique ID for the node. Note that if your data
+ *                                            happens to already contain a property named
+ *                                            "networkCanvasId", your prop will take precedence,
+ *                                            and this cannot be used to identify edge connections.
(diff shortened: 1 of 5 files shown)


rebeccamadsen:
Coolness factor++. Looks good to me!

PR #679, opened 2018-09-12
Fix key on bucket node

### LINKED ISSUE #667: Node labels in the Sociogram NodeBucket flicker
The node bucket on the sociogram flashes incorrect data on master branch. I don't see this happening in alpha.6.

Before drop:

After drop (frame sequence):

bryfox:
Key is used to ensure that previous label is cleared when the next node
is ready to display. This is especially important for async node label
workers.

Fixes #667.


rebeccamadsen:
Good catch! Looks good.

jthrilly:
Great catch!

PR #680, opened 2018-09-12
Improve ordinal bucket item drag

bryfox:
This sets the `scrollDirection` prop on the wrapped nodes inside the Ordinal Bin Bucket to improve drag detection. The bucket is laid out horizontally but doesn't actually scroll, so `NO_SCROLL` is used.

I also removed the selectable/onSelect props since they're never used.

@rebeccamadsen thanks for the fix here.


rebeccamadsen:
This works for me!

PR #686, opened 2018-09-17
set a minimum time for protocol loading spinner

### LINKED ISSUE #670: Could we set a minimum display time for the protocol loading screen?
We need the loading screen in some cases, but when the loading happens almost instantaneously, the experience is jarring.

One solution to this would be to always show the loading screen for a minimum amount of time, even if the protocol is already loaded.

rebeccamadsen:
Resolves #670. This adds a minimum wait time to `LoadScreen`. To verify, add or refresh a protocol session and it should at least show the spinner for a second.

### DIFF
--- a/src/components/Transition/Fade.js
+++ b/src/components/Transition/Fade.js
@@ -4,34 +4,39 @@ import { Transition } from 'react-transition-group';
 import anime from 'animejs';
 import { getCSSVariableAsObject, getCSSVariableAsNumber } from '../../utils/CSSVariables';
 
-const duration = {
+const durationFast = {
   enter: getCSSVariableAsNumber('--animation-duration-fast-ms'),
   exit: getCSSVariableAsNumber('--animation-duration-fast-ms'),
 };
 
-const enterAnimation = {
+const durationSlow = {
+  enter: getCSSVariableAsNumber('--animation-duration-slow-ms'),
+  exit: getCSSVariableAsNumber('--animation-duration-slow-ms'),
+};
+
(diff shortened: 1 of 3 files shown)


src/components/Transition/Fade.js
```diff
@@ -4,42 +4,47 @@ import { Transition } from 'react-transition-group';
...
+const exitAnimation = duration => ({
   opacity: [1, 0],
   elasticity: 0,
   easing: getCSSVariableAsObject('--animation-easing-js'),
   duration: duration.exit,
-};
+});
 
-const Fade = ({ children, ...props }) => (
+const Fade = ({ children, isSlow, ...props }) => (
```
wwqrd: What do you think about implementing this more generally?

e.g.

```jsx
const Fade = ({ children, duration, ...props }) => (
  <Transition
    {...props}
    timeout={duration}
    onEnter={
      (el) => {
        anime({
          targets: el,
          ...enterAnimation(duration),
        });
      }
    }
    onExit={
      (el) => {
        anime({
          targets: el,
          ...exitAnimation(duration),
        });
      }
    }
    appear
    unmountOnExit
  >
    { children }
  </Transition>
);

Fade.propTypes = {
  children: PropTypes.any.isRequired,
  duration: PropTypes.object,
};

Fade.defaultProps = {
  children: null,
  duration: defaultDuration,
};
```

rebeccamadsen: Makes sense to me!

wwqrd:
👍

Issue #469, opened 2018-04-23
Node or edge is repositioned incorrectly after "offscreen" drop


bryfox:
[Desktop]

On the sociogram, if I drag a node & release the mouse button outside the window, then the node is positioned at its initial dragStart point. (Though if I leave the stage and come back, it's positioned near the exit point.)

If there are edges connected to it, then the edges remain 'connected' to an area where the node left the window.

![node-edge](https://user-images.githubusercontent.com/39674/39140166-c54dc460-46f1-11e8-880d-9cc3321010eb.png)

It's only a minor annoyance; any further click, drag, or navigation gets things rendering correctly again.

Expected behavior: either update the edge to follow the re-positioned node, or reposition the node at the window exit point.

(Note: an un-positioned node dropped offscreen is currently replaced on the upper-right stack of nodes, which does make sense, rather than moving to the window exit point.)

wwqrd:
This is an interesting one.

It seems like if you drag something offscreen, and then click the node that pinged back, it'll return to the exit point.

Might point to it being an artefact of rendering optimisations.

wwqrd:
Un-positioned nodes follow a different execution path in that they are "added when dropped", whereas dragged nodes are "updated as moved", so probably isn't crossover there.

rebeccamadsen:
Fixed in #689.

PR #689, opened 2018-09-27
Prevent node occlusion

### LINKED ISSUE #371: Node occlusion on the sociogram screen

rebeccamadsen:
Resolves #371 and #469. This adds the notion of a `DropObstacle`. Currently, the "reset" button, prompt, timeline, and node bucket on Sociogram are obstacles, meaning nothing can be dropped on those components. This also makes Sociogram full screen again. The off-screen node drops in #469 were resolved by adding a render on dragEnd.

Feedback welcome as always.

### DIFF
--- a/src/behaviours/DragAndDrop/DragPreview.js
+++ b/src/behaviours/DragAndDrop/DragPreview.js
@@ -22,6 +22,7 @@ export default class DraggablePreview {
     this.node.setAttribute('class', 'draggable-preview');
     this.x = -1000;
     this.y = -1000;
+    this.validMove = true;
 
     this.update();
 
@@ -55,11 +56,21 @@ export default class DraggablePreview {
 
   render() {
     this.node.setAttribute('style', styles(this.x, this.y));
+
+    if (this.validMove) {
+      this.node.setAttribute('class', 'draggable-preview');
+    } else {
+      this.node.setAttribute('class', 'draggable-preview node--inactive');
+    }
   }
 
   position(coords) {
     this.x = coords.x - this.center().x;
     this.y = coords.y - this.center().y;
+  }
+
+  setValidMove(valid) {
+    this.validMove = valid;
   }
 
   cleanup() {
--- a/src/behaviours/DragAndDrop/DropObstacle.js
+++ b/src/behaviours/DragAndDrop/DropObstacle.js
@@ -14,7 +14,7 @@ const dropObstacle = WrappedComponent =>
     static propTypes = {
       id: PropTypes.string.isRequired,
       accepts: PropTypes.func,
-      watchProps: PropTypes.arr,
+      watchProps: PropTypes.array,
     }
 
     static defaultProps = {
@@ -74,10 +74,11 @@ const dropObstacle = WrappedComponent =>
     }
 
     render() {
+      const { watchProps, ...rest } = this.props;
       return (
         <WrappedComponent
           ref={(component) => { this.component = component; }}
-          {...this.props}
+          {...rest}
         />
       );
     }
--- a/src/behaviours/DragAndDrop/DropObstacle.js
+++ b/src/behaviours/DragAndDrop/DropObstacle.js
@@ -1,0 +1,84 @@
+/* eslint-disable react/no-find-dom-node */
+
+import React, { Component } from 'react';
+import { findDOMNode } from 'react-dom';
+import PropTypes from 'prop-types';
+import { isEqual, pick } from 'lodash';
+
+import getAbsoluteBoundingRect from '../../utils/getAbsoluteBoundingRect';
+import { actionCreators as actions } from './reducer';
+import store from './store';
+
+const watchProps = ['width', 'height', 'x', 'y'];
+
+const propsChangedExcludingNodes = (nextProps, props) =>
+  !isEqual(pick(nextProps, watchProps), pick(props, watchProps));
+
+const dropObstacle = WrappedComponent =>
+  class DropObstacle extends Component {
+    static propTypes = {
+      id: PropTypes.string.isRequired,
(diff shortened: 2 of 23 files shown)


src/behaviours/DragAndDrop/reducer.js
```diff
@@ -95,6 +104,15 @@ const triggerDrop = (state, source) => {
     },
   });
 
+  filter(hits.targets, { willAccept: true })
+    .forEach((target) => {
+      target.onDragEnd(hits.source);
+    });
+
+  if (some(hits.obstacles, { isOver: true, willAccept: true })) {
```
bryfox: Is `willAccept` needed here? It seems like if a node is over an obstacle, then this should return.

If that check is removed, then can we also remove the `withHandlers({ accepts: () => () => true  })` composed props for the obstacles?

src/containers/ConcentricCircles/PromptObstacle.js
```diff
@@ -0,0 +1,38 @@
...
+PromptObstacle.propTypes = {
+  className: PropTypes.string,
+};
+
+PromptObstacle.defaultProps = {
+  className: '',
+};
+
+export default compose(
+  withBounds,
```
bryfox: Why is `withBounds` needed on these components? I don't see what's relying on the state it provides, but it seems very similar to the update logic in DropObstacle and DropTarget.

src/behaviours/DragAndDrop/reducer.js
rebeccamadsen: Interesting question. I was definitely modeling after DropTargets here; I thought we could potentially have a case where a DropObstacle doesn't matter in a certain situation. But we could always add an acceptance check back in if that arbitrary situation ever arises. All the obstacles block all drop actions right now anyway.

src/behaviours/DragAndDrop/DropObstacle.js
```diff
@@ -0,0 +1,80 @@
...
+
+import React, { Component } from 'react';
+import { findDOMNode } from 'react-dom';
+import PropTypes from 'prop-types';
+
+import getAbsoluteBoundingRect from '../../utils/getAbsoluteBoundingRect';
+import { actionCreators as actions } from './reducer';
+import store from './store';
+
+const maxFramesPerSecond = 10;
```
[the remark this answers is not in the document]
rebeccamadsen: That would probably make sense; I can do that!

src/containers/ConcentricCircles/PromptObstacle.js
rebeccamadsen: At one point I thought I needed it, but I don't think we actually do. I'll take that out. Thanks.

rebeccamadsen:
We could show an invalid state when over an obstacle, or we could show a valid state only over a valid drop area. My issue trying to implement the former idea is that moving from an invalid state outside the app (like dragging outside the app) shows a valid state while outside the app (because: no obstacles). Might be a little tricky workaround (outside the app becomes an obstacle/special case? But just on sociogram?). The latter idea works nicely on sociogram, but means the invalid state shows on other interfaces as well. If the invalid state is not obtrusive (lower opacity preview node?) that could be okay...but I'm not sure that's what we're looking for.

wwqrd:
Hmm, I wonder if we should 'lock' dragging to inside the window either way?

Could the obstacle check be part of the Monitor HOCs, and display of valid/invalid states left as a choice within the wrapped components?

jthrilly:
> I could see that, though the Preview may need some animation if you drag to the edge, and it "locks", but you enter the window elsewhere so the node preview "jumps" to the new location. Though animation may make it seem "lagging" if the node isn't right under the mouse when you enter the window again. If electron and mobile are meant to be full screen, this is probably just a browser edge case, though.

I think this sounds a little overly complex. 

> [#660 (comment)](https://github.com/codaco/Network-Canvas/pull/660#issuecomment-420601384) lists two other "ideal" UX behaviors: marking a node as being in invalid state while over an obstacle, and making the cloned node transition back to the last valid drop point. I'll let @jthrilly comment on those.

I would like to see the node enter the disabled state when hovering over an obstacle, if possible. Seems simple enough.

The animation back to the start drag point should also be practical, but I would settle for just cancelling the drop and leaving the node where it was (which is what currently happens).

rebeccamadsen:
Updated to show inactive state over an obstacle.

src/behaviours/DragAndDrop/DragPreview.js
```diff
@@ -55,13 +56,23 @@ export default class DraggablePreview {
 
   render() {
     this.node.setAttribute('style', styles(this.x, this.y));
+
+    if (this.validMove) {
+      this.node.setAttribute('class', 'draggable-preview');
+    } else {
+      this.node.setAttribute('class', 'draggable-preview node--inactive');
```
wwqrd: Should this be: `draggable-preview--obstacle` instead of `node--inactive`?

I know that we currently (and will probably always) only add this HOC to nodes, but it isn't intrinsically node specific.

rebeccamadsen: Good point. I can make that change in another branch since this one was merged already.

PR #691, opened 2018-09-27
Add null check for apiInfo

bryfox:
MDNS may send back announcements containing only information that we consider to be invalid.

This doesn't mean discovery has failed; additional (or previous) announcements may contain the needed information. Returning null here allows discovery to proceed.


rebeccamadsen:
Not sure how to test for this if it's a rare thing, but the fix looks reasonable to me!

PR #692, opened 2018-10-02
Fix large asset import on Cordova

### LINKED ISSUE #684: Cannot import protocols with large assets (Cordova)
Related to #681, but I don't yet know if the fix will be the same.

I'm unable to import the ~24MB development protocol on my iPad. From initial testing, it looks like the video is saved, so the download and extraction step shouldn't be the problem. But the behavior is the same as #681 on my device; after a wait, the app will reload.

bryfox:
Fixes #684. (Android & iOS.)

Test scenario:
1. Create a protocol package from the [development protocol source](https://github.com/codaco/Network-Canvas/tree/master/public/protocols/development.netcanvas)
    - Rename the protocol by changing the "name" property in protocol.json
    - Zip up contents & rename to *.netcanvas
2. Import this file to Server
3. Open NC in Android/iOS (actual device)
4. Pair with Server if needed
5. Import the created protocol
6. Wait

Expected: protocol is eventually imported successfully & displayed. Also note that (assuming you renamed the protocol in step 1) new sessions created from the imported protocol start quickly as expected.

Known issues:
- It's slow [on large assets]
  - Could be worth experimenting with increasing the buffer size to reduce write calls, but this increases memory use, so there would have to be some happy medium we derive from actual device testing.
  - Could theoretically rewrite zip extraction natively
  - Could be a good place for a [progress bar](https://github.com/codaco/Network-Canvas-UI/issues/66), and/or textual feedback like "Extracting {fileName.ext}..."
  - ...keeping in mind that importing is likely a relatively uncommon action?
- Implementation is hard to test. I stuck with the pattern of other filesystem functions, but would refactoring to a new class make sense now or in the future?

### DIFF
--- a/src/utils/__tests__/filesystem.test.js
+++ b/src/utils/__tests__/filesystem.test.js
@@ -2,7 +2,7 @@
 
 import environments from '../environments';
 import { getEnvironment } from '../Environment';
-import { getNestedPaths } from '../filesystem';
+import { getNestedPaths, writeStream } from '../filesystem';
 
 describe('filesystem', () => {
   describe('Cordova', () => {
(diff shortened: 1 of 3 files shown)


wwqrd:
Works for me on an iOS device.

> I stuck with the pattern of other filesystem functions, but would refactoring to a new class make sense now or in the future?

What would this look like? If we can separate out the platform implementations at the same time we could move to native zip too?

bryfox:
I'm not sure exactly what a refactor would look like; I was thinking extracting to a class & breaking it up to make some of the functions more testable. On the other hand, this is really just glue between JSZip and FileWriter, so despite function length, it might be easier to read/understand as it is? Integration tests would probably be more useful than unit tests anyway, but I don't know if we'll have that for cordova any time soon.

If the slowness is a problem, moving to native zip could be worth testing. I'd probably start with buffer size adjustments and compare that to native before starting on a new plugin.

bryfox:
I'd like to treat performance improvements mentioned here as a separate issue (whether part of the next release or not); a 25MB import is slow, but it was faster before mostly because it was crashing halfway through.

jthrilly:
Relates to #881

Issue #693, opened 2018-10-03
Need appropriate icon(s) for protocol import button


bryfox:
The "add" button on the Setup Screen uses an icon intended to mean "add screen"; there should be an appropriate icon for "add protocol".

![add](https://user-images.githubusercontent.com/39674/46431162-0d891e80-c719-11e8-95ad-91bff47c39fb.png)

Alternatively, we could have icons meaning "pair with server" and "browse protocols" as mentioned in #661.

PR #694, opened 2018-10-03
Update state & UI for single paired server

### LINKED ISSUE #661: Support only one paired Server at a time
Based on discussions elsewhere, we've decided to support only one paired Server at a time. This is expected to be the common case, so we can improve the UX for that case.

From @jthrilly:

> The flow for pairing/fetching a new protocol should be redesigned to make the new limitation clear. We might even consider changing the start screen UI so that the “fetch” button changes between a “pair button” and a “browse button”, depending on if a client is paired.

Note that session data can still be exported to a different Server than the protocol was imported from; the protocol just needs to be installed on the Server. (This is already supported.)

Redux state shape should also be adjusted to reflect this.

bryfox:
Resolves #661.

This updates redux state and UI/UX to better support the single-paired server use case / requirement.

### UI updates

The pairing process in unchanged.

Once paired, when a user visits the protocol import screen, they will be presented with their paired server and (if online) the available protocols. An unpair button is added to the paired server card.

unpair

If a server is offline, a retry button is presented (we no longer display the error modal):

offline

The unpair button is linked to a confirmation dialog:
unpair-confirm

### Notes

- The state changes here mean that any previously-paired servers in redux will be ignored. There's now a single `pairedServer` in state.
- I opened a separate issue (#693) to cover the possible pair/browse button toggling mentioned in #661 
- Unpairing only removes the pairedServer from local state. See https://github.com/codaco/Server/issues/170 and https://github.com/codaco/Server/issues/171


rebeccamadsen:
This works well for me!

bryfox:
Great! It sounds like @jthrilly will soon have some additional UI changes (related to #693). I'm going to merge this now; work there can be branched from master (and one won't need to review these changes again).

PR #695, opened 2018-10-03
Convert lint errors to warnings on dev server

bryfox:
This changes eslint errors to warnings on the dev server, so that the webview logs style violations to the console instead of stopping the show with the error overlay. Errors are still treated as errors everywhere else (build log, CI, IDE & local linting).

@wwqrd added this to Architect & I copied it in Server. It makes development easier while keeping the strict lint settings.


rebeccamadsen:
This seems okay to me.

PR #698, opened 2018-10-04
Fix error checking for certain protocol format issues

### LINKED ISSUE #673: If an edge is deleted from variable registry, it's still possible to add it in sociogram
Strictly speaking this is an invalid protocol and should be prevented in Architect.

When an edge type is deleted from the registry after being utilised in interfaces (in architect), this protocol still appears to work in NC.

Is this the right behaviour in NC (robustness principle)? These edges can also be exported.

Registry:
delete edge 1

Network:
delete edge 2

Export:
screen shot 2018-09-12 at 13 30 25

This will become a related issue once merged https://github.com/codaco/Architect/pull/214 - key becomes id, and name is a property in the registry.

### LINKED ISSUE #674: Name Generator with invalid node subject should fail with descriptive error message
Like #673, this is an invalid protocol and should be prevented in Architect.

When an node type is deleted from the registry after being utilised in interfaces (in architect), the following error appears in Name Generator:

delete node 1

 Sociogram appears to work:

delete node 2

Is this the right behaviour (for either) in NC (robustness principle)?

### LINKED ISSUE #675: Protocol without any stages should fail at the loading stage
Like #673 and #674 this is an invalid protocol and should be prevented in Architect.

Is this the right behaviour in NC?

permissive reading - no stages

bryfox:
Fixes #673. Fixes #674. Fixes #675.

This adds two points for validation to address the issues above.

- Adds a very basic validation step when loading a protocol. If no stages are defined, or the variableRegistry key is missing, loading will fail and the error dialog is displayed.
  - This can be replaced by [schema validation](https://github.com/codaco/Network-Canvas/pull/644) in the future
- Adds `assert` to selectors. Failures here manifest as a message on the ErrorBoundary. This is currently applied in the following cases:
  - If an interface requiring a node (subject) can't find the node definition in the registry
  - If an interface [sociogram] requiring an edge can't find the edge definition in the registry

Two other null checks were added to match the 'lenient-input' approach elsewhere in favor of explicit checks.

### DIFF
--- a/src/containers/NodePanels.js
+++ b/src/containers/NodePanels.js
@@ -119,13 +119,17 @@ class NodePanels extends PureComponent {
 const getNodesForDataSource = ({ nodes, existingNodes, externalData, dataSource }) => (
   dataSource === 'existing' ?
     existingNodes :
-    differenceBy(externalData[dataSource].nodes, nodes, nodePrimaryKeyProperty)
+    differenceBy(
+      externalData[dataSource] && externalData[dataSource].nodes,
+      nodes,
+      nodePrimaryKeyProperty,
+    )
 );
 
 const getOriginNodeIds = ({ existingNodes, externalData, dataSource }) => (
   dataSource === 'existing' ?
     map(existingNodes, nodePrimaryKeyProperty) :
-    map(externalData[dataSource].nodes, nodePrimaryKeyProperty)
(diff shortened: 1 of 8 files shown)


jthrilly:
This looks good to me. Ready to merge?

PR #711, opened 2018-10-10
Fix/reset linking state

### LINKED ISSUE #690: Node "linking" prop not reset on sociogram
If I reset the Sociogram while a node is in the "linking" state, then it remains in that state. Later clicking another node will create a network edge, though the edge may not be visible.

Steps:
1. Add nodes to network
2. Go to a sociogram which allows edge creation
3. Drag/drop 3 nodes from bucket
4. Click node 3
5. Click reset
6. Drag/drop 1 node from bucket
7. Click the dropped node

Observed: an edge is created in the network (between 1 & 3), but there is no indication to the user this happened.

rebeccamadsen:
Fixes #690. This resets the state for the sociogram's `connectFrom` link when "reset" button is clicked, so unintentional edges aren't created after a reset. Verification steps are in the issue.

### DIFF
--- a/src/containers/ConcentricCircles/ConcentricCircles.js
+++ b/src/containers/ConcentricCircles/ConcentricCircles.js
@@ -5,7 +5,7 @@ import NodeLayout from './NodeLayout';
 import EdgeLayout from './EdgeLayout';
 import Background from './Background';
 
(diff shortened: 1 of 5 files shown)


rebeccamadsen:
I could see that! We could do that here if there's time, or create an issue if that's preferred.

PR #745, opened 2018-10-15
Prevent app from accepting dropped files/snippets

### LINKED ISSUE #744: App should not accept drop events
If I drop a file or certain clipboard data onto the app, it's currently handled with default browser actions (for example, displaying an image or downloading a file).

bryfox:
Fixes #744.


jthrilly:
Looks good 👍

PR #746, opened 2018-10-15
Fix/timeline hitbox

### LINKED ISSUE #733: Large hit box for forwards button in timeline navigation triggers animation, but not forwards action.
Clicking in the highlighted region triggers button animation, but actually opens the menu (which is the correct action for clicking the timeline itself).

screen shot 2018-10-15 at 15 41 49

Similar issue with the back button.

jthrilly:
- Resolves an issue where the timeline buttons were clipping without reintroducing the overscroll issue.
- Resolves #733. An issue where button clicks were triggering the menu open when they shouldn't.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ branches:
   only:
     - master
 node_js:
-  - "10.11"
+  - "8.11"
 cache:
   directories:
     - node_modules # NPM packages
--- a/src/components/ProgressBar.js
+++ b/src/components/ProgressBar.js
@@ -1,19 +1,21 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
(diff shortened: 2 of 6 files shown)


bryfox:
This still fires the animation as though the button was hit, but the actual navigation doesn't happen.

hit

Actually, the same thing happens on the top arrow button. Is this something we should solve now?

jthrilly:
I don't think so. It isn't going to be possible to solve with the current approach to doing the button animation. I think the main thing is that we don't show the button animation and trigger the menu. You have to fairly obviously click outside the button in order to trigger the animation without the button click, and I think a user will understand what has happened.

jthrilly:
That said, I think an SVG animation for these buttons would be great, down the line. Unfortunately, this hack will have to do for the moment.

PR #750, opened 2018-10-16
Feature/ui modals

wwqrd:
Use UI Modal/Dialog components in Network Canvas

- Updates NodeForm to use Modal and Overview Component
- Replaces remaining Dialog components with the DialogManager feature merged in here https://github.com/codaco/Network-Canvas/pull/712
- Removes old Dialog/Modal components and styles


bryfox:
There seem to be some z-index issues; could be a forcing function for https://github.com/codaco/Network-Canvas-UI/issues/56?

- Both the modal contents and backdrop of the node forms now appear below the Timeline (z-index-wise), which I wouldn't expect.
    - this means I can change prompts while the form is visible
- The app reset confirmation appears below the stages menu (and protocol list, so it's unclickable there)
    - This was an extreme case where I "couldn't" reset, but I wonder if modal state shouldn't be persisted [between view reloads & app launches]. Modals are typically more transient than this?

I'm also not seeing error messages at all (not sure if this is a layering issue or something else). For example: when I start pairing with server but Dismiss from the Server dialog, or try to import a protocol from an invalid URL.

wwqrd:
Definitely agree with you that the notices shouldn't persist, I thought we'd resolved that already in #712.

bryfox:
With the latest changes, looks good to me 👍

package.json
```diff
@@ -34,7 +34,8 @@
     "dist:linux": "npm run build:electron && electron-builder -l --publish=always",
     "dist:mac": "npm run build:electron && electron-builder -m --publish=always",
     "dist:win": "npm run build:electron && electron-builder -w --publish=always",
-    "bump": "sh scripts/bump"
+    "bump": "sh scripts/bump",
+    "preflight": "npm run lint && npm run sass-lint -- --max-warnings 0 && npm test -- --coverage"
```
bryfox: One thing I'm changing on Server is to add `--max-warnings 0` to [es]lint as well.

PR #752, opened 2018-10-16
Check protocol paths before unzipping

bryfox:
This checks the zipped file paths in a protocol before writing output to those paths.

To support [flexibility in the directory structure](https://github.com/codaco/Network-Canvas/wiki/Protocol-Overview), this detects problems. If we standardize on a top-level structure (perhaps moving worker scripts into `workers/`, then we could more easily check against a whitelist here, which is probably preferable.

### DIFF
--- a/src/utils/protocol/__tests__/importProtocol.test.js
+++ b/src/utils/protocol/__tests__/importProtocol.test.js
@@ -14,9 +14,9 @@ const absolutePath = ['/example'];
 const emptyPath = [''];
 const mixedPaths = ['protocol.json', 'assets/example', '../example'];
 
-const traversalError = { message: /directory traversal not allowed/ };
-const absPathError = { message: /absolute paths not allowed/ };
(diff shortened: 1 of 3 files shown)


wwqrd:
I wonder if splitting the platform implementations at the rule level would make it easier to ensure that same rules are followed for each platform? It's a small bit of code, and fine as it is though.

bryfox:
@wwqrd thanks... I extracted the individual rule validations to a new 'protocolValidation' file and standardized how they're applied on each platform.

wwqrd:
Looks good to me code wise.

I tested traversal on os x and got the expected error, attempted absolute paths, but got a protocol.json not found error (perhaps I didn't construct the archive properly), not sure how to test for empty paths.

On cordova (android) I got the expected traversal error. Absolute paths, uncaught FileError, but again assume this has more to do with my dummy archive.

PR #753, opened 2018-10-16
Add schema scaffolding

### LINKED ISSUE #697: Create "generic" JSON schema
See description of initial prototype in #644.

1. We should have a protocol that captures general requirements so that we can auto-generate a basic schema that could be applied to all protocol.json files. This won't capture all requirements, but should serve as a starting point for identifying missing fields, etc.

2. Determine what requirements cannot be captured by auto-generation, but that can be expressed in JSONSchema.

The outcomes of this issue will be:
- to have a basic schema from which we can validate protocols (for QA or at runtime)
- to have an understanding of how much of the protocol requirements a schema can capture, and how we'll best use it in the future

bryfox:
This adds all of the initial scaffolding required for generating and validating a protocol JSON schema. See #697 for discussion.

Dev dependency changes:
- ajv@^6.5.4 (another version was indirectly included before). Required for validation.
- quicktype@^15.0.137. Used to auto-generate an initial schema from an example protocol.
    - Not strictly required; see #697.

Other changes:
- Added scripts to handle the above tasks
- Updated factory protocols to provide better example data

The example protocol and schema are generated in `{projDir}/schema`, and currently ignored by git.

Next steps:
1. Review the generated schema
2. Update to address any immediate concerns
3. Consider checking schema into VC (or moving this to a separate repo)

Resolves #697.


jthrilly:
Looks great, Bryan!

I was able to successfully generate and validate the schema without issue.

There are a couple of small issues in the generated schema, but they are perhaps best solved through traditional scripting methods (ensure X is a member of set Y, etc.). I think we discussed perhaps looking at if we can use a testing framework for those? It was an idea I had anyway. Might be impractical, but on the surface a testing framework would allow us to wrap up the remaining schema logic in small assertions/rules.

While I think of it, one other idea I had was to explore how we might use the new GitHub automation features to add a protocol validation step to our PR checks. This would ensure that changes to the factory protocols are valid. Just another idea to throw around when we meet on Friday.

Thanks for the great work on this!

Issue #727, opened 2018-10-12
protocol loading screen has spurious renders


rebeccamadsen:
This may just be solvable with a "close" button on the loading screen, but sometimes the loading screen gets stuck. I've seen it in electron when doing development and saving a change that makes the build error and then saving a change that doesn't error. But I've also seen this "stuck" state on electron and Android emulator when attempting to load a protocol with errors. Trying to load a protocol from the paired server stalls after Download and Import steps (Android):
```
Uncaught (in promise) FileError {code: 1}code: 1__proto__: Object
```

In Electron I get a dialog, but still get stuck:
```
index.js:2178 Error: ENOENT: no such file or directory, open 'C:\Users\Big Baloo\AppData\Roaming\Network Canvas\protocols\257e210a-e461-42d5-ba59-9c5136dee238\assets\click_the_thing.mp3'
__stack_frame_overlay_proxy_console__ @ index.js:2178
reducer @ errors.js:22
combination @ combineReducers.js:120
rootReducer @ rootReducer.js:40
(anonymous) @ autoRehydrate.js:30
dispatch @ createStore.js:165
(anonymous) @ index.js:104
(anonymous) @ createEpicMiddleware.js:59
(anonymous) @ index.js:11
(anonymous) @ middleware.js:13
dispatch @ applyMiddleware.js:35
./node_modules/rxjs/Subscriber.js.SafeSubscriber.__tryOrUnsub @ Subscriber.js:242
./node_modules/rxjs/Subscriber.js.SafeSubscriber.next @ Subscriber.js:189
./node_modules/rxjs/Subscriber.js.Subscriber._next @ Subscriber.js:129
./node_modules/rxjs/Subscriber.js.Subscriber.next @ Subscriber.js:93
./node_modules/rxjs/operators/switchMap.js.SwitchMapSubscriber.notifyNext @ switchMap.js:126
./node_modules/rxjs/InnerSubscriber.js.InnerSubscriber._next @ InnerSubscriber.js:23
./node_modules/rxjs/Subscriber.js.Subscriber.next @ Subscriber.js:93
./node_modules/rxjs/operators/mergeMap.js.MergeMapSubscriber.notifyNext @ mergeMap.js:145
./node_modules/rxjs/InnerSubscriber.js.InnerSubscriber._next @ InnerSubscriber.js:23
./node_modules/rxjs/Subscriber.js.Subscriber.next @ Subscriber.js:93
[...]

PR #754, opened 2018-10-16
Fix load screen for slow protocol loading

bryfox:
Interim fix for #727.

`getDerivedStateFromProps` is called before every render (as of React 16.4), which means `setState()` triggers it, and if the protocol slowly enough then `minimumTimeMet` settles at `false`.

The change here boils down to removing the `prevProps.isWorking !== this.props.isWorking` check. This could be improved (as it may cause additional rendering), but should fix the immediate issue for now.

The easiest way to test is to download a protocol with assets on Android.


rebeccamadsen:
This seems to solve the problems I've come across, especially with protocols with assets or errors!

PR #755, opened 2018-10-16
Prevent navigation away from app

bryfox:
Same implementation as https://github.com/codaco/Server/pull/185:

While there is no intended mechanism to navigate away from the app URL within the app, this prevents any other (unexpected/untrusted) URLs from loading in the main window.


rebeccamadsen:
Works for me!

PR #756, opened 2018-10-24
Implement persistent app settings

### LINKED ISSUE #730: Software keyboard causing pairing dialog to be pushed to top
Using the software keyboard to enter the pairing code causes the UI to be pushed upwards, with no way to get it back down.

img_0003

jthrilly:
This PR implements persistent app settings and (hopefully) sane defaults. 

- Adds interface scaling option, allowing the user to set the scale factor of the base font size.
- Adds the ability for the user to set and edit the device name.
- Adds the ability for the user to toggle "dynamic scaling" on or off. This will switch the app between viewport units and px for the `--base-font size`.
- Adds the ability for the user to turn the full screen node form on or off, with a default value determined by device characteristics.

Fixes #730 by setting android devices to not use dynamic scaling by default.


src/components/MainMenu/SettingsMenu.js
```diff
@@ -1,56 +1,165 @@
...
+class SettingsMenu extends PureComponent {
+  componentDidMount() {
+    this.setFontSize();
+  }
+
+  componentDidUpdate() {
+    this.setFontSize();
+  }
+
+  setFontSize = () => {
```
wwqrd: I think that this and the component methods probably belong in `<App />`. At the moment SettingsMenu persists on the screen, but that might not always be the case.

PR #757, opened 2018-10-25
Fix downloaded asset URLs

### LINKED ISSUE #681: Use of large assets in protocols exhausts memory (Cordova)
Some iOS devices are having issues with the Video interface screen in the development protocol.

The browser page reloads during (or before) playback, or the app may crash. 

- Instrumented Memory use seems stable and not unusually high
- Safari can play video without issue on a standalone test page

bryfox:
Fixes #681 (large assets included in `download` protocols; the counterpart to #696).

This changes the asset URLs for all assets included in a downloaded protocol on Cordova:
- On Android, sets the asset URL directly from the saved protocol path
- On iOS, copies the asset to the device's tmp directory (for reasons discussed in #681)

To test: download (from Server or a URL) a protocol that contains large assets; for example, an external copy of the development protocol.

Limitations:
- As discussed in #681, this is intended as a temporary workaround until Cordova (or we) support `WKURLSchemeHandler` changes in the future
- Will not run in our [iOS & Android] development mode (i.e., the page must be loaded from `file://`)

To make protocols behave better in development, we could (for example) choose to only apply this to Video and Audio assets. That would mean images would display again in development, but that large images would still cause issues in production (at best, they load slowly).

### DIFF
--- a/src/utils/filesystem.js
+++ b/src/utils/filesystem.js
@@ -410,6 +410,8 @@ const writeStream = inEnvironment((environment) => {
   throw new Error(`writeStream() not available on platform ${environment}`);
 });
 
+// FIXME: this implies that it will recursively create directories, but if targetPath's parent
+//        doesn't already exist, this will error on both platforms
 const ensurePathExists = inEnvironment((environment) => {
   if (environment === environments.ELECTRON) {
     const path = require('path');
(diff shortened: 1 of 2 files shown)


jthrilly:
I think the dev mode limitation is totally a worthwhile tradeoff for now. I've tested this, and found it to be working in all environments.

Issue #758, opened 2018-10-26
Refine protocol requirements


bryfox:
This issue will be updated to capture small clarifications, updates, and additions to protocol validation which can be captured either by Schema changes or scripted verification. This is an extension of #697.

### Enumerable values
- [x] `stage type`
- [x] `SkipLogic.action`
- [x] `SkipLogic.operator`
- [x] `filter.join`
- [x] `filter.rules.options.operator`
- [x] `filter.rules.type`
- [x] `forms.fields.component`
- [x] `validation` keys
- [x] enum for Information items.type (text, video, image, audio)
- <strike>enum for iconVariant?</strike>
- [x] enum for subject.type: only node or edge (note: this is effectively ignored by NC for now.)
- [x] enum for rule.type: alter, ego, edge

### Numerical

- [x] `SkipLogic.value` must be a positive integer

### Other

- [x] `items` is required if Stage type is Information
- [x] some fields are required for other Stage types? (`prompts`?)
- [x] `SkipLogic.value` is required unless `SkipLogic.operator` is `ANY` or `NONE`
- [x] `filter.rules.options.value` is required unless `options.operator` is `EXISTS` or `NOT_EXISTS`
- [x] `lastModified` should have date format enforced
- [x] `filter.rules[].options[].type` is required
- [x] `variables` should be optional for variableRegistry edges

### Variable registry

- [x] form field `component`s must correspond to an appropriate type of `variable` (e.g., a `Checkbox` variable should be defined as boolean in the registry). See [wiki](https://github.com/codaco/Network-Canvas/wiki/Variable-Types)
- [x] `filter.rules[].options[].attribute` should be defined for the associated `filter.rules[].options[].type`
- [ ] `filter.rules[].options[].value` should be of type `filter.rules[].options[].type` (https://github.com/codaco/Architect/issues/251)
    - see comment/question below

The following must refer to valid entries in the variable registry.

- [x] `node.displayVariable` (e.g., nickname)
- [x] `filter.rules[].options[].type`
- [x] `subject.type` must match one of `variableRegistry[subject.entity]`
- [x] `forms[key].entity` + `forms[key].type`
- [x] `forms.fields[].variable`
- [x] `stages[].prompts[].cardOptions.additionalProperties[]`
- [x] `stages[].prompts[].sortOptions.sortOrder[].property`
- [x] `stages[].prompts[].sortOptions.sortableProperties[]`
- [x] `prompts[].additionalAttributes`
[...]

PR #765, opened 2018-10-29
Fix & capture latest requirements in schema

bryfox:
This updates the schema format based on the items identified in #758. The remaining "dynamic data" items listed will be handled outside the schema. This also includes fixes for some of the dynamic object definitions (forms & variables), which were previously not catching some format failures.

I believe this to be pretty close to what we want to capture with the schema today, so I've also added the schema itself to the repo so we can start tracking changes.

Summary of changes to development protocol:
- Fix "coordinates" type ([should be](https://github.com/codaco/Network-Canvas/wiki/Variable-Types) "location")
- Add new "name" prop to variables (https://github.com/codaco/Architect/pull/260)
- Fix `Text` component name (https://github.com/codaco/Architect/pull/300)
- Fix `skipLogic.value`: should be number
- Remove old `version` property


bryfox:
No, thanks for reviewing & merging... I'm working on the remaining (data) validations in a separate branch.

PR #766, opened 2018-10-30
Use new variables for sociogram radar colors

wwqrd:
Outside of Network Canvas app, the global --background value may differ from the color we want for the outer edge of ther radar.

This separates it out into a new variable, which defaults to mirror --background.


wwqrd:
Accidentally bundle a change for the swiper path here, does anyone see a reason to omit it?

PR #767, opened 2018-10-30
Allow users to enter IPv6 addresses

bryfox:
Also support leading zero-compressed addresses.

### DIFF
--- a/src/styles/components/_server-card.scss
+++ b/src/styles/components/_server-card.scss
@@ -1,11 +1,11 @@
 .server-card {
-  height: 11rem; // icon + text
   margin: 2rem;
   width: 10rem;
 
   &__label {
     text-align: center;
     margin: 0;
+    overflow-wrap: break-word;
   }
 
   &--clickable {
--- a/src/styles/components/_server-setup.scss
+++ b/src/styles/components/_server-setup.scss
@@ -7,7 +7,7 @@
     align-items: center;
     display: flex;
     flex-direction: column;
-    justify-content: space-between;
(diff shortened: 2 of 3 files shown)


rebeccamadsen:
Ditto with the server PR -- this looks okay to me, and works for my setup, but I'm not on an IPv6 network, I don't think.

jthrilly:
Just a few CSS issues here with the longer IP addresses:

screenshot 2018-11-09 at 13 41 12
screenshot 2018-11-09 at 13 41 24
screenshot 2018-11-09 at 13 44 26

I will fix these up now.

One other issue is that I couldn't get "Fetch from URL" to work with an IP6 address. I was using a gist I found for simplehttpserver (https://gist.github.com/chrisklaiber/54511886e8e4c18126792fc634f44d57#file-simplehttpserver6-py), which might have been the issue. If someone else can verify it works for them, I am happy to presume its my issue.

bryfox:
@jthrilly what URL did you enter? Are you bracketing the host (e.g., `http://[::1]:8000/foo.netcanvas`, and not `http://::1:8000/foo.netcanvas`)?

jthrilly:
> @jthrilly what URL did you enter? Are you bracketing the host (e.g., `http://[::1]:8000/foo.netcanvas`, and not `http://::1:8000/foo.netcanvas`)?

That was it, Bryan. Thanks. I always forget about that syntax.

bryfox:
Yeah, maybe it's worth showing an example or help text on that screen?

PR #769, opened 2018-10-31
Fix/bundle bonjour

jthrilly:
This PR bundles the x64 Bonjour installer with our app installer, and automatically triggers its installation during the Network Canvas install.

If Bonjour is already installed, the additional installer does nothing.

If Bonjour is not installed, the installer prompts the user if they want to install it. The installation is otherwise uninterrupted.

To test:
- On Windows: `npm run dist:win` - code signing is disabled for now, so you can do this from a VM even though you don't have the certificates.

Remaining issues:

- Should the user be prompted about the installation of Bonjour to explain why it is needed?
- Are there any legal issues with hosting the Bonjour msi package in our GitHub repo? It was extracted from the Bonjour SDK executable, which you can only access as an Apple developer.


bryfox:
Works for me... I tested Windows 10 and 7, installing with & without Bonjour already installed. I also confirmed it doesn't break discovery in an existing iTunes install (which doesn't itself install that standalone Bonjour exe).

bryfox:
Side note: to get the dist task to work on Windows, I had to downgrade my npm version. With 6.4.1 & 6.1, I get the [error described here](https://github.com/npm/npm/issues/20815). With npm 5.7.0, I notice there are minor & patch differences in the generated package lock. At a glance, I don't see anything that would necessarily affect the build, but curious if you had the same issue.

jthrilly:
> Side note: to get the dist task to work on Windows, I had to downgrade my npm version. With 6.4.1 & 6.1, I get the [error described here](https://github.com/npm/npm/issues/20815). With npm 5.7.0, I notice there are minor & patch differences in the generated package lock. At a glance, I don't see anything that would necessarily affect the build, but curious if you had the same issue.

I did, actually, which was one of the reasons that the inconsistent npm and node versions crept into the release before last. I can't say why, but I'm not having that issue any longer.

jthrilly:
> The first thing I saw after clicking the installer was the Bonjour installer confirmation ("developed by Apple"), so that would be nice. I'm not sure what the distribution plan is for most users, but an explanation before downloading might suffice.

Yeah, agreed. I'll add a confirm box for this.



> This Apple [licensing page](https://developer.apple.com//softwarelicensing/agreements/bonjour.php#bonjourwin) might have the information you need.

I did check that out, but it doesn't seem to cover hosting the redistributable on github, which would seem to be against their other terms. Redistributing with our app once we have signed the agreement is fine, of course. I'm probably going to leave it up, and will remove it if anyone asks.

PR #773, opened 2018-11-05
Fix worker node & edge props

### LINKED ISSUE #719: Mock data format is out of date
Follow-on to #648. Generated mock nodes need to use the new format (attributes keyed by ID). IDs would depend on the protocol being used.

### LINKED ISSUE #771: Fix node data passed to label workers
- We currently [pass a node's `type`](https://github.com/codaco/Network-Canvas/blob/master/src/ducks/modules/network.js#L49) as a special property to label workers. A node's type is now an opaque ID, meaningless to user scripts.
- Node attributes are stored as IDs; these also need to be transposed to names for the user script

This needs to be translated to a meaningful variable name, i.e. `variableRegistry.node[uuid].name`.

bryfox:
This fixes the node information sent to worker label scripts; for example, from:

```json
{
    "0e75ec18-2cb1-4606-9f18-034d28b07c19": "Dee",
    "networkCanvasType": "4aebf73e-95e3-4fd1-95e7-237dcc4a4466"
}
```

to:
```json
{
    "name": "Dee",
    "networkCanvasType": "person",
}
```

Also fixes both node & edge information in the network passed to workers.

To test, you can replace the development protocol with [one that uses ID keys](https://raw.githubusercontent.com/codaco/Network-Canvas/chore/dynamic-validation/public/protocols/development.netcanvas/protocol.json).

Fixes #771. Fixes #719.

### DIFF
--- a/src/containers/MainMenu/SettingsMenu.js
+++ b/src/containers/MainMenu/SettingsMenu.js
@@ -8,7 +8,7 @@ import { actionCreators as uiActions } from '../../ducks/modules/ui';
 import { actionCreators as mockActions } from '../../ducks/modules/mock';
 import { actionCreators as dialogsActions } from '../../ducks/modules/dialogs';
 import { actionCreators as deviceSettingsActions } from '../../ducks/modules/deviceSettings';
-import { getNodeEntryForCurrentPrompt } from '../../selectors/session';
+import { getAdditionalAttributesForCurrentPrompt, getNodeEntryForCurrentPrompt } from '../../selectors/session';
 
 const settingsMenuHandlers = withHandlers({
   handleResetAppData: props => () => {
@@ -28,7 +28,7 @@ const settingsMenuHandlers = withHandlers({
       return;
     }
     const [typeKey, nodeDefinition] = props.nodeVariableEntry;
-    props.generateNodes(nodeDefinition.variables, typeKey, 20);
+    props.generateNodes(nodeDefinition.variables, typeKey, 20, props.additionalMockAttributes);
     props.closeMenu();
   },
 });
@@ -48,6 +48,7 @@ const mapStateToProps = state => ({
   protocol: state.protocol,
   nodeVariableEntry: getNodeEntryForCurrentPrompt(state),
   shouldShowMocksItem: !!getNodeEntryForCurrentPrompt(state),
+  additionalMockAttributes: getAdditionalAttributesForCurrentPrompt(state),
   useFullScreenForms: state.deviceSettings.useFullScreenForms,
   useDynamicScaling: state.deviceSettings.useDynamicScaling,
   deviceDescription: state.deviceSettings.description,
--- a/src/containers/MainMenu/SettingsMenu.js
+++ b/src/containers/MainMenu/SettingsMenu.js
@@ -8,6 +8,15 @@ import { actionCreators as mockActions } from '../../ducks/modules/mock';
 import { actionCreators as dialogsActions } from '../../ducks/modules/dialogs';
 import { actionCreators as deviceSettingsActions } from '../../ducks/modules/deviceSettings';
 
+const personEntry = (protocol) => {
+  const registry = protocol && protocol.variableRegistry;
+  const entry = registry && registry.node && Object.entries(registry.node).find(([, nodeType]) => nodeType.name === 'person');
+  if (!entry || entry.length !== 2) {
+    return null;
+  }
+  return entry;
+};
+
 const settingsMenuHandlers = withHandlers({
   handleResetAppData: props => () => {
     props.openDialog({
@@ -22,7 +31,12 @@ const settingsMenuHandlers = withHandlers({
     });
   },
   handleAddMockNodes: props => () => {
-    props.generateNodes(20);
+    const entry = personEntry(props.protocol);
+    if (!entry) {
+      return;
+    }
+    const [typeKey, personDef] = entry;
+    props.generateNodes(personDef.variables, typeKey, 20);
     props.closeMenu();
   },
 });
@@ -39,6 +53,8 @@ const mapDispatchToProps = dispatch => ({
 });
 
 const mapStateToProps = state => ({
+  protocol: state.protocol,
+  shouldShowMocksItem: !!personEntry(state.protocol),
   useFullScreenForms: state.deviceSettings.useFullScreenForms,
   useDynamicScaling: state.deviceSettings.useDynamicScaling,
   deviceDescription: state.deviceSettings.description,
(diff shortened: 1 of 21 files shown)


bryfox:
I've also pushed the mock data updates, since (a) they're related and (b) I'm trying to confirm everything's working once the development protocol's keys don't happen to match names.

bryfox:
Note on mock data: it's only supported (and the button only shown) when the protocol defines a 'person' type, since it's doing variable lookups in the registry to transpose names & IDs.

The teaching protocol should therefore support mock data, but it doesn't here because it hasn't been updated to match the new protocol format (with name props). That problem is caught by newer validations, and those mocks are fixed, in `feature/runtime-validation` (#774).

src/containers/MainMenu/SettingsMenu.js
```diff
@@ -8,6 +8,15 @@ import { actionCreators as mockActions } from '../../ducks/modules/mock';
 import { actionCreators as dialogsActions } from '../../ducks/modules/dialogs';
 import { actionCreators as deviceSettingsActions } from '../../ducks/modules/deviceSettings';
 
+const personEntry = (protocol) => {
```
wwqrd: I think this should live elsewhere - it seems useful generally?

src/ducks/modules/mock.js
```diff
@@ -7,23 +7,27 @@ import { nodeAttributesProperty } from './network';
 
 const MOCK_GENERATE_NODES = 'MOCK/GENERATE_NODES';
 
-const generateNodes = (howMany = 0) =>
+const keyForVarName = (vars, name) => {
+  const entry = Object.entries(vars).find(([, val]) => val.name === name);
+  return entry && entry[0];
+};
+
+const generateNodes = (variableDefs, typeKey, howMany = 0) =>
```
wwqrd: I wonder if we need to use "real" names or we should move to populating with random words/numbers based on variable types in the registry? That way we could populate any node types?

src/ducks/modules/network.js
```diff
@@ -39,15 +39,48 @@ function edgeExists(edges, edge) {
...
-// Returns node data safe to supply to user-defined workers.
-// Contains all user attributes flattened with the node's unique ID.
-// `primaryKeyPropertyForWorker` is used to minimize conflicts, but user data is always preserved.
-export const asWorkerAgentNode = node => ({
+/**
+ * Internally, 'attributes' are stored with UUID keys, which are meaningless to the end user.
+ * This resolves those UUIDs to variable names based on the definitions in the variable registry,
+ * appropriate for user scripts and export.
+ */
+const getNodeAttribtuesWithNamesResolved = (node, nodeVariableDefs) => {
```
wwqrd: getNodeAttrib~~tu~~esWithNamesResolved

src/ducks/modules/network.js
```diff
@@ -39,15 +39,48 @@ function edgeExists(edges, edge) {
...
+  if (!nodeVariableDefs) {
+    return {};
+  }
+  const attrs = getNodeAttributes(node);
+  return Object.keys(attrs).reduce((acc, uuid) => {
```
wwqrd: 👌

src/ducks/modules/mock.js
```diff
@@ -7,23 +7,27 @@ import { nodeAttributesProperty } from './network';
...
       return dispatch(sessionsActions.addNodes({
-        type: 'person',
+        type: typeKey,
         promptId: 'mock',
         stageId: 'mock',
         [nodeAttributesProperty]: {
-          name: `${firstName} ${lastName}`,
-          nickname: lastName,
-          age,
+          [keyForVarName(variableDefs, 'name')]: `${firstName} ${lastName}`,
```
wwqrd: I think this should have a similar interface to `getNodeAttribtuesWithNamesResolved` (sp.), just in reverse?

```
return dispatch(sessionsActions.addNodes({
  type: typeKey,
  promptId: 'mock',
  stageId: 'mock'
  [nodeAttributesProperty]: getAttributesWithIds(
    variableDefinitions,
    {
      name: `${firstName} ${lastName}`,
      nickname: lastName,
      age,
    }
  ),
})
```

Or maybe it could be at the node level?:

```
  getNodeWithIdAttributes(variableDefinitions, nodeProperties, nodeAttributes);
```

src/ducks/modules/mock.js
bryfox: I'm fine with either approach. Since Architect includes a 'person' type by default and this is just a dev feature, I don't mind the hardcoded 'person'. But if you'd prefer the latter, I can update here or in a follow-up PR.

src/containers/MainMenu/SettingsMenu.js
bryfox: Do you think this 'entry' (`[key, val]`) return type would be expected (i.e., good for a utility function)? I don't love it if we're going to reuse the func, but I also don't want to introduce another arbitrary structure.

bryfox:
> Maybe this should be added to the built in protocols?

Coming soon... #774 includes that for the development protocol. I'll make a note to update the teaching protocol as well.

bryfox:
@wwqrd thanks for the feedback... Other than supporting any node type in mock data (see prior comments/question), I think I've addressed all items.

src/ducks/modules/mock.js
jthrilly: Maybe having the dev feature generate nodes of various types would help us catch edge cases. For example hard coded text strings that don't make sense, or other faulty logic.

src/containers/MainMenu/SettingsMenu.js
bryfox: This is now exposed as `getNodeEntryForCurrentPrompt` in the session selector. Return type is Array as before.

src/ducks/modules/mock.js
bryfox: This is now `getNodeWithIdAttributes` in the network module. With the latest changes to mock generation, it's never called. However, I'm leaving it in since it'll be useful when importing external data down the line.

src/ducks/modules/mock.js
bryfox: This is now supported; implemented as @wwqrd described above. Mocks will be generated based on the current prompt's subject and apply any `additionalAttributes` defined on it.

Side note: `additionalAttributes` was being used to name the additional props for `addNodes`, which themselves contained the `additionalAttributes` defined in the protocol; I renamed the former to `additionalProperties` to avoid confusion. (This could affect the add/update actions, but I believe I caught all the usages.)

wwqrd:
👍

PR #774, opened 2018-11-06
Dynamic protocol validation

### LINKED ISSUE #758: Refine protocol requirements

bryfox:
Depends on #773.

This adds dynamic/business logic validations for protocols to resolve #758. Both schema & logic validations are now run when a [downloaded] protocol is loaded. For now, any errors are logged as warnings, but the session begins as normal.

Other supporting changes:
- Updates the development protocol to use UUID keys (matching expected Architect output)
- Moves the schema inside `src/` so it can be imported by the application
- Organizes validation code under ` src/utils/protocol/validation/`
    - See `validateLogic.js` for the dynamic validation definitions

### Known issues

Validation code is currently shared by the runtime app and the npm script, and so is written using module syntax. We could run the validation script through babel, which would mean adding a babel-cli dev dependency and getting the configuration to work between both.


bryfox:
I rebased & pushed some updates which correctly set the required props for edge definitions in the variable registry. (Edges require name and label, but variables are optional.)

Issue #722, opened 2018-10-12
NG Autocomplete styling refactor?


bryfox:
- [x] Border-radius definition is invalid in [search](https://github.com/codaco/Network-Canvas/blob/master/src/styles/components/_search.scss#L9)
- [x] Number offset in node icon
- [x] `--collapsed-square-size` was removed; need to restore, but actual value tbd since node size is now tied to vmin; may need to replace with dynamic calculation if approximation isn't good enough

![ngac](https://user-images.githubusercontent.com/39674/46883570-74e15580-ce20-11e8-8c76-d9bbeb4107b0.png)

Edit: See comments; we may want to revise the design.

bryfox:
Close button styling has also changed; need to reinstate something like the original cross styles (see https://github.com/codaco/Network-Canvas/pull/712#discussion_r224878776)

jthrilly:
I think we should perhaps take the opportunity to revise the UI for this interface, and perhaps create dedicated components for some of these UI elements.

Another possible enhancement would be to reimplement the animation using the FLIP approach that Steve has explored in Architect.

bryfox:
See also #735, which highlights an issue with the current behavior, and which might be solvable with the revised design.

PR #778, opened 2018-11-08
Fix autocomplete styling & animations

bryfox:
This fixes the styling and animation problems described in #722, though I'll leave that open since discussion has started around a redesign there.

This also adjusts the width of the Search component to match the centering of other prompt content.


jthrilly:
Looks great!

Issue #547, opened 2018-06-07
Ordinal Bin 2.0


jthrilly:
- [ ] Animate when the number of bins changes. Ensure nodes animate out.
- [x] Prevent interaction with anything but the first node in the bucket
- [ ] Merge the bucket component with the nodeList.
- [x] Implement "interaction based sorting" in the OrdinalBins component. This will bypass sorting at the NodeList level.

bryfox:
We may also want to look at the NodeList implementation here. If nodes don't change, the list doesn't update its state.

This makes sense on the NameGenerator, where we do want to re-use the panels between prompts. But in the dev protocol, for example, "Contact Frequency"/"Last six months" should not be the same as "Discussion Frequency"/"One". This is a minor issue, but I'd expect to see the animation even if nodes are the same in different buckets. (Currently, nothing happens, including updating scroll position.)

PR #781, opened 2018-11-09
Feature/categorical bin

### LINKED ISSUE #156: Categorical bin interface

rebeccamadsen:
This resolves #156, adding a categorical bin interface. To test, look at the `Contact types` screen on the development protocol.

Still need to add tests and update wiki, but feedback welcome meanwhile.


public/protocols/development.netcanvas/protocol.json
```diff
@@ -1351,6 +1419,47 @@
...
+            },
+            {
+              "property": "last_name",
+              "direction": "desc"
+            }
+          ]
+        },
+        {
+          "id": "discussType",
+          "text": "Discussion type variable, with a much longer prompt than the previous prompt. Please also note that this prompt uses the max number of bins.",
```
bryfox: Should we enforce a 'max number of bins' in the schema?

src/containers/Interfaces/CategoricalBin.js
```diff
@@ -0,0 +1,78 @@
...
+          backward={promptBackward}
+          prompt={prompt}
+          prompts={prompts}
+        />
+      </div>
+      <div className="categorical-bin-interface__bucket">
+        <MultiNodeBucket
+          nodes={nodesForPrompt}
+          listId={`${stage.id}_${prompt.id}_CAT_BUCKET`}
+          id={'CAT_BUCKET'}
```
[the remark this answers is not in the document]
rebeccamadsen: Good point. It's actually required by `DropTarget`, and `MultiNodeBucket` is a `DropTarget`. We could make `id` required for `MultiNodeBucket` so it's more transparent...but on the other hand I don't think `MultiNodeBucket` needs to be a `DropTarget` at all. Maybe it used to be, but neither `CategoricalBin` nor `OrdinalBin` allow drops on the bucket.

As part of [OrdinalBin 2.0](https://github.com/codaco/Network-Canvas/issues/547), it was suggested that this and `NodeList` become one in the force. Or `NodeList` and `CardList` may [unite](https://github.com/codaco/Network-Canvas/issues/549). Or all three. I'm not sure what the answer there is, but for now I could just remove the `DropTarget` from `MultiNodeBucket` and then `id`s wouldn't be needed at all.

src/components/MultiNodeBucket.js
```diff
@@ -116,13 +116,14 @@ class OrdinalBinBucket extends Component {
           nodes.map((node, index) => (
             index < 3 && (
               <NodeTransition
-                key={`${node[nodePrimaryKeyProperty]}`}
+                key={`${node[nodePrimaryKeyProperty]}_${index}`}
```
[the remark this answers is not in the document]
rebeccamadsen: I added index to make the node refresh when its position changes. I made only the first node draggable, so when a node changes position the draggable status may change.

src/components/MultiNodeBucket.js
```diff
@@ -116,13 +116,14 @@ class OrdinalBinBucket extends Component {
           nodes.map((node, index) => (
             index < 3 && (
```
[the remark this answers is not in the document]
rebeccamadsen: I agree -- I can update that.

public/protocols/development.netcanvas/protocol.json
rebeccamadsen: I know Josh indicated that we would never want more than 8, so we could enforce that in the schema. I did not enforce a max in the code so there could be more, but the layout is designed with that max in mind. Eventually the layout will overflow (depending how wide your screen is), and the component is not scrollable, so that is good question -- should the code enforce that max? The schema? Or should the component just allow scrolling as a backup plan?

bryfox: If we want to enforce it, the schema should suffice. I opened #785 which can track this.

src/containers/CategoricalList.js
```diff
@@ -0,0 +1,168 @@
...
+
+import { makeNetworkNodesForType, makeGetVariableOptions, makeGetPromptVariable, makeGetNodeDisplayVariable } from '../selectors/interface';
+import { actionCreators as sessionsActions } from '../ducks/modules/sessions';
+import { CategoricalItem } from '../components/';
+import { MonitorDragSource } from '../behaviours/DragAndDrop';
+import { getCSSVariableAsString } from '../utils/CSSVariables';
+import { getNodeAttributes, nodeAttributesProperty, nodePrimaryKeyProperty } from '../ducks/modules/network';
+
+const colorPresets = [
+  getCSSVariableAsString('--cat-color-seq-1'),
```
wwqrd: Just something to flag here, I had problems with calling `getCSSVariableAsString()` for constants in Architect, I think because since we don't know if the DOM is ready when this code is called (i.e. variables may not be defined yet).

My solution was to call these methods at render time, though there might be a more elegant solution.

rebeccamadsen: Good to know! I can update that.

PR #783, opened 2018-11-12
Fix Server export format

### LINKED ISSUE #718: Server export needs to transpose variable IDs to names
Follow-on to #648. #713 fixed graphml download, but the transposition should probably be a separate step.

- [x] Refactor id to name translation (output: plain object)
- [x] Update server export to use this format
- ~Update graphml export to use this format~

Edit: After looking at the GraphML export, I don't think it makes sense to refactor it to use the format. It would reduce some duplication, but increase the work done since it's tightly coupled to the types defined in the variable registry. A better approach might be to refactor in the future to use centralized 'selection' logic.

bryfox:
Fixes #718. Depends on #774 for protocol changes.

This corrects the export format for server by transposing IDs to names.

Unknown props are passed through as mentioned in the issue, but this is easily changed with a flag.


wwqrd:
Looks good to me

PR #806, opened 2018-12-06
Fix/categorical bin formatting

### LINKED ISSUE #794: Categorical bin animation tweaks
Should the expanding hover move other bins (and rows) around?
jostle mov

Should the transition be animated on drop?
snap mov

rebeccamadsen:
Fixes #794.

* Adds animation for bins on node drop.
* Bins only expand on hover by the margin size so that node placement isn't affected.
* made bins larger for n<4
* reduced text size and increased bins for smaller, collapsed bins.
* smaller nodes in the expanded bin.


jthrilly:
Brilliant! Absolutely nailed it.

PR #808, opened 2018-12-10
Feature/narrative hulls

jthrilly:
Adds the ability to use a categorical variable to render convex hulls around nodes.

Remaining issues:

- [x] undefined group should not render as a group
- [x] drag and drop is broken
- [x] resize window should redraw hulls
- [x] performance should be reviewed
- [x] overlapping group membership should work. not tested.

To test: in development mode, generate test alters, or add nodes and set the categorical variable in the new node form to one or more values. Visit the narrative screen.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3276,6 +3276,19 @@
                 }
             }
         },
+        "concaveman": {
+            "version": "1.1.1",
+            "resolved": "https://registry.npmjs.org/concaveman/-/concaveman-1.1.1.tgz",
+            "integrity": "sha1-bCSCWAslI874L8K+wAoEFebmgWI=",
+            "dev": true,
+            "requires": {
+                "monotone-convex-hull-2d": "^1.0.1",
+                "point-in-polygon": "^1.0.1",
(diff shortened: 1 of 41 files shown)


jthrilly:
There was an issue on iOS where all the hulls were offset from the top, which was caused by the use of `100vh` in the DOM tree all the way through from the protocol to the interface level. I replaced these values with `100%`, which fixed the issue. It may have other unintended consequences.

I also needed to set an explicit height on `.stage` to make this work. It was previously set to `height: 0`. It might have impacted the session navigation animation. @rebeccamadsen - can you take a look, since this was your feature and I think you understand it best?

jthrilly:
Node highlighting on sociogram is broken in this branch.

rebeccamadsen:
> Node highlighting on sociogram is broken in this branch.

It's broken on the parent branch too. `ConcentricCircles` and `NodeLayout` differ in their naming for some props (like `allowSelect` vs `allowHighlight`, and `highlightAttribute` vs `highlight`), and `ConcentricCircles` doesn't pass a `onSelected` function either.

PR #812, opened 2018-12-13
Fix sass-lint warnings

bryfox:


jthrilly:
Time to make these error on the CI, do you think @bryfox?

bryfox:
Works for me; I pushed an updated config.

Issue #764, opened 2018-10-29
Some field inputs don't save the expected values


bryfox:
- NumberInput: renders `input type="number"` but saves value as a string
- ToggleButton: renders a text input (with invalid `type="boolean"`)
- Toggle & Checkbox: render as expected, but submit either undefined (no prop added) or false depending on if I've interacted. (Maybe this one's intentional?)

bryfox:
It looks like [rehydration is inferring the `input type` from the variable type](https://github.com/codaco/Network-Canvas/blob/v4.0.0-alpha.8/src/selectors/forms.js#L15). This might be overriding ToggleButton's "checkbox" type, and explains how "Number" inputs work, as there's no Number/NumberInput component.

jthrilly:
I added a number type input. We need to retest this behaviour in the latest version of NC.

bryfox:
Behavior is the same; an input of type "number" results in a string value saved in redux.

To test:
- Start a session using the development protocol
- On the first prompt, click the 'add node' button
- Fill out the form (note that age has a number type)
- See ADD_NODES action or resulting state with a `string` age

wwqrd:
Just to note we expect the following issue to be resolved when in future, nodes are constructed with all variables set to blank/false by default.

> Toggle & Checkbox: render as expected, but submit either undefined (no prop added) or false depending on if I've interacted. (Maybe this one's intentional?)

PR #817, opened 2018-12-17
Fix/field types

wwqrd:
Update Number input to return an integer
Update Field to define ToggleButton and Number field components, and also to show an error message if field component isn't found (rather than defaulting to a Text input).

Requires: https://github.com/codaco/Network-Canvas-UI/pull/92
Fixes: #764


bryfox:
I pushed some tweaks here and in https://github.com/codaco/Network-Canvas-UI/pull/92. This looks good to me after those changes.

PR #821, opened 2018-12-17
Fix hidden input values

### LINKED ISSUE #819: Hidden input does not render a value
A hidden input defined by a protocol should render the supplied value when added to a form; it no longer does.

bryfox:
Fixes #819.

This restores values to hidden inputs defined by a protocol. There are two parts to this:

- Change the return type of `getDefaultFormValues`. Based on the [`NodeForm` usage](https://github.com/codaco/Network-Canvas/blob/1aee497e17883e080fad714d9cf01ae9cb309ba4/src/containers/NodeForm.js#L89), I'm pretty sure this was the intent.
- Populate the value from the `input` prop. Based on [Redux Field docs](https://redux-form.com/6.0.0-alpha.4/docs/api/field.md/), I think this is what we'd expect, but I'm not very familiar with redux form, so I could be off here...

When using the app, I'm expecting that the form input contains the value, and that the generated node contains the corresponding attribute.

### DIFF
--- a/src/containers/Field.js
+++ b/src/containers/Field.js
@@ -33,7 +33,7 @@ const fieldTypes = {
   [FormComponent.ToggleButtonGroup]: ToggleButtonGroup,
   // In the case of the hidden input component { value } isn't actually passed along, but since
   // this component is a placeholder, assume the interface for now.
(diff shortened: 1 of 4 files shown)


jthrilly:
Looks good!

PR #822, opened 2018-12-17
Update roster configuration API to match edit interfaces in Architect

jthrilly:


bryfox:
I'm not sure if there are other changes coming in this PR, but 753e5b7 looks good.

PR #823, opened 2018-12-18
[WIP] Feature/narrative control panel

jthrilly:


rebeccamadsen:
Outstanding to dos:

- [ ] colors for convex hulls and the list are not in sync (e.g. on categorical interface, prompt Four, put two nodes in the purple "Two" category; on narrative, this creates a blue hull but "Two" is purple in the side list -- should the hull be purple, or the list show blue, though?)
- [ ] make featureNode functional
- [ ] toggling the edges, hulls, and highlights on/off right now is triggered by clicking on the accordion div; this should probably be a toggle button for each accordion, unless that's too cluttered
- [ ] allowRepositioning -- is this the same as `allowPositioning` in `ConcentricCircles`?
- [ ] create a reset button (right now things reset whenever changing presets or leaving the stage)
- [ ] toggle switch to turn annotation mode on/off?
- [ ] update wiki page
- [ ] update/create tests

jthrilly:
I decided to leave the featuring functionality pending some discussion, as I think the interface is looking very busy as things are and it might not be needed.

PR #825, opened 2018-12-18
Add types to external data in development protocol

wwqrd:
In Architect (and probably NC), type is a required property on external data so we can know what subset of the data we can use for each interface.

Related to: https://github.com/codaco/Architect/issues/351

### DIFF
--- a/public/protocols/development.netcanvas/protocol.json
+++ b/public/protocols/development.netcanvas/protocol.json
@@ -697,6 +697,7 @@
     "HIVServices": {
       "nodes": [
         {
+          "type": "eda5e3bb-8e1c-4216-9e06-adc0ff6b7f73",
           "attributes": {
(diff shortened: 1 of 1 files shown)


public/protocols/development.netcanvas/protocol.json
```diff
@@ -1092,6 +1110,7 @@
         {
           "id": "6cl",
           "text": "Who would _you_ consider to be a **close friend**?",
+          "type": "eda5e3bb-8e1c-4216-9e06-adc0ff6b7f73",
```
bryfox: I don't think this one belongs; maybe just a bad copy/paste? If we delete it, the rest looks good.

wwqrd: 👍

PR #831, opened 2019-01-02
Fix categorical bin rendering for window resize and unusual window sizes.

### LINKED ISSUE #809: Resizing the app window on the categorical bin interface causes rendering issues
screenshot 2018-12-11 at 17 18 55

wwqrd:
Resolves #809


wwqrd:
The key should probably be the bin value, or mirror the `CATBIN_ITEM` id, because the index will change.

PR #836, opened 2019-01-09
Feature/quick node add

jthrilly:
This PR adds a new experimental feature: a "quick add node form".

## Concept

We developed the node form concept to allow researchers to customise the attributes collected about an alter as they are elicited. This allows the flexibility to choose to collect one or more variables right from the start, thereby balancing the response burden associated with naming an alter against more efficient data collection.

However, this system is sub optimal if the researcher desires extremely low response burden for naming alters. It would require a node form with a single field. Such a form would be slow to show, would obscure existing nodes (which may spur faster recall if visible), and requires additional clicking (unless using add another mode).

## Implementation

This feature implements a new property, `quickAdd`, which is a boolean attribute on the name generator interface. If enabled, clicking the add a node button will not show a node form. Instead the user will see a single input box containing placeholder instructions ("Type a name and press enter..."). This form will remain open as long as the input is focused, allow the user to rapidly create new nodes by typing a label and pressing the enter key.

The form uses the name generator's subject, and looks up the `displayLabel` property of the stage subject. It creates a new node with only this property set.

`quickAdd` is not a required property, and defaults to false. If present, it takes precedent over the `form` property. Importantly, this means that editing of nodes created via a quick node form is not possible. Instead the intention is for the node to be erased and recreated.

### DIFF
--- a/src/containers/Interfaces/NameGenerator.js
+++ b/src/containers/Interfaces/NameGenerator.js
@@ -151,6 +151,7 @@ class NameGenerator extends Component {
           <QuickNodeForm
             stage={this.props.stage}
             addNodes={this.props.addNodes}
+            nodeIconName={nodeIconName}
           />
         }
         <NodeBin id="NODE_BIN" />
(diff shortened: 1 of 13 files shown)


wwqrd:
This looks amazing!

src/containers/Interfaces/NameGenerator.js
```diff
@@ -123,14 +149,15 @@ class NameGenerator extends Component {
           </div>
         </div>
 
-        { form &&
+        { (form && !quickAdd) &&
```
wwqrd: Maybe use a variable, `formEnabled` for example.

wwqrd:
Looks like this builds on the selector PR, so maybe should be rebased before a final review.

PR #840, opened 2019-01-15
Update to new asset manifest protocol format, and implement basic external data loading.

wwqrd:
Updates to the new assetManifest protocol format specified here: https://github.com/codaco/Architect/issues/363

In order to support this change I've added rudimentary component to manage network (json) asset loading onto components `withExternalData`.

Tested for development environments on iOS/Android/OS X, but not builds.

Partially completes: https://github.com/codaco/Architect/issues/363


wwqrd:
Assuming everything is more or less okay with this.

It will need rebasing on top of changes from https://github.com/codaco/Network-Canvas/pull/835, so suggest merging after that one.

wwqrd:
Rebased but having issues on iOS - XHR not working, unsure how this got missed before.

wwqrd:
I've added a fix for iOS, have removed request cancellation in the meantime because it doesn't make sense in the case of the readFile implementation. It needs to be re-implemented down the line, especially when we want to start supporting true async (https) resources.

PR #841, opened 2019-01-16
Feature/per alter form

### LINKED ISSUE #585: Per-alter form
An interface that shows an arbitrary form for each alter in the current interview network.

rebeccamadsen:
Resolves #585, adding a per-alter form interface: 

per alter form

This updates each node on swipe or when clicking next/previous on the stage buttons. It does not allow switching nodes when the form is in an invalid state.

### DIFF
--- a/src/containers/Interfaces/AlterForm.js
+++ b/src/containers/Interfaces/AlterForm.js
@@ -10,12 +10,16 @@ import { nodeAttributesProperty, nodePrimaryKeyProperty } from '../../ducks/modu
 import { getDefaultFormValues } from '../../selectors/forms';
 import { makeNetworkNodesForType } from '../../selectors/interface';
 import { protocolForms } from '../../selectors/protocol';
+import { Progress } from '../../ui/components';
 import { Form } from '../';
 
 class AlterForm extends Component {
   constructor(props) {
     super(props);
     this.swipeRef = React.createRef();
+    this.state = {
+      activeIndex: 0,
+    };
   }
 
   formSubmitAllowed = index => (
@@ -54,7 +58,7 @@ class AlterForm extends Component {
     } = this.props;
 
     const params = {
-      containerClass: 'alter-form swiper-container',
+      containerClass: 'alter-form__swiper swiper-container',
       direction: 'vertical',
       slidesPerView: 'auto',
       centeredSlides: true,
@@ -69,8 +73,13 @@ class AlterForm extends Component {
             const submitIndex = this.swipeRef.current.swiper.previousIndex;
             if (!this.formSubmitAllowed(submitIndex)) {
               this.swipeRef.current.swiper.slideTo(submitIndex);
-            } else if (this.props.formDirty(`NODE_FORM_${submitIndex - 1}`)) {
-              this.props.submitForm(`NODE_FORM_${submitIndex - 1}`);
+            } else {
+              if (this.props.formDirty(`NODE_FORM_${submitIndex - 1}`)) {
+                this.props.submitForm(`NODE_FORM_${submitIndex - 1}`);
+              }
+              this.setState({
+                activeIndex: this.swipeRef.current.swiper.activeIndex,
+              });
             }
           }
         },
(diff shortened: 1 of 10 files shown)


src/schemas/protocol.schema.v1.json
```diff
@@ -200,6 +201,12 @@
             "$ref": "#/definitions/Item"
           }
         },
+        "introductionPanel" : {
```
bryfox: `items` is for array types; I think this should read

```
"introductionPanel": {
  "$ref": "#/definitions/IntroductionPanel"
},
```

src/selectors/forms.js
```diff
@@ -43,21 +42,3 @@ export const makeRehydrateForm = () =>
     [propStageForm, protocolForms],
     (form, forms) => forms[form] || null,
   );
-
-/**
- * @return {Object} keyed by form name
- */
-export const getDefaultFormValues = createSelector(
```
bryfox: This seems to be needed to populate hidden fields on the NodeForm. To test:

1. Start session with dev protocol
2. Fill out & submit form to create new node

Variable "04298634..." should equal 'network-canvas', but is null in this branch.

src/containers/Interfaces/Information.js
```diff
@@ -18,7 +18,7 @@ const TAGS = [
   'thematicBreak',
 ];
 
-const getItemComponent = (item) => {
+export const getItemComponent = (item) => {
```
bryfox: Since we only care about `text` on the new interface (it's hard-coded, not dynamic), it might make sense to expose ReactMarkdown as a component (with a `content` prop) that we can use in AlterForm as 'normal' JSX. We wouldn't need this export.

src/containers/ProtocolScreen.js
```diff
@@ -17,9 +17,26 @@ import { getCSSVariableAsNumber } from '../ui/utils/CSSVariables';
...
+    if (previousProps.stage.id !== this.props.stage.id) {
+      this.interfaceRef = React.createRef();
+      this.forceUpdate();
+    }
+  }
+
   // change the stage to the next
   onClickNext = () => {
-    if (!this.props.stage.prompts || this.props.isLastPrompt()) {
+    if (this.getInterfaceRef() && !this.getInterfaceRef().isStageEnding()) {
```
bryfox: I'm not sure what might cause the throw in getInterfaceRef, but should the result be cached in a variable here?

src/containers/ProtocolScreen.js
```diff
@@ -28,13 +45,24 @@ class Protocol extends Component {
 
   // change the stage to the previous
   onClickBack = () => {
-    if (!this.props.stage.prompts || this.props.isFirstPrompt()) {
+    if (this.getInterfaceRef() && !this.getInterfaceRef().isStageBeginning()) {
```
bryfox: Same question as for `onClickNext`

src/containers/Interfaces/AlterForm.js
```diff
@@ -0,0 +1,217 @@
...
+  );
+
+  isStageEnding = () => (
+    this.swipeRef && this.formSubmitAllowed(this.swipeRef.current.swiper.activeIndex) &&
+    this.swipeRef.current.swiper.activeIndex === this.props.stageNodes.length
+  );
+
+  clickNext = () => {
+    if (this.swipeRef && this.swipeRef.current) {
+      if (this.swipeRef.current.swiper.isEnd) {
```
bryfox: Given the `!this.getInterfaceRef().isStageEnding()`, is it possible to reach this statement? The functionality seems to be handled by `clickPrevious`.

It would be nice to have this logic encapsulated in a single callback rather than having both 'next' and 'last' handlers here, but that might be out of scope for this PR.

src/containers/Interfaces/AlterForm.js
```diff
@@ -0,0 +1,217 @@
...
+      activeIndex: 0,
+    };
+  }
+
+  onSlideTransition = () => {
+    if (this.swipeRef.current && this.swipeRef.current.swiper) {
+      const submitIndex = this.swipeRef.current.swiper.previousIndex;
+      // try to submit (on fail, form is "touched" and errors can show)
+      this.props.submitForm(this.getNodeFormName(submitIndex));
+      if (!this.formSubmitAllowed(submitIndex)) {
```
bryfox: UX note: it felt unusual that I couldn't go *back* if the form didn't validate. By contrast, on the modal NodeForm, I can always back out by cancelling the modal if I'm not ready to fill the required fields for some reason. Here, I'd have to use the settings/timeline menu.

src/containers/ProtocolScreen.js
```diff
@@ -28,13 +45,24 @@ class Protocol extends Component {
...
+      this.getInterfaceRef().clickPrevious();
+    } else if (!this.props.stage.prompts || this.props.isFirstPrompt()) {
       this.props.changeStage(`${this.props.pathPrefix}/${this.props.previousIndex}?back`);
     } else {
       this.props.promptBackward();
     }
   }
 
+  getInterfaceRef = () => {
```
bryfox: Would be nice to document the contract for interfaces using this functionality (e.g., implement `clickLast`, `isEnding`, etc.) Maybe there's a better place, but here would work.

src/selectors/forms.js
jthrilly: So _that's_ what this was for! Thanks.

There's no mechanism currently to create hidden variables in a form in Architect or elsewhere. I feel like I've been back and forth on this a million times, but for the sake of clarity let's just remove hidden inputs types.

src/containers/Interfaces/AlterForm.js
jthrilly: I thought this, too. I think going back should always be allowed. Perhaps something like "if the navigation is backwards and the form is invalid, don't submit the form".

src/selectors/forms.js
bryfox: Sounds good; I imagine `additionalProperties` covers most of the same use cases. I opened #843 to cover the rest of this.

src/containers/Interfaces/AlterForm.js
```diff
@@ -0,0 +1,217 @@
...
+    return (
+      stageNodes.map((node, index) => {
+        const nodeAttributes = node ? node[nodeAttributesProperty] : {};
+
+        const initialValues = {
+          ...nodeAttributes,
+        };
+
+        return (
+          <div key={node[nodePrimaryKeyProperty]}>
```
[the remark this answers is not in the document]
rebeccamadsen: I pulled out the slide content as a PureComponent -- it seems to lag less for me, but let me know what you think. There *should* be a way to only render [visible content](https://idangero.us/swiper/api/#virtual), but I can't get it to work with anything other than strings so it's not as helpful as I would hope. If we don't need to use finger swipes to advance, we should consider not using Swiper and then we could just render the visible form(s).

src/schemas/protocol.schema.v1.json
rebeccamadsen: I updated this, hopefully I did it right. :)

src/selectors/forms.js
rebeccamadsen: Thanks for the open issue.

src/containers/ProtocolScreen.js
rebeccamadsen: Done.

src/containers/Interfaces/Information.js
rebeccamadsen: Done.

src/containers/Interfaces/AlterForm.js
rebeccamadsen: It was possible...if you tried to advance but it wasn't a valid form and so wasn't ready to move to the next Stage. I combined the logic and simplified a bit because we moved away from supporting swiping.

src/schemas/protocol.schema.v1.json
bryfox: It probably doesn't matter, but I don't think `"type": "object"` is needed here (adjacent to the #ref), since the `IntroductionPanel` type reference defines the object type. (example: "skipLogic", below.)

src/containers/Interfaces/AlterForm.js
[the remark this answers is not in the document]
bryfox: Nice — much better! If visible content is possible, that might be good to address in the future if users see performance issues (the first render in particular), but this works well for now.

wwqrd:
I'd just like to say I think this looks great!

PR #851, opened 2019-02-03
Feature/alter alter form

jthrilly:


wwqrd:
Does this PR add an alter edge form?

Looks like fix/remove-hidden-inputs needs to be merged first, but apart from that this looks good to me.

PR #852, opened 2019-02-03
Fix/edge attribute model

jthrilly:


src/ducks/modules/network.js
```diff
@@ -91,20 +127,22 @@ export default function reducer(state = initialState, action = {}) {
         nodes: (
           () => state.nodes.map(
             (node) => {
-              if (node[nodePrimaryKeyProperty] !== action[nodePrimaryKeyProperty]) { return node; }
+              if (node[entityPrimaryKeyProperty] !== action[entityPrimaryKeyProperty]) {
+                return node;
+              }
 
               // If the node's attrs contain the same key/vals, remove them
```
wwqrd: I know that this isn't part of the change, but these lines removing properties don't look necessary to me?

rebeccamadsen: Was this part of when we "toggled" boolean attributes to be present or null instead of marking them true/false? Did we change that behavior?

jthrilly: I think this is more like if you try to add a node with a UID that already exists. Am I misunderstanding which lines you are referring to?

jthrilly:
> Would it make sense to create a method like `getEntityAttributes(entity)` for ids? That way you wouldn't need to know the structure of the entity and a change like the one in this PR would have only required changing that one function.

Absolutely that would make sense. I did think about it, but this was just an "on the plane" kind of hack-fest.

src/ducks/modules/network.js
wwqrd: I was referring to [135-139](https://github.com/codaco/Network-Canvas/blob/f49d2eddcac3be017e33d351b11786947b40195a/src/ducks/modules/network.js#L135,L139), but I mistakenly thought this was removing the props only to add them again [142-148](https://github.com/codaco/Network-Canvas/blob/f49d2eddcac3be017e33d351b11786947b40195a/src/ducks/modules/network.js#L142,L148).

That said, since we now add default attributes to new nodes, maybe this is incorrect for another reason? - Shouldn't the props get reset to default rather than removed entirely?

rebeccamadsen: Makes sense to me to update this behavior in another issue.

jthrilly: I think this action (`TOGGLE_NODE_ATTRIBUTES`) is probably a bad approach in retrospect. There is simply too much going on in terms of the logical implications of toggling attributes of various types. 

If anything a refactor would look like `toggleAttributes(nodeID, booleanAttribute)`, but I think we should probably not use it at all. I think the sociogram was the only place to use it, and I updated that to use `updateNode` instead, so this should probably just be removed entirely.

Issue #837, opened 2019-01-09
Add another / finished behaviour problem


berniehogan:
When creating a node, toggling the add another check box behavior works as expected (though I still maintain that the reason for a checkbox is not that great wrt fullscreen, the behavior is so unconventional!) 

When clicking on a node dragged in, when I checked the add another box, it created a duplicate of the original node (in the developer protocol), or didn't do anything (teaching protocol). It's a bug that seems to emerge from what is already a very strange flow (as opposed to "cancel" | "add another" | "next". 

This might be the result of the 'finished' not clearing the data in the node panel, so that when I select 'finish' and the page doesn't clear (because add another is checked). Thus, I mistakenly think that I have not added a node (there is modest feedback in the page behind the modal dialog but I can't rely on that feedback since it might be fully obscured by the dialog). I guess that this makes 'not clearing the text' the bug. But I think it's caused more fundamentally by the strange design of the add another checkbox.

rebeccamadsen:
This looks like it was resolved in #854, so can it be closed?

PR #854, opened 2019-02-07
remove add another option from main node form

jthrilly:
Resolves issue #837 by removing the "add another" option from the node form.

The thinking behind this is that the quick add mode now provides a mechanism for sequential adding, and this removes UI clutter.

Corresponding PR in Architect forthcoming.


wwqrd:
Looks good, I removed a couple of un-needed work-arounds now that feature isn't available.

PR #856, opened 2019-02-08
Feature/ego form

### LINKED ISSUE #584: Ego form interface
An interface that displays an arbitrary form, where the values of the form are assigned to the ego node.

rebeccamadsen:
Resolves #584. 

ego interface

Implements ego interface.

This also updates export for ego and edge attributes, and fixes a bug with graphml export.

Small note: there is a fix in codaco/Network-Canvas-UI#98 for the error message that doesn't show on the third ego form.

Will add tests and update wiki before merge.


wwqrd:
Couple of things I noticed about the styles:

- Animations don't match the alter form ones, that doesn't seem right to me. I think the Alter ones are the look to go for of the two:
demo: alter, then ego

- Content overflows on iPad aspect ratio (some of the checkbox styling is a little off too, but possibly unrelated to this PR):
screenshot 2019-02-12 at 11 04 59

src/containers/Interfaces/AlterForm.js
```diff
@@ -124,7 +126,7 @@ class AlterForm extends Component {
               key={index}
               node={node}
               index={index}
-              updateNode={updateNode}
```
wwqrd: Just a small thing, it's better practice not to define handlers in the render method at all, otherwise the output is considered changed and the dom is always refreshed:

```jsx
class SlideForm extends PureComponent {
  handleSubmit = (formData) => {
    const { node, onSubmit } = this.props;
    onUpdate(node, formData);
  };

  render() {
    return (
      <Form
        onSubmit={this.handleSubmit}
      />
    );
  }
}

class AlterForm extends Component {
  handleUpdate = (node, formData) => (
    this.props.updateNode(node[entityPrimaryKeyProperty], {}, formData));
  };

  render() {
    return (
      <SlideFormNode
        node={node}
        onUpdate={this.handleUpdate}
      />
    );
  }
```

rebeccamadsen:
>  Animations don't match the alter form ones

I made them not match on purpose, to help make the visual connection that on an ego interface, all the questions apply to the same person, but on the alter interfaces each form is for a different person. It would probably simplify things to use the same animation for both, though, if that's preferred!

I'll go back to the pesky overflow again and adjust those handlers. Good catch!

src/containers/Interfaces/AlterForm.js
rebeccamadsen: Good call.

wwqrd:
That's a good point, they do play different roles.

I still think the card switch is much better visually. Maybe we can split the difference and have the "You" node remain stationary on the page while the form segment rotates? @jthrilly Any thoughts?

rebeccamadsen:
Or possibly just not have a "You" node at all? It was the "You" node scrolling away that I disliked, even if a new one took its place. It's not strictly necessary unless we feel it adds visual reminders who we are answering questions about.

I am seeing the checkbox strangeness in master, so this looks like a separate issue.

jthrilly:
Having thought about this some more, I am going to make some more fundamental changes to this interface. Please hold off on any further work.

PR #859, opened 2019-02-13
Fix/subject root property

### LINKED ISSUE #759: Move subject definition to 'stage' for all interfaces
We've had some discussion about changing how `subject` is defined.

Currently, it may be defined at either the stage or prompt level; we may want to standardize on one subject per stage. (This might reduce confusion with conflicting subjects in forms and panels.)

jthrilly:
Resolves #759 

Moves `background`, `sortOrder`, and `subject` to the stage configuration level, and out of the prompt level.

This standardizes all our interfaces.

Forthcoming PR in architect to support this.

### DIFF
--- a/public/protocols/development.netcanvas/protocol.json
+++ b/public/protocols/development.netcanvas/protocol.json
@@ -1036,15 +1036,25 @@
       "id": "sociogram",
       "type": "Sociogram",
       "label": "Sociogram",
+      "subject": {
+        "entity": "node",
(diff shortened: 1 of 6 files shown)


wwqrd:
Happy with how simple this was to implement!

jthrilly:
I decided that `sortOrder` should remain a prompt-level property, so this needs to be correct here before merging. 

Once that has happened we need to verify that Architect and NC can understand each other's config.

PR #862, opened 2019-02-15
Feature/refine narrative interface

### LINKED ISSUE #839: Refinements to narrative interface
Feedback suggested a number of refinements to the narrative interface

- We would like to add the ability for narrative annotations to persist. This should be implemented as a 'pin' toggle button in the preset switcher area.
- Remove multiple colors for node highlighting in favour of either a radio group allowing a single attribute to be selected to be shown, or a checkbox group. In either case, node selected state should be white.

rebeccamadsen:
Resolves #839. Requires codaco/Network-Canvas-UI/pull/99, to register when annotations have finished fading.

This allows for "freezing" annotations, pinning them in place at full opacity. Unfreezing restarts the fade effect for all annotations.

This also update the highlight attributes to be selected as part of a radio group, allowing users to toggle which attribute is shown. Color highlighting has been removed in favor of just the white highlight.


jthrilly:
Looks great!

PR #866, opened 2019-02-20
hide freeze button when free draw disabled

rebeccamadsen:
Hides the "freeze" button when freeDraw is disabled for the protocol.


jthrilly:
Nice improvement - love the attention to detail.

PR #874, opened 2019-03-04
add case id for server protocols

### LINKED ISSUE #871: imported protocol not prompted for Case ID
When selecting a protocol imported from Server, I am not prompted for a Case ID. Instead, NC uses "CREATE_NEW" for the session id, and has no case id.  This is a problem, because this first interview then gets overwritten any time a new imported protocol "creates" a session.

rebeccamadsen:
Fixes #871. To verify, connect to Server and select a new Server protocol. It should prompt for a case ID, just as the existing protocols do.


wwqrd:
Looks fine to me 👍

PR #875, opened 2019-03-05
default for options

### LINKED ISSUE #870: categorical bin and ordinal bin edge case
Categorical bin and Ordinal Bin do not handle some edge cases well. Either edge cases should be prevented in the protocol (Architect?), and/or should be handled more gracefully in NC. For example, if an ordinal variable or categorical variable has no options, both interfaces show this error: `cannot read property 'map' of undefined`. 

Additionally, we should consider what the interfaces should do in the case that a categorical or ordinal variable is selected that does not have a restriction on how many options can be selected for a node. If a node is in Category 1 and Category 4, that will do unexpected things. Is this a case we need to prevent in Architect, an issue to handle, or both?

rebeccamadsen:
Fixes #870. This ensures that `null` categorical and ordinal `options` are handled gracefully, and adds a test for that. 

To check, remove options from the variable registry for a categorical or ordinal variable used in an interface, and see that the interface still loads.

### DIFF
--- a/src/schemas/protocol.schema.v1.json
+++ b/src/schemas/protocol.schema.v1.json
@@ -948,7 +948,7 @@
           "items": {
             "$ref": "#/definitions/OptionElement"
           },
-          "minItems": 1
+          "minItems": 2
         }
       },
       "required": [
--- a/src/schemas/protocol.schema.v1.json
+++ b/src/schemas/protocol.schema.v1.json
@@ -947,7 +947,8 @@
           "type": "array",
           "items": {
             "$ref": "#/definitions/OptionElement"
-          }
+          },
+          "minItems": 1
         }
       },
       "required": [
--- a/src/schemas/protocol.schema.v1.json
+++ b/src/schemas/protocol.schema.v1.json
@@ -954,7 +954,26 @@
       "required": [
         "type"
       ],
-      "title": "Variable"
+      "title": "Variable",
+      "allOf": [
+        {
+          "if": {
+            "properties": {
+              "type": {
+                "enum": [
+                  "ordinal",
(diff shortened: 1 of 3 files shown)


wwqrd:
Could you clarify the schema update for me? Will this reject protocols without options?

rebeccamadsen:
The schema would reject `options` of length 0. `options` itself is not required, it just has a minimum length if it exists. Although maybe it's possible to make `options` required for certain variable types, like categorical and ordinal??

jthrilly:
I think the preferable thing would be to require options _and_ require options.length > 1. I think having 1 option is equally as invalid, since this should be modeled as a bool.

jthrilly:
In #870 you ask about where this validation should be enforced - I think the correct place is the protocol schema, which will shortly be externally synchronized across both Architect,  Network Canvas, and Server. 

 As a reminder, the principle is "Architect only opens and creates valid protocols, and Network Canvas only imports valid protocols". At all points, the protocol schema is the source of the validation requirements.

wwqrd:
Should we make options required in the schema before merging this? Looks good otherwise.

rebeccamadsen:
Options is required for categorical and ordinal variables [here](https://github.com/codaco/Network-Canvas/blob/b19d49b48f02dba3b38f943ce6fac5a037e81880/src/schemas/protocol.schema.v1.json#L960-L975), and when present, the schema expects at least [2 items](https://github.com/codaco/Network-Canvas/blob/b19d49b48f02dba3b38f943ce6fac5a037e81880/src/schemas/protocol.schema.v1.json#L951).

wwqrd:
My bad, looks finished then!

PR #886, opened 2019-04-09
Fix/issue 883

### LINKED ISSUE #883: droppable status in side panel
"Already mentioned" vs "People from your previous visit" panels are inconsistently triggering when dragging nodes from the main panel to a side panel. It varies from one prompt to the next, and when you restart a session. Sometimes you can delete a node via "people from your previous visit", and sometimes not. Sometimes it errors out. 

For example, drag "Baz" to the main panel. It is droppable on both side panels, even though it isn't present anywhere else in the session. Drop it in "Already mentioned", and `promptIDs` is empty now, though the node is still in the network.

Another example, drag "Baz" to the main panel. Return to the start screen. Navigate to your session again. "Baz" is no longer droppable on "People from your previous visit" even though it should be.

It used to be that in both cases, "Baz" would be droppable on "People from your previous visit" but not on "Already mentioned". I think it should just not be droppable on "Already mentioned" if that's the only prompt in the list.

jthrilly:
Resolves #883.

> Another example, drag "Baz" to the main panel. Return to the start screen. Navigate to your session again. "Baz" is no longer droppable on "People from your previous visit" even though it should be.

This was a bad one. It took me 4 hours, and the solution ended up being one line!

The problem was that the <NodePanel /> component was triggering the generation of an incorrect map of node IDs in the parent component when the withExternalData HOC updated the externalData prop, because the update was being triggered inside of `componentWillRecieveProps`, and therefore happening before the externalData was available to create the map. Phew.

To say that again simply: `componentWillRecieveProps` was calling a function and passing nextProps as a parameter. Further down the stack was a reference to `this.props` that was undefined, because the new props hadn't reached the component yet. When the missing prop became defined later in the lifecycle, the update function was not called again because the comparison test (which depended on other props) was not passed.

Initial solution was to pass nextProps.externalData down the stack to the function that ultimately used it to calculate the node id map. I then realized I could just do the update later in the lifecycle, inside componentDidUpdate, and then not have to pass any props anywhere, since they would be available to all functions.


jthrilly:
I've decided to reinstate not allowing a node to go into the 'existing' panel when on the only prompt it has been nominated on.

PR #887, opened 2019-04-10
Feature/better external data

### LINKED ISSUE #885: Updates to external data
## Data formats
- Can we handle CSV data? One approach to this would be translating CSV into our network JSON format in Architect. Alternatively, we need to implement a CSV loader in network canvas.

## Allow the use of variable names OR uuids in external data

External data looks like this: 

```
{
  "nodes": [
    {
      "type": "eda5e3bb-8e1c-4216-9e06-adc0ff6b7f73",
      "attributes": {
        "8a35cd77-7bc4-4c7e-b98a-673b6a21321f": "About My Health",
        "8ee3a187-d4be-458e-8abb-71efcc071949": "",
        "1e9ce62c-44c5-484d-9c26-0d20cc7d7238": "About My Health",
        "66646d68-cd33-407b-a349-56707021df72": "About My Health",
        "18fbc928-d027-42de-bc96-ff5c09bf4944": "",
        "67132d2b-c371-4c57-a5eb-6520083f9d22": "55 E Washington St, Chicago, Illinois, 60602",
        "bdc60147-fe7a-4c3c-a164-e5b370f6a281": 41.882945,
        "931a7b23-e433-4e7e-8e13-48b72e5f0549": -87.62572,
        "0a5f7952-a9a7-4e87-9353-2d17a59284ee": "55 E Washington St, Chicago, Illinois 60602, United States"
      }
    },
}
```

When external data is loaded, Network Canvas should remap variable names according to the following logic:

- Try to look up the variable in the codebook, assuming UUID.
- If that fails, search the codebook for the variable name and replace with UUID.
- Same applies to `type` (if present).

Away from this, we need to investigate how 'type' works. Type should specify how the nodes are rendered BEFORE they are brought into the interview. When brought into the interview, the subject for the stage determines how the node looks.

jthrilly:
Resolves #885 

- Allows external data that uses variable names rather than UUIDs to be imported, and makes a "best guess" at converting them to UUIDs on import. 
- Variables that cannot be mapped to the codebook are left unchanged
- Allows CSV files to be specified as external network data, and converts them to network JSON when loaded.


jthrilly:
This needs to be tested on cordova before merging. Load development protocol from settings to do this.

wwqrd:
Already acknowledged with @jthrilly, csv isn't working for cordova (ios + android).

jthrilly:
In response to the poor performance of this, I made a stab at moving it to a web worker. The long term idea would be to move all blocking code to workers.

It has definitely helped, but not as much as I had hoped. 

I think this means that the long, long term solution is to calculate this data once at a non performance critical moment (such as starting an interview session, or when the protocol is installed), and then store the decoded and restructured data. Then, stream it in as we do with media assets.

wwqrd:
I think it would be great to have this generate at interview start.

Do we have any way of determining which bits are slow? If it's due to having large files/objects in memory then we may need to think of alternatives, since we can't stream json.

PR #889, opened 2019-04-11
Feature/resume interview 2

### LINKED ISSUE #550: Resume Interview 2.0

### LINKED ISSUE #884: Ability to push interview data to any paired server with current session protocol
We need the ability for a NC client to push to any paired server that has the current session's interview protocol available. 

This may already be implemented, but we have to validate.

jthrilly:
Resolves #550 and #549

Main purpose of this PR is to redesign the resume interview screen to allow for batch exporting protocols to server, filtering and sorting, and drag to delete.

To test:

- Create some sessions
- Visit the resume session screen. Test filtering, sorting, and drag to delete.
- Tap some sessions to select them, and use the export flow. Try to break this, by doing various devious things to disrupt the export.

Changelog:
- resolves #884 (protocol not installed on server). User sees error message indicating this.
- allows pressing enter to submit the new session (case ID) form
- fixes MDNS on iOS by updating zeroconf plugin
- allows batch export of interviews
- updates quick add mode to new variable format


rebeccamadsen:
This looks great! I know it's a lot, but much of it worked nicely. Here are the notes I took as I stepped through.

* Some node labels are missing ("null"). After creating edges using a node that came from the "NameGen with CSV external data" prompt, the labels on the nodes thereafter were `null`.
* The labels for "Links" on narrative interface are missing.
* Pressing enter twice quickly on the new node form gets you two nodes for the price of one. Or click really fast on the "Finished" button and you can get more.
* Export interview data to File tries to open a protocol instead of saving something? I could be misunderstanding what should happen?
* Importing a new protocol from Server gave a success message, but if I already had a protocol, the new protocol didn't show up in the protocol list until I navigated elsewhere and back again.
* Attempting to export a previously exported session gives a useful error message, but the option to "Delete successfully exported sessions?" made me think I could use that to delete this session, since it had once been exported successfully. It did not. I was able to delete the session by resuming the session, going to the finish screen, and selecting the delete option there.
* Successfully importing a protocol lands you on the "Import a protocol" view. It might be nice to land on the view of the new protocol if we think users are just importing one protocol at a time and using it right away? Not really an issue, just a thought.
* I was able to export from NC, though I got an error trying to export from Server. I haven't teased out if that's a problem with the version of the protocol I have on Server or something else. I can try again, but I'm out of time today.
* Android emulator seemed slower to respond than normal. Maybe I need to play with my settings? I am having trouble downloading and installing the development protocol there, but had success with a smaller protocol.
* Changing the interface scale in settings leads to visual distortions on the background image in Windows; they resolve if the screen rerenders (clicking resume session, for example).
interface scale distortions

jthrilly:
Thanks for this! Really useful.

Just for posterity:

> * Some node labels are missing ("null"). After creating edges using a node that came from the "NameGen with CSV external data" prompt, the labels on the nodes thereafter were `null`.

I need to look into it, but this sounds like it is intentional. The CSV file deliberately doesn't contain variables defined in the variable registry, and the node label worker hasn't been updated to handle that.

> * Pressing enter twice quickly on the new node form gets you two nodes for the price of one. Or click really fast on the "Finished" button and you can get more.

This is probably a bug in master, too? I don't think I have changed anything to do with this behaviour, but I will test.

> * Export interview data to File tries to open a protocol instead of saving something? I could be misunderstanding what should happen?

Ha! I actually haven't implemented this, and completely forgot. I am considering leaving this feature out, but I will see what I can do.

> * Attempting to export a previously exported session gives a useful error message, but the option to "Delete successfully exported sessions?" made me think I could use that to delete this session, since it had once been exported successfully. It did not. I was able to delete the session by resuming the session, going to the finish screen, and selecting the delete option there.

I'll adjust this to only show the option if one or more sessions were successfully exported.

> * Successfully importing a protocol lands you on the "Import a protocol" view. It might be nice to land on the view of the new protocol if we think users are just importing one protocol at a time and using it right away? Not really an issue, just a thought.

You're right - this does sound more likely. I think it is the way it is now because it makes it easier to test, but I will change it. 

> * Android emulator seemed slower to respond than normal. Maybe I need to play with my settings? I am having trouble downloading and installing the development protocol there, but had success with a smaller protocol.

This is going to be a painpoint in general, I think. Android is slower. I'll do some testing on devices and see whats up.


I'll be working on fixing the issues you identified today.

PR #894, opened 2019-04-24
add exit button and full screen app toggle

### LINKED ISSUE #872: should NC have a "quit" option when full screen
In full screen mode, the only way to quit NC in Windows is to alt-tab and close it on the taskbar, or "End Task" through the "Task Manager". Is that a problem for our users? I know we've been back and forth on this one.

rebeccamadsen:
Resolves #872. Adds an "Exit" button (if supported) and a toggle for full screen application (for electron).

image

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 public/protocols/**/*Worker.js
 src/utils/networkQuery
+src/utils/protocol
--- a/public/components/windowManager.js
+++ b/public/components/windowManager.js
@@ -3,6 +3,7 @@ const mainMenu = require('./mainMenu');
 const appUrl = require('./appURL');
 
 const isMacOS = () => process.platform === 'darwin';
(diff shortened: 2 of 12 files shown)


wwqrd:
If the issue is that leaving the app is an opaque process in windows, I'm not confident that this solves that?

rebeccamadsen:
Do you have a suggestion? I understood that we don't want to make it too easy for users to leave the app, but offer something beyond a keyboard shortcut they might not know?

wwqrd:
I can't really think of anything satisfying:

- We could have an always visible "close fullscreen" button, like a video player? (maybe just on Windows)
- We could not default to fullscreen on Windows?
- When in fullscreen we could show an always visible close button?

jthrilly:
I think this solution is fine, but because the experience of full screen is so bad on windows, we should default to windowed mode there.

PR #903, opened 2019-05-10
Preview mode 2.0 (Architect)

wwqrd:
Changes to enable Network Canvas to be controlled through electron IPC.

- Adds `.app--preview` in order to disable certain UI if `global.NETWORK_CANVAS_PREVIEW` is set
- Adds `remote` modules which hook into IPC and redux.
- Fixes (?) Environment checkers which didn't seem to be calling `isElectron()`

Required by: https://github.com/codaco/Architect/pull/478

Partially resolves: https://github.com/codaco/Architect/issues/361


rebeccamadsen:
Looks awesome! I get an error when navigating to the sociogram with background image for the development protocol (`TypeError: Cannot read property 'then' of null`), but I think that's just an outdated link to the development protocol? It was only with the "import development protocol" version; my local development protocol file loaded the background image just fine. Other assets look good.

wwqrd:
@rebeccamadsen I've made a separate PR for those changes causing the issue, we resolve dev protocol there.

wwqrd:
I've rebased this onto master, so there shouldn't actually be any visable changes.

rebeccamadsen:
Looks good to me!

PR #909, opened 2019-05-28
Fix for #899

### LINKED ISSUE #899: Ego form does not scroll smoothly on iOS
The ego form interface does not scroll smoothly on iOS.

jthrilly:
Root cause was that changing the DOM while momentum scrolling (which is enabled by `-webkit-overflow-scrolling: touch;`) caused momentum to be lost, appearing to the user as scrolling getting "stuck". We change the DOM to toggle the appearance of the progress-bar, and to toggle some text when scrolled to the bottom.

This PR adds environment detection for Android and iOS, and uses it to disable these two actions. iOS users will not see the progress bar animate in, nor will they see "click next to continue" appear when they get to the bottom of the screen.

Fixes #899.

### DIFF
--- a/src/containers/Interfaces/EgoForm.js
+++ b/src/containers/Interfaces/EgoForm.js
@@ -1,11 +1,12 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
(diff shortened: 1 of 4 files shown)


wwqrd:
Looks good to me.

Issue #334, opened 2018-02-08
[iOS] protocol import field does not always display keyboard


bryfox:
iPad/iOS 11. No hardware keyboard.

When I tap the protocol import text field to focus, the keyboard shows but the view doesn't always scroll up as expected (so that the input field would be visible). User has to dismiss keyboard and re-tap the field.

This happens maybe 1/3-2/3 of the time.

May be an issue deeper in the stack, but I only see it on this field, and cannot reproduce elsewhere (e.g., on the filter input for Roster list).

![field-focus](https://user-images.githubusercontent.com/39674/35984806-2ff45a0e-0cc3-11e8-955c-66512ec8399d.jpeg)

(edit: removed incorrect statement about logs.)

bryfox:
This appears to be related react "controlled component" inputs, but possibly only within a redux form? 

- Replacing the field with a plain input (same styling) does not have the issue
- Replacing the ReduxFormField on that screen with a standalone controlled component input still has the issue
- The controlled component input on the RosterList does not have the issue.

jthrilly:
Keep sharing anything you find Bryan. I have a feeling this issue will crop up elsewhere.

bryfox:
This is not an issue with UIWebView, nor localhost running in mobile safari; only WKWebView.

[This cordova issue](https://issues.apache.org/jira/browse/CB-12037?jql=project%20%3D%20CB%20AND%20labels%20%3D%20wkwebview-known-issues%20AND%20text%20~%20%22input%22) about [a removed pref](https://developer.apple.com/documentation/uikit/uiwebview/1617967-keyboarddisplayrequiresuseractio) seemed related, but I tried the stated workaround [as implemented by ionic](https://github.com/ionic-team/cordova-plugin-ionic-webview) and could still reproduce the problem — even though that webview allows me to set focus programatically/arbitrarily.

So I still really don't know what's happening. If there's a timing issue involved, it doesn't seem to be because of iOS preventing artificial focus.

bryfox:
It's not timing; it's position. Depends a bit on the device/sim used (as layout can vary slightly), but tapping around the top of (or just above) the field causes the issue in the screenshot; tapping the bottom area does not.

Removing `display: flex;` from the app__content makes things display as expected; otherwise, changing height-affecting display properties on various elements on the screen can worsen or improve the situation. I guess the flex layout calculations confuse whatever logic is in the webview to scroll the viewport when a field takes focus.

[...]

PR #910, opened 2019-05-28
add cordova keyboard plugin, switch iOS to px font sizing

### LINKED ISSUE #900: auto focused inputs cause incorrect rendering on iOS
If an input in a modal is automatically focused, the cursor renders in an incorrect position when the modal opens on iOS

jthrilly:
Fixes #900.

The behavior with iOS viewports is complicated, and we have changed it several times to address different issues. In #334 we removed a meta viewport property that makes iOS scroll to a focused input because it was causing content to be pushed off of the screen.

In this PR I add a Cordova plugin that makes iOS behave the same way as Android, which is to shrink the viewport when an input is focused. This means the above issue no longer occurs, and the property can be reinstated.

As a consequence, iOS can no longer use viewport units for font sizing. I updated the default device setting to reflect this. I also removed the ability to set dynamic font sizing on Cordova entirely, since it will break the app. I actually think having all mobile on PX sizing, and all electron on dynamic by default (with the option for px) is better.

This has numerous other benefits, including making the quick add input visible. However, it is a big change for this late in the day, and so some testing would be welcome.

Small note: I also reduced the overall font size to match the defaults in UI (16px and 1.675vmin).

### DIFF
--- a/src/utils/DeviceInfo.js
+++ b/src/utils/DeviceInfo.js
@@ -72,7 +72,7 @@ const deviceDescription = () => {
 };
 
 // Disable dynamic scaling on android because vmin is resized by software keyboard
-const shouldUseDynamicScaling = () => !(isCordova() && device.platform === 'Android');
+const shouldUseDynamicScaling = () => !(isCordova());
 
 export default deviceDescription;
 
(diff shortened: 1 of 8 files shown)


src/utils/DeviceInfo.js
```diff
@@ -72,7 +72,7 @@ const deviceDescription = () => {
 };
 
 // Disable dynamic scaling on android because vmin is resized by software keyboard
-const shouldUseDynamicScaling = () => !(isCordova() && device.platform === 'Android');
+const shouldUseDynamicScaling = () => !(isCordova());
```
wwqrd: `!isCordova()`?

PR #917, opened 2019-05-31
Feature/scrollbars

jthrilly:
Adds scrollbars to the app to help with the discoverability of scrollable containers. These are custom styled, and so not as obtrusive as native scrollbars.

Screen Shot 2019-05-30 at 11 50 10 PM

Outstanding issues:

- The use of `-webkit-overflow-scrolling: touch;` disables custom scrollbars entirely on iOS, but without it, there is no momentum scrolling (native feeling) on iOS.
- If the above were acceptable, removing the property does show the custom scrollbar, but iOS uses reverse direction scrolling, so attempting to drag the scrollbar down will scroll the container down (opposite behavior of windows, android, linux). 

Mitigating factors:
- Scrolling is probably best on iOS devices. 

For now, the plan would be to have custom scrollbars for all platforms apart from iOS.


jthrilly:
Nobody is particularly keen on this, but I still believe it is the correct approach. I am going to implement the feature behind a setting, which will default to "off"

Issue #929, opened 2019-06-17
Check for updates silently on app start


wwqrd:
App automatically checks for changes on start, and notifies the user if changes are available. It does not notify user if no updates are avaible.

**Currently it also throws an error when the network is down.** We want this to happen when a user checks for updates manually, but not on app start up.

This needs implementing accross all apps:
- [x] Network Canvas
- [ ] Server
- [ ] Architect

PR #930, opened 2019-06-17
Update silently

wwqrd:
Updates updater to fail silently when `checkForUpdates(silent = true)`

Partially resolves https://github.com/codaco/Network-Canvas/issues/929


wwqrd:
Tested locally but not verified as working.

PR #931, opened 2019-06-19
freeze button should not show when no freeDraw

### LINKED ISSUE #879: Narrative freeze button displays when "freeDraw" is missing
In Narrative interface, if `behaviours` is missing the `freeDraw` parameter, the interface interprets that as "false", and it does not allow drawing. I think that's the correct default behavior. However, the "freeze" button still displays; it does not display if `freeDraw` is defined as "false". The "freeze" button should also not display if free drawing is not defined.

rebeccamadsen:
Fixes #879. To check, create a protocol with no behaviors toggled (so freeDraw is undefined, not false). The "freeze" button (snowflake) should not display. With a protocol like development protocol that has freeDraw toggled true, the snowflake should display still.

### DIFF
--- a/src/containers/Interfaces/Narrative.js
+++ b/src/containers/Interfaces/Narrative.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { compose } from 'recompose';
+import { get } from 'lodash';
 import PropTypes from 'prop-types';
 
 import { getProtocolStages } from '../../selectors/protocol';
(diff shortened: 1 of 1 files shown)


wwqrd:
Works for me!

I noticed that unless allow repositioning is explicitly set to false (e.g. is never set) then the nodes are positionable, would that be worth fixing here too?

jthrilly:
> I noticed that unless allow repositioning is explicitly set to false (e.g. is never set) then the nodes are positionable, would that be worth fixing here too?

Isn't this fixed in architect with a default value?

wwqrd:
That rings a bell but I can't find an issue for it? Maybe because NC can parse it, we assumed it behaved as defaulting to false?

Assuming we can't track that other issue down: https://github.com/codaco/protocol-validation/issues/11 would be a case for updating the behaviour in NC.

jthrilly:
We agreed after having some issues with default values that our logic would be that NC would interpret the absence of a property as a false value, and architect would never provide null or undefined property values. These two things together should mean this works.

If the absence of the property value is not being treated as false by NC that is the bug I guess. I vaguely remember manually making this property default to true because it seems intuitive. However the way to do that now we have more comprehensive logic is to implement it as a default configuration in Architect, so we will need an issue there.

wwqrd:
Okay, lets handle those as a separate from this issue then

https://github.com/codaco/Architect/issues/553
https://github.com/codaco/Network-Canvas/issues/933

PR #944, opened 2019-06-27
Feature/remove protocol files

### LINKED ISSUE #935: No way to remove an interview protocol from device
- Update "reset app data" to also clear protocol files 
- Add ability to remove an individual protocol (from state, and from disk).

wwqrd:
- Reset app now deletes /protocols/ folder in app directory.
- Adds delete button to protocol cards

Also:
- Adds promise functionality to dialogs, e.g.

  ```js
  dispatch(openDialog()).
    then((confirm) => {
      if (!confirm) {
        console.log('abort!');
        return;
      }
  
      console.log('do thing');
    });
  ```

Resolves #935


rebeccamadsen:
This definitely removes protocols from the /protocols directory. It was great at preventing a delete when existing sessions hadn't been exported, at deleting a protocol if there were no existing sessions, or deleting the entire protocol directory when I reset all data.

However, I was able to delete a protocol with an existing session that I had already exported. The session I had exported stuck around, and I could attempt to resume the session. I got a blue screen and had to reset data to get back to the start screen. I think we should either: (1) auto-remove exported sessions for that protocol, or (2) disallow deleting a protocol if there are *any* sessions -- exported or unexported.

jthrilly:
Or third option - not allow resuming sessions where there is no corresponding protocol in `installedProtocols`.

Zooming out and looking at the initial intent of this, I think intentionally removing a protocol would imply that any sessions (regardless of export status) attached to that protocol should also be removed. 

However we need to make this super clear to users. Perhaps deleting a protocol with no sessions is a "confirm" dialog, but deleting one with sessions is a "warning", and they have specific text about each scenario.

wwqrd:
So to summarise the changes required:
- Ensure when a delete action is performed on a protocol, associated sessions are also deleted.
- Allow the user to delete a protocol that has non exported sessions (but show a warning)
- Add an additonal warning when: deleting a protocol that has sessions, but which have been exported.

Making the final behaviour:
  - If a protocol has non exported sessions: allow deletion but show warning dialog
  - If a protocol has sessions that have been exported: allow deletion but show confirm dialog
  - If a protocol has no sessions: allow deletion and confirm deletion

Issue #927, opened 2019-06-10
Refactor ProtocolScreen


wwqrd:
Can we improve the `onClickNext()`, `onClickPrevious()` methods?

Suggestions:
- Generalise the `skipToIndex` logic so that it can be used by both methods
- Can we make the distinction clearer between? `interfaceRefInstance.clickPrevious()`, `promptBackward()`, and `this.props.changeStage(`${this.props.pathPrefix}/${skipToIndex}`);`
- Can we update the logic to remove the nested if/else?
- Can we remove `interfaceRefInstance`? If not, can we make the distinction clearer betwwen clickPrevious and promptBackward?

jthrilly:
Plan for this is to replace the current start screen with a sortable and filterable `CardList` with the default sort option being by most recently installed.

PR #948, opened 2019-07-26
Fix/multiple ego forms not submitting

### LINKED ISSUE #946: Ego properties are not being exported for csv individual export
Both union and individual exports are missing many of the answers to ego question.

There is an example protocol on slack here: https://netcanvasteam.slack.com/archives/G1Q262J3Y/p1562770387011900

### LINKED ISSUE #947: RadioGroup is not considered completed if value is `0`
Screenshot 2019-07-16 at 18 00 24

Example protocol here: https://netcanvasteam.slack.com/archives/G1Q262J3Y/p1562770387011900

rebeccamadsen:
Fixes #946. The issue was not with exporting, but rather with multiple consecutive ego forms. This ensures that consecutive ego forms aren't a problem, and fixes an issue with ego default values not populating correctly.

Also fixes #947. We had [updated protocol validation to allow it](https://github.com/codaco/Network-Canvas/pull/925/), but had not updated field validation to accept `0` as valid.

### DIFF
--- a/src/containers/Form.js
+++ b/src/containers/Form.js
@@ -155,6 +155,7 @@ export default compose(
   connect(makeMapStateToProps),
   autoInitialisedForm,
   reduxForm({
+    enableReinitialize: true, // form could have ego out of sync because submit is in progress
     touchOnChange: true,
     touchOnBlur: false,
     onSubmitFail: scrollToFirstError,
--- a/src/containers/Interfaces/EgoForm.js
+++ b/src/containers/Interfaces/EgoForm.js
@@ -26,7 +26,7 @@ const TAGS = [
   'thematicBreak',
 ];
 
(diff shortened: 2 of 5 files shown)


wwqrd:
This fix appears to work fine for EgoForm, it also fixes 0 value validations in the RadioGroups.

I'm not keen on the amount of indirection in the clickNext clickPrevious implementations, is there a way we could avoid the use of refs?

I think since this fixes the existing implementation it should still be merged.

jthrilly:
> I'm not keen on the amount of indirection in the clickNext clickPrevious implementations, is there a way we could avoid the use of refs?

I think this would fall under the remit of this existing refactor: https://github.com/codaco/Network-Canvas/issues/927

PR #952, opened 2019-08-01
bump version to 4.0.3

jthrilly:
Release branch for testing v4.0.3


jthrilly:
> Out of curiosity how are the versionCode and bundleVersion generated?

I manually increment these, since they are only used by the App Store and the Play Store respectively to uniquely identify the compiled app (independently of the version number, which is frustrating...).

Sometimes there is a mistake in the Cordova configuration that doesn't require a code change, but does need me to update config.xml. If I already submitted a binary with a given versionCode or bundleVersion I can't reuse it, so I will bump it again. This hasn't happened recently though, thankfully.

PR #954, opened 2019-08-07
ensure unique form names

rebeccamadsen:
This fixes an issue with consecutive alter edge forms (from the development protocol). Basically, similarly to the recent issue with ego forms, alter form and edge form names were not unique if the forms were consecutive in the protocol. Alter edge form updates just wouldn't be submitted correctly for the second (or later) AlterEdgeForm interface.

This updates the names to use the stage index so the forms are unique and submit without issue.


src/containers/Interfaces/AlterForm.js
```diff
@@ -126,6 +127,7 @@ class AlterForm extends Component {
               subject={stage.subject}
               node={node}
               index={index}
```
wwqrd: Should we change these to `edgeIndex`/`nodeIndex` now we have multiple indexes?

rebeccamadsen: Good point!

rebeccamadsen:
> Code looks reasonable, are there steps I can take to verify that this works corretly? I loaded the development protocol and didn't spot any obvious errors with or without this fix?

Steps to reproduce:
1. In master, add some nodes and edges to a session in development protocol.
2. Navigate to Alter Edge Form (friends).
3. Fill in some of the fields on the edges.
4. When you advance to Alter Edge Form (professional), use the back arrow to go back to the Alter Edge Form (friends).
5. The first edge form doesn't fill in the fields properly, and won't save any fields you change now.

In this fix, going back and forth between the edge form stages isn't an issue anymore.

wwqrd:
Thanks for the steps to reproduce, looks great!

PR #955, opened 2019-08-15
Integration testing starter for Network Canvas

wwqrd:
Setup and example tests for Network Canvas based on https://github.com/codaco/integration-tests.

This allows us to create network-canvas specific integrations tests, which have been split into two categorise:
- functional tests, that interactively test features of the app and interface
- regression tests, that compare image snapshots of specific views of the app

Regression image snaphots need to be generated locally first before they can be committed to the project and tested in CI. To run these tests run `docker-compose run update_snapshots`

To run locally:

```
npm install
npm run start:integration
npm run test:integration
```

To run on docker:

```
npm run test:integration:docker
```

Further instructions are available here: https://github.com/codaco/Network-Canvas/blob/25e0b035228e0d57e31e9aec6d2febc5b2792136/test/README.md

#### Notes
- Upgrades jest to version 23, which has some API differences.

### Todo
- [x] Visual snapshot testing, can we do diffs
- [x] Can we link this with CI?


test/test/start-screen.test.js
```diff
@@ -0,0 +1,64 @@
...
+    await fakeDialog.mock([{ method: 'showOpenDialog', value: mockFilenames }]);
+    await app.client.isVisible('.getting-started');
+    await app.client.click('[name=add-a-protocol]');
+    await app.client.waitForVisible('.protocol-import-dialog__tabs');
+    await app.client.click('.tab=Local file');
+    await app.client.waitForVisible('h4=Protocol imported successfully!');
+    await app.client.click('button=Continue');
+    await app.client.pause(timing.medium);
+    await app.client.click('button.overlay__close');
+    await app.client.pause(timing.medium);
```
jthrilly: Why does the test end with a pause?

wwqrd: This was to wait for the dialog to close, I've updated it to specifically wait for that now.

wwqrd:
Added an REACT_APP_TEST exception in src/App so that snapshots can be generated with the default window size which should be the same accross machines. I notice that the screenshots are captured at retina resolutions on my machine, so it would be good to get confirmation that this works on non-retina machines too.

wwqrd:
- How do you update a visual snapshot? Would be good to be able to do this selectively as we did with code snapshots

run with `-u` flag, I'll add that the readme

`npm run test:e2e -- -u`

- If we are serious about visual regressions, we will need find a way to repeat all snapshot tests at specific screen sizes. Of course there is platform specific CSS which we can't do much about at the moment, but we should at least test the breakpoint based CSS. Can we somehow automate this with a single set of screen size definitions?

Not with the v4 API, once we update electron we should be able to do that.

- Can we do anything more complicated than clicking? Im thinking about tests for dragging/drawing (think drag and drop, and sociogram annotation).

Yes we can do that, yet to see how successfully.

- It seems like the specification of the "target" element can use all the usual CSS selectors, and I see in the sample test there are a few different types. Do you think this might be fragile in the longer run? I wonder if we shouldn't start being more specific with testable elements that we click. Should we define a convention for naming, and a consistent way of targeting an element in a test? Maybe this is overkill - just a thought.

My thinking here is that we would add some kind of id references, it's not perfect.

- I think the next step (possibly in a separate PR) is to create tests for those core interactions we discussed, using the development protocol as the basis for it. The way I see it is that the development protocol is the sandbox for exposing the app functionality that we want to test with these interaction tests. One issue is how we get the development protocol into network canvas for the purposes of running these tests. We could go back to submoduling it in?

My plan for this was to use the built in UI. I'm not sure that submoduling it is an option since the development protocols aren't under version control as I understand it? We could add a script to fetch it.

jthrilly:
After discussion this morning, we agreed on the following next steps: 

- development protocol loading in, ready for tests to be backfilled against that
- (separate PR) backfill tests for core app interactions using the development protocol. can be done in stages.
- get it running on CI

wwqrd:
- I think `moveTo` is used by spectron, which has a sub dependency on webdriver. The version of spectron is pinned to the version of electron. This means that whilst we get a warning, the versions are stable and I would not expect the sub dependency to change. In terms of will changing electron change the API for spectron, yes it might, from what I've seen these are mostly additions, so hopefully nothing too drastic.
- I'm not sure about this, I envisaged there being 2x and 1x snapshots but maybe that's an inherrantly bad idea.
- We can move test/test and the docker container to something more meaningful (although RE docker, that should be run using the npm task anyway, so effectively already abstracted).

wwqrd:
Further to the snapshots quesion, you can run these using normal jest arguments to filter tests, e.g. `npm run test:integration -- functional`

PR #962, opened 2019-08-23
Feature/update electron

rebeccamadsen:
Updates electron to 5.0.9, node to 12.8.0, and npm to 6.10.2. Will need to run `npm install`. On windows, will also need to rebuild mdns to target this version of electron.

```
cd node_modules\mdns
node-gyp rebuild --target=5.0.9 --arch=x64 --dist-url=https://atom.io/download/electron
```

Note: this does not yet update react, cordova, or the various other dependencies, so we can separate concerns in the updating process.


rebeccamadsen:
I increased the size for assets and entrpoints before webpack 4 generates a warning, but we should consider [other optimizations](https://webpack.js.org/guides/code-splitting/).

jthrilly:
So should this run as things are? I get an unhandled promise rejection about a menu item:

```
(node:58869) UnhandledPromiseRejectionWarning: TypeError: Invalid template for MenuItem: must have at least one of label, role or type
    at Function.Menu.buildFromTemplate (/Users/joshua/Projects/Network-Canvas/node_modules/electron/dist/Electron.app/Contents/Resources/electron.asar/browser/api/menu.js:171:11)
    at setApplicationMenu (/Users/joshua/Projects/Network-Canvas/public/components/windowManager.js:12:24)
    at /Users/joshua/Projects/Network-Canvas/public/components/windowManager.js:66:5
    at new Promise (<anonymous>)
    at createWindow (/Users/joshua/Projects/Network-Canvas/public/components/windowManager.js:25:10)
    at Object.getWindow (/Users/joshua/Projects/Network-Canvas/public/components/windowManager.js:75:12)
    at Object.start (/Users/joshua/Projects/Network-Canvas/public/components/appManager.js:102:8)
    at App.<anonymous> (/Users/joshua/Projects/Network-Canvas/public/electron-starter.js:31:14)
    at App.emit (events.js:199:15)
(node:58869) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
```

The app launches to a white screen with "DevTools was disconnected from the page".

Going to look into if there is anything obvious.

public/components/windowManager.js
```diff
@@ -31,6 +31,9 @@ function createWindow() {
       minHeight: 800,
       center: true,
       title: 'Network Canvas',
+      webPreferences: {
+        nodeIntegration: true,
```
jthrilly: We had been attempting to remove the need for nodeIntegration, as per this discussion: https://github.com/codaco/Network-Canvas/issues/472

What are the remaining areas of functionality that require it?

src/index.js
```diff
@@ -43,9 +43,8 @@ const startApp = () => {
 };
 
 if (isElectron()) {
-  const { webFrame, ipcRenderer } = require('electron'); // eslint-disable-line global-require
+  const { webFrame, ipcRenderer } = window.require('electron'); // eslint-disable-line global-require
```
jthrilly: I'm a bit confused about the re-introduction of `window.require` - didn't we work around this here: https://github.com/codaco/Network-Canvas/pull/447 ?

If so, what has changed?

jthrilly:
I think the menu code is the issue, or at least the way it is imported. Commenting it out allows the app to load. This might help: https://stackoverflow.com/questions/55200616/how-can-a-menuitem-be-separated-into-its-own-module-in-electron

PR #965, opened 2019-09-02
Schema v2 

jthrilly:
This PR Updates Network Canvas to Schema v2
- Updates the start screen ProtocolCard component to show either a warning or an error state, depending on the schemaVersion of an installed protocol.
  - Warning is shown when protocol is compatible, but out dated. Session start still allowed.
  - Error (not currently used) will be shown for incompatible versions.
- This requires support for stage level node filtering on the following interfaces:
  - Alter Form
  - Alter Edge Form
  - Categorical Bin
  - Ordinal Bin
  - Sociogram
- Also updates networkQuery to allow skip logic using categorical/ordinal variables
- Also updates the developer protocol import button to the version implementing these features.


src/config.js
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable import/prefer-default-export */
+
+// Target protocol schema version. Used to determine compatibility & migration
+export const APP_SCHEMA_VERSION = 2;
+
+export const APP_SUPPORTED_SCHEMA_VERSIONS = ['1.0.0', 1, 2];
```
wwqrd: Does this make sense now that there is a plan to auto-migrate? e.g. if there is a migration path, it's supported, otherwise it isn't?

jthrilly: Discussed on slack, and resolved 👍

src/components/Setup/ProtocolCard.js
```diff
@@ -1,25 +1,126 @@
...
+        onClick={() => selectProtocol(protocol)}
+      >
+        <Icon className="start-button__protocol-icon" name="protocol-card" />
+        <div className="start-button__text">Start new interview</div>
+        <Icon className="start-button__arrow" name="chevron-right" />
       </div>
-      { protocol.description ? (<p className="protocol-card__description">{protocol.description}</p>) : ''}
     </div>
   );
 };
```
wwqrd: Consider making this a class and defining some of these staetments as constants outside the render method.

src/components/Setup/ProtocolCard.js
```diff
@@ -32,11 +133,22 @@ ProtocolCard.defaultProps = {
...
   }).isRequired,
 };
 
-export default ProtocolCard;
+function mapDispatchToProps(dispatch) {
+  return {
+    openDialog: bindActionCreators(dialogActions.openDialog, dispatch),
+    deleteProtocol: bindActionCreators(installedProtocolsActions.deleteProtocol, dispatch),
+  };
+}
```
wwqrd: React-redux will handle binding if you provide an object:
```
const mapDispatchToProps = {
  openDialog: dialogActions.openDialog,
  deleteProtocol: installedProtocolsActions.deleteProtocol,
};
```

src/components/Setup/ProtocolCard.js
```diff
@@ -32,11 +133,22 @@ ProtocolCard.defaultProps = {
...
+
+export { ProtocolCard as UnconnectedProtocolCard };
```
wwqrd: By convention these haven't previously specifed `Unconnected`.

wwqrd:
Testing of filters complete for:
  - [ ] Alter Form
  - [ ] Alter Edge Form
  - [ ] Categorical Bin
  - [ ] Ordinal Bin
  - [ ] Sociogram

wwqrd:
Alter form crashes on filter, perhaps related to includes rule?
Screenshot 2019-10-09 at 17 08 01

src/components/Setup/ProtocolCard.js
jthrilly: > By convention these haven't previously specifed `Unconnected`.

I'm not following on this one. Can you explain? We've used `UnconnectedComponentName` throughout for exports where we just want to test the unconnected component, haven't we?

wwqrd:
Filters don't appear to work for:
- Alter Edge Form
- Sociogram

With a new protocol I couldn't replicate the crash seen in the development protocol.

src/components/Setup/ProtocolCard.js
wwqrd: Maybe there are some exceptions, but I think the convention has been:
```
export { ConcentricCircles };

export default ConcentricCircles;
```

Usage:
```
// unconnected
import { ConcentricCircles } from './ConcentricCircles';

// connected
import ConcentricCircles from './ConcentricCircles';
```

jthrilly: Hmm. I actually think that's less clear. Its probably something we did in NC, but not elsewhere. Its pretty widespread though: https://github.com/codaco/Network-Canvas/search?q=Unconnected&unscoped_q=Unconnected

Unless you feel really strongly (in which case I would update all uses for consistency), I think this should stay the same.

jthrilly:
All updated, apart from the ProtocolCard refactor. I think its probably fine as-is.

src/components/Setup/ProtocolCard.js
wwqrd: I mean, I think it'd make more sense to have `ConcentricCircles` and `ConnectedConcentricCircles`, but that's worse in my opinion! Since this is only required for testing, I don't think it matters either way. The convention in Architect is without `Unconnected`, and at least that's the case for a few components in NC.

src/components/Setup/ProtocolCard.js
```diff
@@ -1,25 +1,125 @@
...
 import PropTypes from 'prop-types';
-
+import { connect } from 'react-redux';
+import cx from 'classnames';
+import { actionCreators as dialogActions } from '../../ducks/modules/dialogs';
+import { actionCreators as installedProtocolsActions } from '../../ducks/modules/installedProtocols';
 import { Icon } from '../../ui/components';
+import { APP_SUPPORTED_SCHEMA_VERSIONS, APP_SCHEMA_VERSION } from '../../config';
+
+const ProtocolCard = ({ protocol, openDialog, deleteProtocol, selectProtocol }) => {
```
wwqrd: I still think this should be a class. Creating handlers etc in the rendering method causes React to update the dom every render, regardless of whether it has changed.

jthrilly: > I still think this should be a class. Creating handlers etc in the rendering method causes React to update the dom every render, regardless of whether it has changed.

No sure, I just don't think that's a big deal with this component. The functions within render are trivial logic evaluations against an object.

rebeccamadsen:
Will other interfaces have node filtering eventually? Narrative, for example?

jthrilly:
> Will other interfaces have node filtering eventually? Narrative, for example?

Narrative was overlooked for now, as its system of presets might eventually warrant a more sophisticated approach. Perhaps filtering on the preset level.

PR #971, opened 2019-09-23
Update sessions removeEdge action to use edgeId

### LINKED ISSUE #923: sociogram "reset" doesn't delete edges
Using the hidden "reset" button on sociogram resets positions of nodes but does not remove existing edges. We should delete the edges as well. Action "REMOVE_EDGE" fires, but the network is not updated as expected.

wwqrd:
Update `REMOVE_EDGE` action creator to specify edge id. Also adds some testing.

Looking at the bigger picture, the session / network split might benefit with a little refactoring. I'm thinking along the lines of define the network (deferred) actions in the network module, and make a higher order function to wrap them in a thunk with a session id. I think this would make testing responsiblities a bit clearer and testing easier.

Resolves #923


wwqrd:
> Looking at the bigger picture, the session / network split might benefit with a little refactoring. I'm thinking along the lines of define the network (deferred) actions in the network module, and make a higher order function to wrap them in a thunk with a session id. I think this would make testing responsiblities a bit clearer and testing easier.

Having looked into this a bit more I'm not sure how much it'd improve things. For example, if we separate sessionId from the addNode action so it becomes: inSession(addNode(nodeAttributes)), where inSession simply adds the sessionId to the action, and the sessions reducer handles delegation to the network. The addNode action is still coupled to the session by the requirement of the protocolId. In other words, is it any clearer?

jthrilly:
> Having looked into this a bit more I'm not sure how much it'd improve things. For example, if we separate sessionId from the addNode action so it becomes: inSession(addNode(nodeAttributes)), where inSession simply adds the sessionId to the action, and the sessions reducer handles delegation to the network. The addNode action is still coupled to the session by the requirement of the protocolId. In other words, is it any clearer?

Its an interesting thought, though! There's a lot of interdependence between protocol, session, and network. I'm not sure how much is reducible, really. Might be worth a longer-term review/making an issue, though, as more clarity in the long run will help with maintainability and project sustainability.

For now I think we need to push ahead with feature development and bug fixes until we roll over into our final project year.

wwqrd:
I agree that are other priorities right now!

(Just jotting down some more thoughts)

I think the general approach would be to normalize the data, I did think about that a little bit, but given the 1-1 relationship between sessions and networks, it seems like the current approach is fine (protocols are already normalized).

One thing that might make the interfaces clearer (and the thing easier to test) would be to have action creator methods in network that sessions delegates to (rather than creating actions as objects in sessions).

PR #972, opened 2019-09-25
Fix external data node labels

jthrilly:
This PR fixes an issue where the node label was not being correctly set when using external data.


jthrilly:
I need to add test coverage for this before merging

jthrilly:
I'm also going to make this work with either 'Name' or 'name', as this is proving to be consistently difficult for users.

Issue #987, opened 2019-10-24
Skip logic fails for ego rules


wwqrd:
Skip logic is relying on a node count after using `filter()`. `filter()` doesn't apply to ego, and in this case a `query()` should be used.

However, the schema has been updated to drop the `asserts` property that `query()` requires, so this library needs to be updated to work with the new schema first.

### Edit

Updating to use the latest version of network query's `query()` method fixes the ego rule skip logic for the development protocol. But another related bug has been found:

If an ego form preceeds a stage with skip logic, the skip logic is assessed with an old version of the app state. If the skip logic contains an ego rule, this mean that the skip logic will evaluate incorrectly,

PR #990, opened 2019-10-30
Fix/skip stage 

wwqrd:
This updates the the ProtocolScreen and Interfaces to allow blocking previous/next actions.

This is necessary to allow form submission to happen before skip logic is evaluated. Previously this caused a bug with EgoForm and skip logic (https://github.com/codaco/Network-Canvas/issues/987), but was potentially an issue with AlterForm and AlterEdgeForm also.

Changes include:
- The helper methods on interfaces are now:
  - `beforeNext()`, called when the user clicks back/forwards (called by `ProtocolScreen` on the interface reference)
  -  and `onComplete()`, to be called by the interface when it is ready to move on (called by the interface on the passed prop)
- Prompt/Stage navigation logic is now handled by thunks in `ducks/modules/navigate`, open to better naming suggestions?

Depends on: https://github.com/codaco/networkQuery/pull/27
Resolves: https://github.com/codaco/Network-Canvas/issues/987


wwqrd:
Given these changes, I'd almost be tempted to have interfaces manage their own prompt navigation, it could even be a HOC for standard behaviour like the name generator.

E.g. the interface would receive a beforeNext(), and if still had prompts to navigate, it would do that, otherwise it would call onComplete(), and ProtocolScreen would take over for stage navigation.

That way the ProtocolScreen would only ever have to manage stage navigation. Whereas at the moment it handles that and *sometimes* manages within stage navigation.

jthrilly:
> Given these changes, I'd almost be tempted to have interfaces manage their own prompt navigation, it could even be a HOC for standard behaviour like the name generator.
> 
> E.g. the interface would receive a beforeNext(), and if still had prompts to navigate, it would do that, otherwise it would call onComplete(), and ProtocolScreen would take over for stage navigation.
> 
> That way the ProtocolScreen would only ever have to manage stage navigation. Whereas at the moment it handles that and _sometimes_ manages within stage navigation.

I think this would be "nice" rather than being a big improvement - although when you spell it out, it probably should have worked this way to begin with.

I would suggest we create a maintenance issue for it, reference this PR, and then move on for now.

I'll take a look at this today, and will produce a release branch for NC and Architect if we merge it.

rebeccamadsen:
Ego rules work for me, as does the skip logic after a form based interface! However, I noticed three things.

I can't click the "Next" or "Previous" Arrow buttons on the left side for the Ego interface. I get this:

```
Uncaught TypeError: Cannot read property 'props' of undefined
    at beforeNext (bundle.js:317796)
    at Protocol.<anonymous> (bundle.js:322599)
    at callCallback (bundle.js:194635)
...
```

That's for a valid or invalid submission. I can navigate to the next stage by pressing enter in a field.

In the development protocol, if I navigate to stage 22 "Skip or Show" using the menu before answering any ego questions, I get the following error:

```
TypeError: Cannot read property '3377af3f-3c79-41da-9b0b-6570fb519b93' of undefined
```

Finally, Alter form with no nodes and alter edge form with no edges can't proceed to the next stage. I see a `1 of 0` at the bottom but "next" does nothing.

wwqrd:
> Ego rules work for me, as does the skip logic after a form based interface! However, I noticed three things.

This was caused by beforeNext not being bound to the component. Should now be fixed.

> In the development protocol, if I navigate to stage 22 "Skip or Show" using the menu before answering any ego questions, I get the following error:
> 
> TypeError: Cannot read property '3377af3f-3c79-41da-9b0b-6570fb519b93' of undefined

This is caused by the ego node not having an attributes property. I can update networkQuery to handle that, but this actually seems like malformatted network to me?

> Finally, Alter form with no nodes and alter edge form with no edges can't proceed to the next stage. I see a 1 of 0 at the bottom but "next" does nothing.

That case wasn't property handled in the beforeNext handlers, they've both been updated and should now work.

rebeccamadsen:
The latest updates help improve the 0 nodes and edges case, and fix the ego next/previous situation. A couple new observations:

If I have edges, the alter edge form doesn't navigate to next, it stays on the intro screen. The alter node form doesn't have this problem.

On narrative on the development protocol, I get this warning, but I haven't found if it results in an issue. Sociogram didn't have the same error:
```
Warning: Failed prop type: The prop `stage` is marked as required in `ConcentricCircles`, but its value is `undefined`.
```

The oddest one is navigating backwards to the first stage, the stage index only goes to 1 and doesn't update to 0, so clicking forward again will skip a stage. There is a warning going backwards:
```
Warning: Hash history cannot PUSH the same path; a new entry will not be added to the history stack
```

wwqrd:
Thanks for catching those! I'm tempted to make the alter/edge iterators a shared HOC?

Should now be fixed:
- Alter edge form navigate away from the intro screen, caused by a flipped sign.
- Interview stuck on first screen, caused by broken url generator.

> On narrative on the development protocol, I get this warning, but I haven't found if it results in an issue. Sociogram didn't have the same error:
> 
> ```
> Warning: Failed prop type: The prop `stage` is marked as required in `ConcentricCircles`, but its value is `undefined`.
> ```

I think this is unrelated, as it happens on master too? (when refreshing the app whilst on the narrative screen)

rebeccamadsen:
This fixes things for me. Tests pass locally for me. I found one more thing, but am not sure if it is an intentional change. On the alter node/edge forms, in master we allow users to go backwards to the previous stage even if the current form is invalid. We only save the submission if the form is valid going backwards. Here, we only allow a valid form to move to the previous/next prompt/stage in either direction, so the only "out" is the menu. (Forward on master obviously requires a valid submission, just as it is here.)

wwqrd:
I didn't mean to change the behaviour. Why did it work that way? Would seem non-obvious to the user in my opinion? (Sometimes this saves, sometimes it doesn't). Maybe we could add a dialog? "Warning navigating away will loose changes"?

src/containers/Interfaces/AlterForm.js
```diff
@@ -71,18 +71,27 @@ class AlterForm extends Component {
 
   isIntroScreen = () => this.state.activeIndex === 0;
 
-  isLastAlter = () => this.state.activeIndex === this.props.stageNodes.length
+  isLastAlter = () => this.state.activeIndex === this.props.stageNodes.length;
```
jthrilly: is `stageNodes` calculated _after_ any stage node filtering has been applied? Just want to make sure on that one.

I initially didn't want automatic skipping for "empty" stages (incase the user wanted to use the information panel to show something to be seen by all participants). Its also less clear what is happening - when clicking a a stage in the menu, you get a prompt if that stage would be skipped based on skip logic, and can choose to force show it. You can't do that with this - its always going to be skipped, by the stage itself.

jthrilly: I think that considering we are keeping stage filtering and skip logic separate for now, we shouldn't ever automatically skip a stage. This should be under user control.

jthrilly:
> I didn't mean to change the behaviour. Why did it work that way? Would seem non-obvious to the user in my opinion? (Sometimes this saves, sometimes it doesn't). Maybe we could add a dialog? "Warning navigating away will loose changes"?

Yeah, I was surprised to see that it saved when going backwards when the form was in a valid state. I would expect it only to save when going forwards, and when the form is valid. I suppose there is a scenario where the user leaves the page to go back and add information they forgot. That's probably why it works that way. However, _not_ saving a form in an invalid state breaks that convenience. 

I think if the form can't be submitted, and the user navigates backwards, they should get a dialog that they will loose data they have entered.

src/containers/Interfaces/AlterForm.js
wwqrd: I think the naming is confusing things here, `shouldSkipAction` is unrelated to the skip logic. This just determines whether or not the stage should block the navigate action or not. I will change the name to make it clearer.

The `stageNodes` prop isn't changed at this is exactly what is displayed to the user when they navigate alters, it uses the `makeNetworkNodesForType` selectors which includes the filter function.

wwqrd:
> I think if the form can't be submitted, and the user navigates backwards, they should get a dialog that they will loose data they have entered.

So if the form is invalid AND they navigate away they get the dialog (with an option to continue), but if it's valid it should save and then navigate away?

src/containers/Interfaces/AlterForm.js
wwqrd: FYI the skip logic has been completely removed from the interface/stage components and moved to here: https://github.com/codaco/Network-Canvas/blob/d75099c3d1bf4fa48bad7c95dae50bad911967ee/src/ducks/modules/navigate.js#L33,L61

wwqrd: > I initially didn't want automatic skipping for "empty" stages (incase the user wanted to use the information panel to show something to be seen by all participants)

The current implementation maintains this behaviour (of showing info panels with no nodes/edges)

jthrilly: Thanks for the explanation. FWIW I wasn't confused about this in the way you think - I was referring to this line because you changed the code following it in a commit called "Skip empty alter and alter edge forms". I wanted to ensure that (1) `stageNodes` took into account filtering for the definition of 'empty', and (2) 'empty' per alter and per alter edge forms were not being automatically skipped (as the title of the commit implies). Seems like the confusion was that I was assuming `shouldSkipAction` was immediately skipping the stage, when it isn't.

jthrilly:
I would tend to lean towards not creating a situation with accidental data loss, but we could also achieve that by warning when navigating backwards when the form is dirty. It just made more sense to me to save if possible, and warn if not. Totally happy to go with what you think is best though!

src/containers/Interfaces/AlterForm.js
wwqrd: Yep, should skip action has nothing to do with stage skipping, I've renamed it to `shouldContinue`, which hopefully clarifies intent a little. I also added some explanation of the beforeNext method inline with the code.

rebeccamadsen:
Just checked the most recnt update. I think we still have data loss if the user navigates away using the menu, but that happens on other interfaces too, so not new. Otherwise this helps.

However, I'm not seeing form data save in redux for edges. The user selection is there until you navigate to another stage and then back again. The alter node form looks like it saves though.

Also, when going backward with a clean form, we go back to the previous node or edge. When we go backward with an invalid form, it skips to the previous stage entirely even if there were other nodes/edges before the previous stage.

wwqrd:
> Also, when going backward with a clean form, we go back to the previous node or edge. When we go backward with an invalid form, it skips to the previous stage entirely even if there were other nodes/edges before the previous stage.

wwqrd:
> However, I'm not seeing form data save in redux for edges. The user selection is there until you navigate to another stage and then back again. The alter node form looks like it saves though.

This appears to be resolved now, I think it may have been due to the updateEdge action creator being called incorrectly, which has been resolved in the refactor.

rebeccamadsen:
Everything is looking good on my run through. If attempt to click previous when we are on the first stage, though, can we disallow that? Because it prompts that I'll lose my form data, but I don't actually lose anything because we don't change the stage.

wwqrd:
> Everything is looking good on my run through. If attempt to click previous when we are on the first stage, though, can we disallow that? Because it prompts that I'll lose my form data, but I don't actually lose anything because we don't change the stage.

I've added this in, it only applies to EgoForm at the moment, because the others would be on the intro screens at that point. I've made it it so instead of warning it attempts to submit the form (and show errors).

PR #997, opened 2019-11-19
Fix navigation to take into account lack of finish screen

wwqrd:
New navigation changes didn't take into a potential account lack of finish screen (which is omited in preview mode).

This change adds a `screenCount` which is separate from `stageCount` to take into account the additional screen.


rebeccamadsen:
I'm having trouble advancing to the next stage from a stage with prompts. For example, on the development protocol, I can't advance forward from stages like name generators and sociogram and categorical, etc.

wwqrd:
Looks like I missed some brackets on the logic - should be fixed now 👍

PR #1002, opened 2019-12-02
Date picker

wwqrd:
Add new DatePicker component for the datetime input.

Screen Recording 2019-12-11 at 15 50 11 mov

Notes:
- day and month are optional (with types: 'full' (default), 'month', and 'year')
- date is ISO formatted:
  - full: '2019-12-00'
  - month: '2019-12'
  - year: '2019'
- can be navigated to with 'tab', but not edited with keyboard

Differences from above video
- 'clear' button is shown as soon as panel is opened, so that the panel can be closed.
- 'out of range' dates are more dulled out.

Feedback on draft:

- ~Try multi-column (paginated) years (5 per row)~
- ~For condensed view, shrink panels as progress is made in picker.~

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20127,6 +20127,12 @@
         "yallist": "^2.1.2"
       }
     },
+    "luxon": {
+      "version": "1.21.2",
+      "resolved": "https://registry.npmjs.org/luxon/-/luxon-1.21.2.tgz",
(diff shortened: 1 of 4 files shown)


wwqrd:
Sorry this is actually *not* ready yet, we need a corresponding protocol validation update.

PR #1008, opened 2019-12-16
Prevent missing name tag causing a crash in scrollToFirstError

wwqrd:
This may be fixed elsewhere but was causing issues on master.


jthrilly:
I think this should be a crash. If a component doesn't have a name corresponding to its variable ID, something has gone wrong (although didn't we change this to `data-name` since `name` wasn't valid on anything other than `<input>`?

PR #1011, opened 2020-01-09
Fix/android rendering

jthrilly:
This PR updates the Swiper component used on the start screen and the alter form/alter edge form interfaces. It also restyles these interfaces slightly (moves the preview node/edge to the top of the form and adjusts the transition). To achieve this required rewriting two components in a functional style so they could use hooks.

The previous version had rendering issues on Android, which this addresses.


src/containers/Setup/ProtocolList.js
```diff
@@ -14,106 +14,96 @@ import { actionCreators as uiActions } from '../../ducks/modules/ui';
...
-    });
-  }
+  const [showNewSessionOverlay, setShowNewSessionOverlay] = useState(false);
+  const [selectedProtocol, setSelectedProtocol] = useState(null);
 
-  handleCreateSession = (caseId) => {
-    this.props.addSession(caseId, this.state.selectedProtocol);
-    this.handleCloseOverlay();
-  }
+  const onClickProtocolCard = (protocolUID) => {
```
wwqrd: Should we stick to the `handle` naming convention?

src/containers/SlidesForm/SlidesForm.js
```diff
@@ -30,206 +20,196 @@ const confirmDialog = {
...
    */
-  beforeNext = (direction) => {
-    if (this.shouldContinue(direction)) {
-      this.props.onComplete();
-      return;
-    }
-
-    if (this.isIntroScreen() && direction > 0) {
-      this.nextItem();
+  const beforeNext = (direction) => {
```
wwqrd: How did you find this new API?

PR #1015, opened 2020-01-22
Improved rendering on devices with small screens and large font sizes

jthrilly:
This is a bulk PR, designed to address several items of user feedback about rendering on Android, particularly on devices with small screens and large font sizes. Specifically:

- Fixes inability to scroll ordinal bins on android. This was caused by a breaking change to flexbox rendering in Chrome 71 (and thus would also have impacted electron eventually).
- Implements ellipses for bin titles to avoid overflow issues. The user should vary the length of the bin title and/or the font size of the device to avoid truncation.
- _Removes_ ellipses for categorical bin rendering, since we have more vertical height to play with.
- Implements line clamping for prompt titles beyond the third line. This may need to be reverted following testing, since multi-line ellipsis (line clamping) somehow isn't in CSS. Check for weird clipping artifacts on long prompts - should be a clean linear fade out.
- Reduces the height of the prompt area, generally.
- Restyles the list-select interface to be more space efficient
- Restyles all "filter" inputs (session menu and list-select interface) to look more like inputs.
- Fixes an issue causing the categorical bins to clip when hovered over.
- Adjusts the height of the per-x-form interface forms to not clip over the progress bar at the bottom.
- Widens the ego form so that it can be scrolled from the sides.
- Makes the finish interview screen scrollable when there is insufficient vertical height to fit on all content. This will become more important as we refactor this screen to show other export options.
- Removes the settings menu from within the interview, in preparation for #1017. Adjusts right margin.

Bonus:
- Updates UI to include a fix for the RelativeDatePicker component that caused an app crash.


src/styles/containers/_ego-form-interface.scss
```diff
@@ -24,12 +25,17 @@
 
   &__form-container {
     height: 100%;
-    max-width: 60rem;
     width: 100%;
+  }
+
+  &__form-container-scroller {
```
wwqrd: Can any of these changes be fed back into the regular scroller styles?

jthrilly: I think these specifically concern the ego form, so no unfortunately not.

PR #1017, opened 2020-01-29
Miscellaneous feedback for small screen devices

jthrilly:
This PR is the home of a new menu style for NC designed to allow us to save space. It also includes miscellaneous small fixes for issues identified by a tester that related to this.

To do: 
- [x] Fix integration tests
- [x] Single column alter form layout
- [x] Font sizing issues - Fixed in https://github.com/codaco/Network-Canvas-UI/pull/126
- [x] Native menu items
- [x] Multi line ellipsis not working on iOS 
- [x] Menu open animation stagger doesnt work with long lists 
- [x] Dummy content for new settings sections


rebeccamadsen:
In general this looks beautiful and usage feels great. There is a lot of work here!

Changing to `npm run dev:electron` tripped me up at first! Can we update the instructions on starting in development and the README, (e.g. Content will only run in electron: `npm run electron:dev` and `npm run [platform]:dev`) so it matches?

The filter on stages is a little odd. If I type, it sort of updates, as in some of the items move higher as I type "socio", but obvious non-matches still show. If I type something with no matches, like "sociop", everything disappears. Deleting the "p" then shows only the sociogram stages as matches, like I would expect.

Typo on this line of text for the Device Name: "This is the name that your device will appear as when paring with Server." Pairing needs an "i" after the "a".

Clicking on Data Export Options give me a warning on console:
```
Warning: Failed prop type: The prop `exportFormat` is marked as required in `ExportOptions`, but its value is `undefined`.
```

Trying to enable "Run Fullscreen" gives me an error:
```
TypeError: electronWindow.isFullScreen is not a function
```

Expanding the categorical bin gives this warning on console:
```
Warning: Failed prop type: Invalid prop `expandedBinIndex` of type `number` supplied to `CategoricalBin`, expected `string`.
```

I think this is in master, but just to be thorough, when trying to type in the Filter field for sessions list gives me this error and the screen goes blank until I refresh:
```
```

jthrilly:
> Changing to `npm run dev:electron` tripped me up at first! Can we update the instructions on starting in development and the README, (e.g. Content will only run in electron: `npm run electron:dev` and `npm run [platform]:dev`) so it matches?

Great catch, sorry. I did that just for my own sanity and honestly meant to change it back. 



> The filter on stages is a little odd. If I type, it sort of updates, as in some of the items move higher as I type "socio", but obvious non-matches still show. If I type something with no matches, like "sociop", everything disappears. Deleting the "p" then shows only the sociogram stages as matches, like I would expect.

That sounds like a bug, although I'm not sure why it could be happening! I'll look into it.

> Typo on this line of text for the Device Name: "This is the name that your device will appear as when paring with Server." Pairing needs an "i" after the "a".

Probably a good time to add the spellcheck module and an npm task from the documentation type to help catch things like this!



As for the rest, I will definitely look at those. Thanks for a very thorough review Rebecca! Much appreciated.

jthrilly:
> The filter on stages is a little odd. If I type, it sort of updates, as in some of the items move higher as I type "socio", but obvious non-matches still show. If I type something with no matches, like "sociop", everything disappears. Deleting the "p" then shows only the sociogram stages as matches, like I would expect.

Turns out this was caused by a last minute decision to update the framer-motion library to the latest version, which apparently broke the exit animation!

jthrilly:
Okay, implemented a fix for the stage filtering, as well as a new test that covers stages menu interactions, including filtering.

jthrilly:
> Typo on this line of text for the Device Name: "This is the name that your device will appear as when paring with Server." Pairing needs an "i" after the "a".

I implemented an eslint spellcheck plugin to hopefully catch these in the future. Ironically, this isn't actually a spelling error, as "paring" is apparently a word.

jthrilly:
> I think this is in master, but just to be thorough, when trying to type in the Filter field for sessions list gives me this error and the screen goes blank until I refresh:

@rebeccamadsen  - Hmm. I'm not seeing that. Do you mean on the start screen? I don't get that at all, so can you create an issue for it?

PR #1023, opened 2020-02-11
Fix master failing tests

wwqrd:
Some failing linting and tests were not being picked up by travis.

This fixes the travis configuration, and the outstanding tests.


.travis.yml
```diff
@@ -27,6 +27,6 @@ jobs:
         npm run sass-lint -- --max-warnings 0 &&
         npm run test &&
         npm run build
-      script: npm run build:ci &&
+    - script: npm run build:ci &&
```
jthrilly: 👀 

So, what's happening with this then? Did I accidentally revert this?

src/containers/Interfaces/__tests__/CategoricalBin.test.js
```diff
@@ -15,7 +15,7 @@ describe('CategoricalBin', () => {
   it('renders CategoricalBin interface', () => {
     const component = shallow(<CategoricalBin {...requiredProps} />);
     expect(component.find('.categorical-bin-interface')).toHaveLength(1);
-    expect(component.find('.categorical-bin-interface').children()).toHaveLength(4);
+    expect(component.find('.categorical-bin-interface').children()).toHaveLength(3);
```
jthrilly: I have _no idea_ why this was ever 4. Can't see anything in the commit history about there being another child element in this component. Why did it ever pass!?

Side note: I don't feel bad about removing tests for things like "renders correctly", since we can move these to the integration tests now, which will be more meaningful.

config/webpack.config.prod.js
```diff
@@ -318,8 +318,5 @@ module.exports = {
     tls: 'empty',
     child_process: 'empty',
   },
-  performance: {
-    maxAssetSize: 6000000,
-    maxEntrypointSize: 6000000,
-  },
+  performance: { hints: false },
```
jthrilly: Again with this one: I got an error on the previous commit that the combined bundle size was over 5mb so I disabled this, only to find on the next run that the bundle size was way less. 

What is going on?

.travis.yml
wwqrd: Not a big fan of pinning the blame on people - but in this case I think it was me!

src/containers/Interfaces/__tests__/CategoricalBin.test.js
wwqrd: I changed this one to a snapshot test since I think that better captures the essence of what this test is for. 

I think the advantage of these tests over the integration tests is just that they run that much faster, which is useful for local development - but you are right we could remove it.

PR #1030, opened 2020-03-28
Export GraphML and CSV data v2.0

### LINKED ISSUE #897: Implement bulk file export
Bulk interview export is only implemented with export to Server. This should be extended to creating a zip file or multiple single files containing graphML or CSV.

Prerequisites:

- Shared GraphML export code #796

rebeccamadsen:
Resolves #897. This allows bulk graphml export from the existing sessions list as a zip file containing each selected session's graphml, or as a single graphml if only one session is selected.


src/utils/SaveFile.js
```diff
@@ -1,79 +1,236 @@
...
+const saveZippedSessionsElectron = (exportedSessionIds, sessions, installedProtocols) => {
+  // eslint-disable-next-line global-require
+  const electron = require('electron');
+  const fs = window.require('fs');
+  const { dialog, shell } = window.require('electron').remote;
+  const tempPath = (electron.app || electron.remote.app).getPath('temp');
+
+  const defaultFileName = 'networkCanvasExport.zip';
+
+  return new Promise((resolve, reject) => {
```
wwqrd, line 156: I'd opt for resolving this earlier and leaning on the promises more for readability e.g.:

```
// The dialog method only needs to capture the filename, then it can resolve
const saveDialog = () => 
  new Promise((resolve, reject) => {
    dialog.showSaveDialog({
      filters: [{ name: 'zip', extensions: ['zip'] }],
      defaultPath: defaultFileName,
    // eslint-disable-next-line consistent-return
    }, (filename) => {
      if (filename === undefined) resolve({ cancelled: true });

      resolve({ cancelled: false, filename });
    });
  });

// Then we can start to compose the promise from parts with understandable names:
const exportFiles = (exportedSessionIds, installedProtocols, tempPath) => {
  const promisedExports = exportedSessionIds.map((sessionId) => {
    const session = sessions[sessionId];
    const sessionCodebook = installedProtocols[session.protocolUID].codebook;
    const dataFileName = path.join(tempPath, `${session.caseId}_${sessionId}.graphml`);

    // export one file at a time
    return exportElectronFile(fs, session, sessionCodebook, dataFileName);
  });

  return Promise.all(promisedExports);
};

const exportArchive = filename =>
  exportFiles()
    .then(sourceFiles =>
      // by nesting when you need to keep things in context, rather than breaking out into let
      // variables we can make these promises really composable...
      archive(sourceFiles, filename))
        .then(() => sourceFiles.map(dataFileName => tryUnlink(dataFileName, fs)),
    );


// The finally function is much easier to follow (and hopefully the constituent parts too!):
saveDialog()
  .then(({ cancelled, filename }) => {
    if (cancelled) { return; }

    return exportArchive(filename)
      .then(() => shell.showItemInFolder(filename));
  });
```

src/utils/SaveFile.js
```diff
@@ -1,79 +1,236 @@
...
+    const { dialog, shell } = window.require('electron').remote;
+
+    return new Promise((resolve, reject) => {
+      dialog.showSaveDialog({
+        filters: [{ name: 'graphml', extensions: ['graphml'] }],
+        defaultPath: `${session.caseId}_${sessionId}.graphml`,
+      // eslint-disable-next-line consistent-return
+      }, (filename) => {
+        if (filename === undefined) resolve('cancelled');
+        exportElectronFile(fs, session, sessionCodebook, filename)
```
wwqrd, line 218: Do we need to pass around things like file system objects?:
- They are going to different functions per platform, so if those functions have different requirements, mirroring the interfaces seems a moot point?
- In the case of electron, it looks like `archive()` already provides it's own version of `fs`? Perhaps the same goes for cordova?

That said, I'm supprised to see seperate functions here but for platform requirements to be also handled elsewhere (in `archive()`)? My preference would be for platform checks as close to "the metal" as possible, with identical interfaces so the end result might look like: `exportFile().then(shareResult);`.

Granted that might not be practicable here because of the constraints of Cordova's API?

src/utils/archive.js
```diff
@@ -0,0 +1,116 @@
...
+  );
+});
+
+/**
+ * Write a bundled (zip) from source files
+ * @param {string} destinationPath full FS path to write
+ * @param {string[]} sourcePaths
+ * @return Returns a promise that resolves to (sourcePath, destinationPath)
+ */
+const archiveCordova = (sourcePaths, targetFileName, fileWriter, filesystem) => {
```
wwqrd, line 71: To me this looks like it might run more efficiently if we start with the zip promise, and then write the files one at time as we read them. Rather than 1. Read _all_ the files into memory 2. Write all the files from memory?

I'm not sure about how zip works internally, so that may not be the case?

src/utils/SaveFile.js
rebeccamadsen: `exportFile()` would be ideal, for sure. I started with what was working in electron on Server, which wouldn't work for Cordova, from the save as dialog to filesystem to archiver. But perhaps more of the middle could be shared? I'll take a look.

src/utils/SaveFile.js
rebeccamadsen: Good call, thanks!

src/utils/SaveFile.js
```diff
@@ -1,79 +1,236 @@
...
+
+const saveZippedSessionsCordova = (exportedSessionIds, sessions, installedProtocols) => {
+  const defaultFileName = 'networkCanvasExport.zip';
+  const fileType = 'text/xml';
+  const shareOptions = { message: 'Your network canvas zip file.', subject: 'network canvas export' };
+
+  let fileSystem;
+  let dataFilenames;
+  let fileEntries;
+  return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
rebeccamadsen: Thanks.

src/utils/archive.js
rebeccamadsen: I can see if I can reverse this, thought I think it is reading one file and zipping it before doing the next.

jthrilly: We need to be mindful of memory presure on cordova too. In general I would take any performance hit that kept memory use low.

wwqrd: @rebeccamadsen Hmm you're right, this is writing the files to the zip object one at a time. Looks like since we can't use streams the zip itself has to be generated as a single blob.

rebeccamadsen: It looks like there is an option for [`generateAsync`](https://stuk.github.io/jszip/documentation/api_jszip/generate_async.html) to 'streamFiles', but that some programs may not be able to read the zip file. I haven't tried it yet because that sounded worse to me, but maybe we can revisit if cordova memory is a bigger issue than some programs not being able to read the zip? I have it using the highest compression, so hopefully that helps.

Electron does use streaming, using the same method as Server.

jthrilly: @rebeccamadsen - what did you decide on this? I looked into the compatibility of generateAsync, and I think the tradeoff is worth it. All native ZIP apps on the major OS's support it. We need the memory reduction on cordova for scenarios where there are many sessions to export.

jthrilly:
I just wanted to say that because of some urgent work required for a contact tracing project, I am going to be moving over to look at this with you. 

The main priority for me is going to be CSV export. It shouldn't impact your work, so please carry on as normal.

jthrilly:
I have ported most of this code into `network-exporters`, and have zipped GraphML and CSV export working in NC on electron.

jthrilly:
Removed export from finish interview screen, since it will have its own flow when the new start screen drops.

PR #1036, opened 2020-04-28
Feature/better setup screen

### LINKED ISSUE #1029: Refactor start screen
There has been much feedback that the current start screen is not ideal for users. The plan for this is to replace the current start screen with a sortable and filterable `CardList` with the default sort option being by most recently installed.

### LINKED ISSUE #1034: sessions list filter error
When typing in the filter field for the sessions list on the start screen, sometimes this error occurs:

```
(anonymous function)
C:/Users/Big Baloo/Documents/GitHub/NetworkCanvas/src/components/FilterableListWrapper.js:56
  53 |   const nodeDetails = Object.values(node[entityAttributesProperty]);
  54 |   // Include in filtered list if any of the attribute property values
  55 |   // include the filter value
> 56 |   return nodeDetails.some(
     | ^  57 |     item => item.toString().toLowerCase().includes(filterValue),
  58 |   );
  59 | },
```

`item` is sometimes `null`, hence the error.

jthrilly:
Resolves #1034 
Resolves #1029 
Resolves https://github.com/codaco/Network-Canvas/issues/927


wwqrd:
What's the status on this? Is it ready to review?

PR #1047, opened 2020-05-18
check dirty state before warning

### LINKED ISSUE #1043: On EgoForm navigating back notifies about lost changes even if there are no changes

rebeccamadsen:
Fixes #1043. This skips the confirmation dialog when navigating backwards from a "clean" ego form.


jthrilly:
Could this also handle the per alter form, and the per alter edge form cases?

wwqrd:
I think the dirty check and dialog should probably go in a named function?

```js
const confirmIfChanged = () => {
  if (!isFormDirty()) { return Promise.resolve(true); }
  return openDialog(confirmDialog);
};

...

confirmIfChanged()
  .then(handleConfirmBack);
```

rebeccamadsen:
> See [#1047 (comment)](https://github.com/codaco/Network-Canvas/pull/1047#issuecomment-634063451)

Thanks for the suggestion!

PR #1048, opened 2020-05-21
Dyad Census Interface

### LINKED ISSUE #1033: [META] Dyad Census Interface
We need to create a dyad census interface.

General plan:

- Iterate through alter list.
- While holding current alter constant, show card for each _other_ aler with simple confirm/deny for presence of edge

wwqrd:
New Dyad Census Interface

Screenshot 2020-05-22 at 13 15 07

This interface asks the participant to answer whether or not two nodes are connected by an edge type.

The interviewer can select one edge type per prompt, and the prompt is asked for every unique node pair (not accounting for directionality).

Resolves #1033

## Pending
- [x] Merge in protocol validation changes
- [x] Update development protocol to include example
- [x] Update Architect to support this interface type


wwqrd:
Test protocol: [DyadCensus.netcanvas.zip](https://github.com/codaco/Network-Canvas/files/4667920/DyadCensus.netcanvas.zip)

wwqrd:
Is missing preview image in timeline:
Screenshot 2020-05-22 at 13 51 18

rebeccamadsen:
I like the look and feel. There are a few places in the code with "TO DO" mentioned, which I assume is part of the draft in progress status.

I noticed some odd behavior with the prompts. If you answer questions up to the second prompt, the progress bar at the bottom resets the count but not the progress.

Also, if you then go back a question using the arrow, and then continue answering questions, you get bumped back to prompt "friends" again after answering the "work together" questions, and have to use the stage arrows to move forward to the next stage.

wwqrd:
Nice thanks!

The progress bar should be fixed now.

> Also, if you then go back a question using the arrow, and then continue answering questions, you get bumped back to prompt "friends" again after answering the "work together" questions, and have to use the stage arrows to move forward to the next stage.

Please could you clarify this one for me? Either I'm not able to reproduce the issue or have a different idea of how this should work? It's based off the flow from the Alter Form,

rebeccamadsen:
Progress bar fixes works great!

The prompt interaction just gets a little funky when you try going backwards and across prompt boundaries instead of just staying within a single prompt. I don't think anything would seem weird at all if there were only one prompt in the stage.

One difference between Alter Form and this is that Alter Form (and Ego, and Alter Edge) don't have multiple prompts.

I'll try a video in slack to show one example.

wwqrd:
I'm wondering if the progress bar should count for both prompts e.g. you are only considered "half way through" after completing 1 of 2 prompts.

The progress tracker (in state) is intended to record the total progress of the interface (e.g. selected answers), and there is no way to 'skip' a pair or a prompt (completed or not). Equally, you must click 'next' through any pairs you clicked 'back' through.

Given all that it might be more clear to the user if the progress bar is for all pairs accross all prompts. The confusing issue then would be that the prompts are counted in the timeline progress bar - but maybe not a bad thing?

wwqrd:
@jthrilly feedback:

- [ ] we do need to track when a user has already answered "no", and not reset this when the stage is revisited. 
- [ ] Progress calculation needs to be within prompt.
- [ ] when the prompt changes, there needs to be more visual indication. I think the option panel and the pair section should hide and then show again.
- [ ] When an answer is given in the option panel, we need more feedback. As you choose an option, both the pair AND the options "slide up" and are replaced by the next pair. 
- [ ] We need feedback if the user hasn't answered the question but tries to click next. I added an invalid modifier to the choice component CSS, and a "shake" keyframe animation for attention.

PR #1049, opened 2020-05-27
Disable .session-panel clicks for Architect preview

wwqrd:
This adds back in menu click prevention for Network Canvas when the app is viewed in preview mode in architect.

It also removes the old click prevention which is now defunct.


jthrilly:
I could be mistaken, but I think this should also disable the menu button. Is that disabled elsewhere? It is probably okay to remove with `display:none`/`visibility:hidden` rather than simply removing pointer events, since I think it might confuse users to have it still be visible.

PR #1053, opened 2020-06-01
Feature/session info

### LINKED ISSUE #1045: change case ID option
It would be great to be able to change the case ID after creating it, in case the user needs that option. Perhaps as a menu option.

### LINKED ISSUE #1046: show session information in menu
We should show relevant session information in the menu, so interviewers have context at their fingertips. A partial list: protocol, session duration, case ID, etc.

rebeccamadsen:
Adds session information to the settings menu. Right now, this includes case id, protocol name, session duration, date of last export, and session stage/prompt.

It also allows case id to be changed on this view.

Duration is simply the difference between "now" and the creation of the session. We can flesh this out differently, though I wasn't sure how complicated we want this right now.

Resolves #1046. Resolves #1045.


jthrilly:
I like this a lot! I modified it to move it out of the settings menu and into the session navigation sub menu (that way it is only ever seen with pressing the menu button from within a session - more immediate, and avoids having to check for an active session).

I also removed session progress as the timeline shows it, and added network summaries instead.

Screen Shot 2020-06-02 at 5 18 48 PM

jthrilly:
I'll update the interview duration to be realtime in the new start screen branch, as I have a `useInterval` hook in there already.

PR #1054, opened 2020-06-01
fix useSteps for dyadCensus

rebeccamadsen:
This is a fix for an error on the DyadCensus stage. After resetting App Data, I saw this error on the DyadCensus stage of the development protocol: `TypeError: Cannot destructure property 'step' of 'undefined' as it is undefined`. This seems to fix it for me, unless someone has a better solution.

### DIFF
--- a/src/containers/Interfaces/DyadCensus/DyadCensus.js
+++ b/src/containers/Interfaces/DyadCensus/DyadCensus.js
@@ -233,7 +233,7 @@ const DyadCensus = ({
                     animate="show"
                   >
                     <div className="dyad-interface__progress">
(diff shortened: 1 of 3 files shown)


jthrilly:
I'm not sure if this is the correct fix, as I find the intent of this a bit hard to follow (looks to be stuck between two designs?).

One thing I noticed is the progress bar isn't working, so this should be fixed here too. I also think an integration test is needed just to cover this stage loading correctly and functioning with some visual snapshots.

One super small thing is that the edge seems to have a visual glitch (probably to do with rounding) so that you can still see a very small portion of it when it is supposed to be in the hidden state. It can be fixed by increasing the negative margin, increasing the width, and setting a negative (or lower) z-index.

PR #1065, opened 2020-07-30
implement mock session generation

jthrilly:
This PR ports over the mock session generation from Server to NC. It is implemented within the developer options settings menu.


jthrilly:
>     1. This seems to leave the last created session "open" - if a session was already in progress this can seem especially weird.

Yeah, this is because we use the `ADD_SESSION` action to set the active session, which is a bit silly. I went to remove it before I realised that the preview mode uses it. I thought that given this is a developer facing feature it didn't matter too much.

> Coding style seems a bit devious, e.g. [...Array(sessionCount)].forEach, but elsewhere times()

Yeah :/ I hadn't seen `times` before, and in general I try and avoid lodash stuff if I can help it. In this instance, the `Array()` syntax is coming from Server, where it is advantageous from a performance perspective to create an empty array of elements that you then populate as you map over. Here, we aren't mapping so it doesn't make sense, I agree.

PR #1068, opened 2020-08-10
Release for 5.2.0

jthrilly:
This release is designed to change update behaviour so that we can handle complicated upgrade paths.

- [ ] Implements CodePush for cordova updates
- [ ] Requires specific version of Server <-> NC comms API
- [ ] Removes automatic update checking on startup
- [ ] Implements a rich upgrade dialog on Electron


jthrilly:
This work has been abandoned in favor of a different approach.

PR #1069, opened 2020-08-11
Use embedded prior versions within Network Canvas

jthrilly:
This is an initial attempt at implementing embedding prior versions of the app, for backward compatibility.

Outstanding issues:
- [x] Submodule NC
- [ ] Npm task that compiles submodules for platform and integrates with existing build process
- [ ] Update `core/5.1.0` of Network Canvas to function in required ways (exit interview).
- [ ] Work out how shared state should be managed


jthrilly:
Work on this has been cancelled.

Issue #1066, opened 2020-07-30
Performance degrates substantially as redux store gets larger


jthrilly:
The performance of the app degrades substantially as the redux store grows.

To reproduce: 
- install the development protocol
- generate 10 sample sessions

Observations:
- Dragging on the sociogram (which emits a debounced action) becomes unusable. This issue has already been identified (#1031)
- App becomes generally much more sluggish

wwqrd:
@jthrilly Is this task still outstanding? I see that there was a merge against changing to a different storage backend, does that resolve this issue?

jthrilly:
No, the data storage engine is one piece of this. This issue might be better re-phrased as "app performance generally degrades as the redux store grows".

The underlying speed of the storage system only papers over the cracks in this case (and actually, I'm not certain how the performance of sqlite through a cordova plugin compares with the web view's local storage implementation - we just needed to do that for other reasons). 

There is another issue relevant to this here: https://github.com/complexdatacollective/Interviewer/issues/1031 (drag and drop performance hitches, which worsened as redux store grew). 

I was able to resolve the above issue satisfactorily by correctly throttling edge update events, but once again the _underlying_ issue is that interacting with large (and perhaps specifically deeply nested) objects in javascript is quite slow.

You may recall that we discussed this on slack a while back. Our idea was to looking having separate stores. One for the app and its settings, one for protocols, and one for each session. I think looking into this is the only real solution to these issues.

wwqrd:
So essentially every state change is causing a `JSON.stringify()` because it's persisted. I can see why that would slow things down! One immediate improvement that could be made to that is to not persist most of the stage interaction events, maybe "change stage" only.

In addition to separate redux stores, anything but the active protocol/session could be kept on disk and out of state.

jthrilly:
> One immediate improvement that could be made to that is to not persist most of the stage interaction events, maybe "change stage" only.

Unfortunately, this would lead to a lot of data loss potential, so I don't think we can do it. The best solution is to increase the performance of the app following the example in #1031, and to separate out the stores so that each interview session always starts from a baseline performance, and only degrades according to its own contents.

jthrilly:
On Sociogram performance specifically: this problem can be resolved by porting over the approach used on the force directed layout mode (`RealtimeCanvas`). This bypasses the react render cycle, and only emits actions at specific key moments (node layout completes, drag ends, etc.). Performance here is a smooth 60fps.

On the issue more broadly:

[...]

PR #1071, opened 2020-08-14
Version 6.0.0

### LINKED ISSUE #1029: Refactor start screen

### LINKED ISSUE #1031: Drag and Drop Performance issues
In the development protocol on the CSV name generator, when starting a to drag a node there is a console warning about the event handler duration. It only fires when the drag first starts, and there is a very visible hitch from the user perspective at the same time. It kind of ruins the 'feel' of dragging.

It seems like the issue is that when a drag start is detected, a redux action is dispatched (`DRAG_START`) which takes 150ms to execute. This only fires once, and right at the start, but it is enough to hitch the animation and ruin the UX. Temporarily commenting this out makes performance buttery smooth.

We believe this is causing some unnecessary re-rendering. To investigate further we need to to rendering profiling of the component tree.

### LINKED ISSUE #1034: sessions list filter error

### LINKED ISSUE #1055: Update react-redux for hooks support
Version 7 of react-redux includes hooks for state and action dispatches, which would allow us to use custom React hooks without needed to pass dispatch/state as arguments. e.g.

```jsx
const useCustomHook = (dispatch, someStateFromStore) => {
  const doThing =  (v) => dispatch({ type: 'DO_THING', payload: v });
  const modifiedValue = someStateFromStore * 10;
  
  return [modifiedValue, doThing];
};

const MyComponent = ({ dispatch, someStateFromStore }) => {
  const [value, doThing] = useCustomHook(dispatch, someStateFromStore);

  return <div>{value}<div>;
};

connect(mapStateToProps)(MyComponent);
```

Could simplified to:

```jsx
const useCustomHook = () => {
  const someStateFromStore = useSelector(selectorReturningObject, shallowEqual);
  const doThing =  (v) => dispatch({ type: 'DO_THING', payload: v });
  const modifiedValue = someStateFromStore * 10;
  
  return [modifiedValue, doThing];
};

// Much tidier component:
const MyComponent = () => {
  const [value, doThing] = useCustomHook();

  return <div>{value}<div>;
};
```

As of writing best candidate would be:
react-redux 7.2

This would require:
react 16.8.3 or later

We could also update redux to version 4.x.x although this is not a requirement.

### LINKED ISSUE #1067: ready to upload to server message incomplete
The message on the "ready to export" dialog trails off as incomplete. My screen shows "Ready to export 1 session to ." It's expecting the paired server to have a `name`, but I don't have a name for my paired server. Should we be using the ip address instead? Or is the paired server supposed to have a name?

uploadmessage

jthrilly:
Obviously this PR is pretty substantial, and I don't expect that it can be reviewed in minute detail. Instead I would like you to focus on the core functionality, and to check that the essential functions of the application are working for you. We will deal with bugs in future PRs.

## Changes:

- Introduces substantially redesigned card-based start screen (Resolves #1029) including sections for common actions. All sections should be sensitive to app state and device state, meaning that they will contextually show themselves when their action is possible, and will disable subsets of their actions according to device network state/server pairing state.
- Adds a "preloading" animation that lets the app transition into the start screen view nicely on first launch.
- Adds a "welcome section" that can be optionally shown/hidden (default is shown on first launch) containing links to documentation.
- In preparation for stable, bumps the app version to 6.0.0 all app bundles and local storage keys to a new identifier.
- Removes compatibility with schema versions prior to 4, since this will be the launch schema version.
- A new `useServerConnectionStatus` hook has been written, which allows components to poll for heartbeat messages from server (not currently implemented).
- The flow for pairing with Server and fetching protocols has been redesigned and largely re-written on the component level. Pairing takes places from a dedicated Server communication card, and fetching protocols happens from a dedicated protocol import/management card. Resolves #1067. 
- A toast UI system has been written and implemented for protocol import, protocol export, and server pairing success.
- The display of protocol cards, session cards, and cards representing a computer running Server have been standardised within NC-UI and implemented here.
- Refactors import and export behaviour to not use redux state, but instead to dispatch toast actions.
- Redesigns the finish interview screen to remove delete session and export options. Includes a `finishedAt` session property, as well as in-session duration information.
- Fixes incorrect windows file menu items.
- Implements the ability to have custom validation functions passed to a field and through a form config object. These validation functions can contain custom reject messages.
- Removes obsolete cruft related to react-router, and implements latest libraries. Fixes #1055 
- Updates redux and redux-persist.
- Changes the app frame on windows to show the menu bar, and changes the startup behaviour to windowed. Full screen option is still available, but ive seen too many people be confused by it.
- Resolves #1034 
- Resolves #1031 (and helps with #1066) by successfully throttling dispatch of actions that update node position while dragging
- Changes drag manager portalling to the body element to help with z-index issues.
- Removes the `selectable` behaaviour in favour of components binding their own events and managing props of children.
- Implements a new filtered list wrapper for start screen items, and preserves the old one for use in the interview.
- Implements an `<ExternalLink />` component and an `openExternalLink` method to allow documentation links to be opened from within the app.
- Removes the development protocol as a submodule. As far as I could tell it wasn't being used.


jthrilly:
- [ ] Pressing "tab" while the toggle show welcome message switch is active makes the panel vanish!

Kate's feedback:

- [ ] Don't show "compelete" toast when export was cancelled

rebeccamadsen:
Looks pretty great! I can add one more observation. The date field does some funky things with aligning here, which isn't in master.

date

jthrilly:
Update:

- I can't see anything wrong with the server protocol fetch dialog that would cause that wonky rendering, but I saw it a single time too. I've updated framer-motion, and I haven't seen it yet so we will just have to see if we can get a firm reproduction.
- Date picker bug was resolved in UI
- I'm going to leave the "complete" toast for now - I think its consistent with the pattern.
- I can't fix the tab issue. Leaving it for now

jthrilly:
Outstanding iOS issues:

- [ ] Server discovery is broken
- [ ] New case form doesn't let you type into it
- [ ] Timeline navigation invisible
- [ ] Ego form progress bar too close to bottom of the screen
- [ ] Import from file doesn't trigger file chooser
- [ ] Forms push content off screen (regression)
- [ ] Header background gradient renders incorrectly

wwqrd (approved):

PR #1077, opened 2020-09-26
make sure beforeNext runs on enter pressed

### LINKED ISSUE #1025: alter form interface sometimes has "extra" nodes
From user feedback in slack:

Sometimes the alter form interface shows screens for "10 of 3" or "4 of 2" nodes. See these screenshots:
Screenshot_20200214-135658_NetworkCanvas

Screenshot_20200214-141023_NetworkCanvas

It seems like this could be related to node filtering on the stage level, but checking values for the rules and predicates within networkQuery looked as expected and we have not successfully reproduced this yet.

The user also mentions issues with data not showing in the form if they navigate backwards in the interview, however it is not clear whether this was only for the nodes that shouldn't even be showing on the alter form stage.

Placing here so we can check into this further. Protocol is in slack.

rebeccamadsen:
Resolves #1025.

Steps to reproduce the issue:
0. create a protocol with a name generator and an alter form that has a text field.
1. add protocol and start a new interview in NC.
2. add a node.
3. navigate to the alter form.
4. Click next arrow to get past the intro to first node on the alter form.
5. Mouse in to the text box and hit enter on the keyboard. Will not move to the next stage and says "2 of 1" at the bottom progress bar.

Basically, pressing the enter key in a form uses the redux hidden submit button to submit the form, bypassing all of our `beforeNext()` logic. The `beforeNext()` logic checks to make sure it's okay to submit a form, and whether we move to the next stage or simply the next node in the current stage. So this makes sure that logic actually still happens before submitting for both Alter and Alter edge form.

### DIFF
--- a/src/containers/ProtocolScreen.js
+++ b/src/containers/ProtocolScreen.js
@@ -36,8 +36,8 @@ class Protocol extends Component {
     this.beforeNext = {};
   }
 
-  onComplete = () => {
-    const pendingDirection = this.state.pendingDirection;
+  onComplete = (directionOverride) => {
+    const pendingDirection = directionOverride || this.state.pendingDirection;
     const pendingStage = this.state.pendingStage;
 
     const navigate = (pendingStage === -1) ?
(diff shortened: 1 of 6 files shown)


wwqrd:
Looks great to me!

Out of scope: seeing the propagation of the submit button makes me think that this behaviour needs revising at some point?

wwqrd (approved):

PR #1089, opened 2020-10-29
only allow boolean additional attributes

### LINKED ISSUE #1079: Dragging a node to the side panel should remove any additional attributes
As per the title, if a node is created on one stage, dragged from a side panel into the main panel on a second stage, and then dragged back into the side panel on that same stage, any attributes from the second stage should be removed or at least reset.

rebeccamadsen:
Fixes #1079. ~~This uses the value on prompt load for all nodes in order to revert any additional attributes for the prompt.~~

~~Note: This means changes to a node while still on that prompt will be reverted to initial values, but dragging a node over on any subsequent visits to that prompt will not change the additional attributes. This is different from current functionality, and likely future functionality.~~

Adjusted to only allow boolean additional attributes (enforced by protocol-validation) and to simply toggle the additional attribute value when a node is removed from a prompt.

### DIFF
--- a/src/containers/NodePanels.js
+++ b/src/containers/NodePanels.js
@@ -2,8 +2,9 @@ import React, { PureComponent } from 'react';
 import { compose, bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
+import { find } from 'lodash';
 import { getCSSVariableAsString } from '@codaco/ui/lib/utils/CSSVariables';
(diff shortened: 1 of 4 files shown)


rebeccamadsen:
This is ready for another look.

jthrilly (approved):
Assuming the changes in https://github.com/complexdatacollective/protocol-validation/pull/43#pullrequestreview-527208566 are made, this looks good to merge.

jthrilly (approved):

wwqrd:
Closing because github.

PR #1091, opened 2020-11-04
Feature/updater

jthrilly:
This PR implements a new update system, using toasts rather than `electron-updater`.

Essentially, the `useUpdater` function calls the github releases endpoint, and compares the version of the latest release with the version of the current app. If there is a newer version available, it shows a dialog that uses `react-markdown` to render the body of the release, along with a platform specific link to download and install. This markdown area also supports rendering links that open correctly from the app. Images are not tested.

The link points to the Appple App store on iOS, the Google Play Store on Android. On electron it uses the platform and the assets attatched to the release to find either a DMG file or an EXE, and points directly to the browser download URL.

The toast itself is dismissable either in a single instance (by pressing the 'x' button) or for this release by using a dedicated "dismiss for this release". This button stores dismissed app versions in state.


rebeccamadsen:
This works on windows as far as I can tell, but I have a question. Didn't the previous approach to installing close any active open NC apps when the newer version installed? I could be remembering wrong. I just noticed that with this approach I ended up with two open NC's -- one of the "old" version and one of the newer -- which seems a bit odd. Could also be a testing artifact since I downgraded my local version to force it to "see" an update.

jthrilly:
> This works on windows as far as I can tell, but I have a question. Didn't the previous approach to installing close any active open NC apps when the newer version installed? I could be remembering wrong. I just noticed that with this approach I ended up with two open NC's -- one of the "old" version and one of the newer -- which seems a bit odd. Could also be a testing artifact since I downgraded my local version to force it to "see" an update.

I think this could be because the apps are technically separate now. Anything prior to 6.0.0 is considered a different app, so the installer won't close it.

rebeccamadsen (approved):
+1

rebeccamadsen (approved):

PR #1095, opened 2020-11-09
Feature/cordova storage

### LINKED ISSUE #1072: Switch to proper robust data storage implementation
We need to use appropriate storage adaptors for `redux-persist` because localStorage can be erased **at any time** by the OS, and has varied limits across platforms: https://cordova.apache.org/docs/en/latest/cordova/storage/storage.html

This will lead to sudden spontaneous data loss.

It is also very slow (serialize/deserialize) and has limits on what it can represent. 

Existing adapter options are listed on the redux-persist repo: https://github.com/rt2zz/redux-persist

It looks like an optimal solution would be to find a way to use SQL lite across electron and cordova. This may require writing our own plugin for cordova. Cordova plugin is here: https://github.com/storesafe/cordova-sqlite-storage

Potential plugin already developed here: https://github.com/sumedh22/redux-persist-cordova-sqlite#readme

IndexedDB, localStorage, localForage, and web SQL are all **not** appropriate, since they are also susceptible to OS deletion on iOS. 

This relates to #1066, and will ideally contribute to solving it. For example, we might choose to put each session in its own SQL lite database.

jthrilly:
Fixes #1072

Turned out to be quite a pain! I ended up writing my own storage adapter for `redux-persist` around the `cordova-sqlite-storage` plugin (which is...interestingly maintained).

If you have problems getting it to run it might be to do with the plugin dependencies. Make sure you npm install, and then do `cordova prepare`. Failing that, try running `cordova platform remove android`, and then `cordova prepare`.

If you still can't run it, just provide some feedback on the implementation (I've tested on all platforms so I am confident it works on cordova).

Todo: 
- [x] Implement normal storage or a different storage engine for Electron
- [x] Implement reset data method
- [x] Test on iOS
- [x] Test for large datasets


wwqrd (approved):
We took at look at this together with particular focus on the circular reference in `store.js`.

Some suggested fixes included:
1. Moving the database initialisation outside of `store.js`
2. Making `store.js` asynchronous

In the case of (1.) there was a preference for making the storage plugin self-contained. In the case of (2.) it would require changes throughout the app where store is treated as a global reference.

In conclusion we decided to stick with this working implementation as is.

PR #1098, opened 2020-11-11
implement codebook hash in sessionVariables

jthrilly:
Updates Interviewer to send include `codebookHash` in the session variable properties sent to `network-exporters` and to Server via the API.

Updates the network exporters submodule to encode this data correctly in graphml files.

### DIFF
--- a/src/styles/containers/StartScreen/_export-error-list.scss
+++ b/src/styles/containers/StartScreen/_export-error-list.scss
@@ -10,6 +10,7 @@
       margin: 0.75rem;
 
       .icon {
+        flex: 0 0 auto;
         height: unit(3);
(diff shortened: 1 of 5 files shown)


src/styles/containers/StartScreen/_export-error-list.scss
```diff
@@ -10,6 +10,7 @@
       margin: 0.75rem;
 
       .icon {
+        flex: 0 0 auto;
```
jthrilly, line 13: Fixes a small visual glitch where amount error text caused the icon to be smaller or larger

rebeccamadsen (approved):
Looks good to me!

PR #1099, opened 2020-11-23
implement health endpoint and update network-exporters implementation

jthrilly:
This PR updates the network-exporters submodule to match a slightly tweaked implementation from Server. It also points to the new `/health` endpoint.

I can't be sure exactly why, but this change required adding `archiver` to the externals section of the webpack config.


jthrilly:
Archiver might always have been broken in production builds, since the webpack config wasn't the same. Archiver also depends on async for queue, like the new network-exporters, so this was confusing things.

Strategy ended up being:

- Add `archiver` to externals
- Move archiver and fs require statements to be inside of `isElectron` codepaths, to prevent them from running on cordova

PR #1100, opened 2020-11-23
Stable Release (6.0.0)

jthrilly:
This PR hosts RC changes for the stable release (6.0.0)

Issues: 
- [x] Padding between server cards in discovery view
- [x] Remove About screen
- [x] Update UI for RadioGroup bug fix
----
- [x] Incorrect networkQuery behaviour relating to EXISTS/NOT_EXISTS
----
- [x] Server protocol import picker should close when a protocol is selected
- [x] replace remoteProtocolName with protocolUID for consistency with Server and update network-exporters
----
- [x] Add "select unexported" option to accompany "select all" option in export panel
- [x] Fix node preview for name gen quick add interface. Avoid flash of "no name available", make button submit, and make clicking background hide.
- [x] Remove scrollbar hidden from scrollable and use native platform scrollbar behaviour. Remove "show scrollbars" option.
- [x] Installing a protocol didn't set "last modified"
- [x] Remove installer DMG background
----
- [x] minValue validation allows '0' as a valid value. For example minValue 4 allows 0, but disallows 1,2,3.
----
- [x] iPad version of Interviewer shows "version mismatch" when paired with w10 version of Server
- [x] Dyad census incorrectly shows linking animation when moving between choices
- [x] When in full screen, the menu bar causes the app to push content off of the bottom (needs verification).
----
- [x] Sample protocol must be updated
- [x] Update preview thumbnails
- [x] Interviewer checks for an update when in preview mode. Is this a problem?
- [x] CardList ratio error has reappeared
--- 
- [x] Dyad census doesn't show when there are only two nodes in some protocols (dev protocol not impacted)


rebeccamadsen (approved):
This has been pretty well tested by now!

PR #1110, opened 2021-01-15
Update edge creation on Sociogram to allow cancel

### LINKED ISSUE #1105: Cannot cancel edge creation [Sociogram]
When you click on a node in sociogram, there is no way to cancel edge creation. You must create an edge and then delete it. 

Ideally clicking the node a second time would cancel edge creation.

wwqrd:
Resolves #1105


rebeccamadsen (approved):
Looks great!

PR #1111, opened 2021-01-18
Reset filter in DataExportSection

### LINKED ISSUE #1104: select options should remove filter on manage interview list
When using the "filter" on the manage interview data list, "select all" and "select un-exported" both still select all and un-exported, even if those aren't showing. These two select toggles should clear the filter. Otherwise, sessions may be selected that the user cannot see because of the filter.

For example, here four sessions are selected, but only one is visible:
image

Follow up question: should typing in the filter de-select the "select" toggles? Because the reverse can also be true if the user selects all, and then types in the filter.

wwqrd:
This implements a change to reset the filter box when the toggles are selected.

> Follow up question: should typing in the filter de-select the "select" toggles? Because the reverse can also be true if the user selects all, and then types in the filter.

Changed so that the selection is reset when the filter is changed (regardless of the state of the buttons). I initially thought this seemed drastic, but this use of the filter wrapper is quite unique in the way that multiple cards can be selected, and it's easy to get into a situation when the selected cards are not visible.

Resolves #1104


rebeccamadsen (commented):
This works pretty well. The one oddity was when I have "select all" enabled, and then type in the filter field, my first character is wiped away as the toggles are reset. This doesn't happen if I start typing with just a few selected.

jthrilly (approved):

PR #1118, opened 2021-02-17
Release branch for 6.0.1

jthrilly:
Changelog:

- Fixed an issue during installation where the user would have to cancel Bonjour installation even if they had indicated they did not want to install it.
- Added the ability to install the app on Linux using the snap store, as well as providing AppImage and deb packages.
- Fixed various console warnings and errors.
- Implemented the ability to cancel edge creation on the Sociogram interface by clicking/tapping the first selected node.
- Changed the behavior of the filter and select functions of the data export section to be more logical.


rebeccamadsen:
> Implemented the ability to cancel edge creation on the Sociogram interface by clicking/tapping the background.

Actually looks like canceling edges happens by clicking on the node, not the background?

jthrilly:
> Actually looks like canceling edges happens by clicking on the node, not the background?

Thanks for catching this!

PR #1121, opened 2021-03-09
allow markdown in field labels and options

rebeccamadsen:
Updates to include the latest in UI, markdown for field labels/options. Also updates Categorical Bin and Ordinal Bin to allow for markdown in options.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "@babel/preset-env": "7.10.4",
     "@babel/preset-react": "7.10.1",
     "@codaco/eslint-plugin-spellcheck": "0.0.14",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -212,10 +212,6 @@
     "nsis": {
       "include": "build-resources/scripts/bonjour.nsh"
     },
-    "win": {
```
jthrilly, line 215: What happened here...?

jthrilly (dismissed):
I'm going to revert the presumably unintentional windows build change, and then merge this.

package.json
rebeccamadsen: Doh! That was a mistake, for sure.

PR #1122, opened 2021-03-10
Update to npm7

wwqrd:
The aim of this PR is simply to upgrade to npm 7! However, npm 7 introduces strict peer dependency checking, so this PR also reviewed and updated those dependencies.

#### One of those dependencies affected the linting rules
Updated the code to follow the new rules, there were a lot of changes!

#### One of those dependencies affected `redux-observable`
Solved by removing this dependency and using `thunks` instead. Observable/rxjs is elegant, but the API has changed a lot between versions, and it is a difficult library to test with. We are using `thunks` elsewhere so this change streamlines things.

**Note: might be a good idea to update UI and secure api modules before merging this?**


.eslintrc.json
```diff
@@ -22,9 +22,11 @@
     "rules": {
       "react/prop-types": [0],
       "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
+      "react/jsx-props-no-spreading": "off",
       "react/forbid-prop-types": "off",
       "react/no-array-index-key": "off",
       "jsx-a11y/no-static-element-interactions": "off",
+      "jsx-a11y/click-events-have-key-events": "off",
```
jthrilly, line 29: This seems like a really bad one for accessibility, but I can't quite get my head around how to solve it properly. Do we need to create `onKeyDown` and `onKeyUp` handlers that mimic whatever the click does?

jthrilly (approved):
I think this is functional enough for anything else to be fixed in separate PRs.

PR #1126, opened 2021-03-31
add schema 5 validation support

rebeccamadsen:
This adds support for "unique", "differentFrom", and "sameAs" validation options and updates to schema version 5.

Questions about implementation:
1. `validation: { "unique": null }` was stripped in json, so I put an empty string for now instead of null so it wouldn't be stripped. Is there a better option for representing "unique"? We could put the variable uuid, but it isn't necessary because it's contained in the validation function. Other thoughts? 
2. What does "unique" mean for ego? I didn't implement this yet, but are we wanting to compare ego variables across sessions? This would only verify across the sessions currently in Interviewer, so I'm not sure it means anything?
3. Similarly, will other forms (e.g. SessionInformation, ServerAddressForm, ProtocolUrlForm, and NewSessionOverlay) need validation options?
4. The error message for "differentFrom" and "sameAs" for simplicity here notes the value, because it is handy in the validation function. Should this be a different piece of information? Variable name (could be meaningless to the user)? Variable label (could be wordy)?

~To do item: add tests for the new validation options.~


jthrilly:
>     1. `validation: { "unique": null }` was stripped in json, so I put an empty string for now instead of null so it wouldn't be stripped. Is there a better option for representing "unique"? We could put the variable uuid, but it isn't necessary because it's contained in the validation function. Other thoughts?

I think you should follow the pattern we used for required, which is `required: true`.



>     2\. What does "unique" mean for ego? I didn't implement this yet, but are we wanting to compare ego variables across sessions? This would only verify across the sessions currently in Interviewer, so I'm not sure it means anything?

Yeah, unique doesn't apply to ego. We definitely don't want to think of this as cross-session. I'm not sure how unique is implemented, but I would imagine it works backward from a variable ID, to a subject, to the network filtered by that subject type. We should ensure this doesn't get applied to ego variables by updating the schema validation to prevent it in that case, and also obviously by not implementing it in Architect. Because of that, I don't think the implementation needs to gracefully handle use on an ego variable.



>     3\. Similarly, will other forms (e.g. SessionInformation, ServerAddressForm, ProtocolUrlForm, and NewSessionOverlay) need validation options?

Again, I've not seen your implementation yet, but these new validation options are all implicitly drawing on the network data model, so it isn't appropriate to use them for general purpose form validation as in the researcher facing UI.



>     4\. The error message for "differentFrom" and "sameAs" for simplicity here notes the value, because it is handy in the validation function. Should this be a different piece of information? Variable name (could be meaningless to the user)? Variable label (could be wordy)?

In the case of differentFrom and sameAs I think its okay to reference the variable name,  even though it won't mean much to the participant.

jthrilly:
I implemented the changes I described above in the protocol validation repo :)

src/containers/Interfaces/Sociogram.js
```diff
@@ -99,6 +99,7 @@ const Sociogram = (props) => {
           <NodeBucket
             id="NODE_BUCKET"
             node={nextUnplacedNode}
+            allowPositioning={allowPositioning}
```
jthrilly, line 102: Just curious about what's happening here. Is this a bug that you've fixed?

rebeccamadsen: So, the NodeBucket wasn't showing up for me on Sociogram without this!

PR #1128, opened 2021-04-06
Feature/tie strength census

jthrilly:
Implements the Tie-Strength census interface.

Screen Shot 2021-04-09 at 11 13 23 AM

- The purpose of the interface is to provide a dyad census that also assigns a value to an ordinal variable. Clicking an ordinal value will either create an edge and then assign/update a value, or simply assign a value. The interface always generates a button representing "no relation ship", which the researcher can specify the label of. 
- It uses the new boolean input styles, and updates the Dyad Census interface to be visually the same
- Stage configuration details and boolean variable support have been added to schema 5, which this PR also implements
- This interface is our key deliverable for our ongoing collaboration work with the Perry lab.
- Other changes: 
  - Moved markdown allowed tag definitions to reference UI.
  - Updated use of ReactMarkdown to instead use MarkdownLabel where appropriate

## Testing

The development protocol has been updated with an example of this interface.

Todo:

- [x] Implement interface
- [x] Add tests
- [x] Add timeline icon
- [x] Update UI to include new Boolean input
- [x] Update DyadCensus to use Boolean


jthrilly:
Two bugs with this one:
- [ ] When in preview mode, if this is the last stage and there are no pairs to evaluate, clicking next will take you past the end of the interview (`no "" interface found` error)
- [ ] Kate reports that using this with an ordinal value that is 0 doesn't create an edge.

wwqrd:
> Above bug might also apply to DyadCensus - test and fix here if so.
Confirmed

PR #1130, opened 2021-04-15
Release branch for 6.1.0

jthrilly:
## Network Canvas Interviewer 6.1.0

This is the first significant feature update to Interviewer since our launch. These changes required **a new protocol schema version**, although all existing protocols will continue to function. Review the changelog below, and take extra care when updating if you are currently in the process of collecting data.

### Changelog:

- **Added a new interface: the "Tie-Strength Census"**. This interface follows the same paradigm as our dyad census, except it assigns a value to an ordinal variable if an edge is present. This can be used to capture both the strength and existence of a relationship in one interaction. It resulted from a collaboration with Dr. Brea Perry at the Indiana University Network Science Institute. 
- **Introduced a new input component: "BooleanChoice"**. This new component is available for boolean variable types, and differs from the existing toggle switch in two ways: (1) it requires user interaction to set a value (it has no default), and (2) it supports rich labels describing each 'choice'.
- **Introduced markdown support for field labels and option values**. This means that all field labels and option values for ordinal/categorical form variables can now contain limited rich text formatting, such as bold and italic text. A WYSIWYG editor has been added to Architect to make formatting these labels more straightforward.
- **Introduced new validation options: unique, differentFrom, and sameAs**.
  - **sameAs** - Requires that a variable value be the same as another variable value.
  - **differentFrom** - Requires that a variable value be different from another variable value.
  - **unique** - Requires that a variable value be unique among all entities of the same type.

# Focus for testing:

Please heavily test the new features! In particular, form interfaces and the new component, as well as the new interface! Look out for any visual oddities in existing components caused by changes to UI. Look out for situations where wider markdown support is creating issues.


rebeccamadsen:
- [ ] If I answer the "decline" option on a Tie-strength census, should we already "know" the answer is "no" for a Dyad census too? We know the "yes" answers.
- [ ] maxSelected validation expects interaction before it will show as valid

jthrilly:
>     * If I answer the "decline" option on a Tie-strength census, should we already "know" the answer is "no" for a Dyad census too? We know the "yes" answers.

No - these are separate interfaces, and may have separate prompts. We just want to track responses within a stage for now, even when they seem to deal with the same thing. Its also extremely unlikely someone would use both.

PR #1138, opened 2021-06-02
Roster Interface Overhaul

wwqrd:
Complete overhaul of the small and large roster interfaces.

- Both interfaces now look identical, with the following exceptions:
  - For large rosters, search parameters can be set (sort options are not shown)
  - For small rosters, sort options can be set, and search becomes exact-match (normally "fuzzy")
- If the list has over 100 items the behaviour is slightly different:
  - Roster is blurred out until the participant starts a search.
- supports reducedMotion

### Outstanding issues
- [x] Search bar styles do not match standard text input.


jthrilly:
Just going to dump feedback as I look at it:

- [ ] First run, and the loading spinner never goes away. I get:
```
Failed to load resource: net::ERR_FILE_NOT_FOUND
:3000/#/session/a491f36d-638f-47a7-a08a-c37f38481e5c/8:1 Uncaught (in promise) TypeError: Failed to fetch
```
I think we need error handling for this. Perhaps an error boundary?
- Reset app data and reinstall dev protocol and working great! 👍 Absolutely LOVE the design! You nailed it.
- [ ] How about the "grab" cursor for the card now that we are dragging?
- [ ] This is going to be a big ask...but could the card drag preview be a node instead? so like, picking up a card shows the node as it will be when dropped into the side? maybe vice-versa too, so picking up a node shows it as a card? could it....animate? 😬 
- [ ] On the subject of dragging, it would be fantastic if the droppable zones could give the user feedback about what happens if they drop the thing currently being dragged there, rather than just highlight. Same with when they are hovering. So specifically this applies to the large roster with the blurred cards: it would be great if when I pick up a node, it could switch to showing "Drop here to remove", and when hovering perhaps the node could have some sort of "deleting" state? Possibly too big of an ask, I know. 
- [ ] Nodes in the side panel are big chungus. Node sizing is controlled using the css variable `--base-node-size`, which by default is set to `calc(--var(--base-font-size) * 10))`. The value used in the existing side panels is scaled to this: `--base-node-size: calc(var(--base-font-size) * 6.6);` for a slightly smaller node. If you add that variable definition to the panel list, it should work.
- [ ] I get what you were thinking with dimming the cards after they have been added to the interview, but for me it breaks the 'tactility' of the interface, because the same entity is suddenly in two places. I think we are aiming for an interface metaphor based on newtonian rather than quantum physics. So, once added to the interview, cards should be removed from the list. Preferably triggering a layout animation.
- [ ] The sort buttons should mirror the design of the existing interface, whereby only the currently active sort is styled in color: primary. The rest are color: platinum.
- [ ] Add text shadow to "too many to display" to help with background contrast
- [ ] Add dimming with a linear gradient to the blurred background to help it not draw the eye
- [ ] Can we have a more pronounced stagger?
- [ ] Searching for "health" in the demo protocol on the large roster gives no results. I've used this query since day one to test this interface, because of how many clinics contain that word (to see examples, search for "Chicago". I've got no idea why this is happening, but its really important that we don't mess up the search feature (or make it behave differently), as this will impact people's in-progress research.
- [ ] Tabbing away from the app and back to it resets the search query. Clicking back into it also resets it. Neither should happen IMO.
- [ ] Perhaps the search should have a "clear button next to it, or that contextually appears? Maybe the "adornments" concept should include the ability to have more than one (or one on each side)? So something like `<SearchInput adornmentLeft={<MagnifyIcon />} adornmentRight={<ClearButton />} />`. This might also be a cleaner props interface because we dont need position.
- [ ] The Harling protocol I will send you on Slack crashes on load. Probably some invalid assumptions in your sortable properties hook, since this is valid according to schema 5. Stack trace is:
```
useSortableProperties.js:9 Uncaught TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at useSortableProperties.js:9
```
- [ ] Can we show an example of multiple prompts, and how we handle the transition?

package.json
```diff
@@ -135,6 +135,8 @@
     "react-transform-hmr": "^1.0.4",
     "react-transition-group": "^2.5.0",
     "react-virtualized": "~9.22.3",
```
jthrilly, line 137: Just wondering if react-virtualized is still used outside of this?

jthrilly: If not, we could remove

jthrilly:
I had a chance to check this out yesterday briefly. Here's what I found:

- Big one: not all the cards render in the HIV services example in the development protocol! I'm not sure why this could be, but if you look at the CSV file, there are 200+ items, and the latter ones aren't rendering.
- I had some issues with card responsiveness for large text sizes (might need to go to one column based on aspect ratio and scale factor?).
- Related to the above, long attribute labels need text breaking and hyphenation, and that column needs to have a fixed width relative to the card
- I think we should use the  "grab" cursor for cards, not the pointer (there is no click action
- Question: how is the performance vs old one? I notice on the Harling protocol that loading felt slower, but it could be subjective
- There does seem to be a consistent and very noticable performance issue as you cross into a droppable boundary. I think it has always been there, and likely can't be fixed, but it would be good to investigate and at least create an issue for it.
- The search experience feels a little strange at present. I sometimes see a flash of "no results available" when actually the state is "searching". On that point, is there a dedicated "searching" state?
- Did we ever add an error boundary in case the asset loading fails for some reason? I didn't see it checked off above and just wanted to check

wwqrd:
> - Big one: not all the cards render in the HIV services example in the development protocol! I'm not sure why this could be, but if you look at the CSV file, there are 200+ items, and the latter ones aren't rendering.

I think this is related to a bug where all elements had the key `undefined`, should be fixed now.

> - I had some issues with card responsiveness for large text sizes (might need to go to one column based on aspect ratio and scale factor?). I think there is space for a third column on 16:9 screens, too. In general we do want to look at responsive column numbers. I've used a resize listener hook to mimick the functionality we will get with the new `@container` queries with quite good results in scenarios such as this. This is used in the hover marquee component for example. Maybe it could help here, by letting us monitor the width of the list container, and set the column count based on it? Alternatively, we could use the old approach which was to look at the screen aspect ratio and calculate column count based on that, which is obviously much simpler. 

I've added in column calculations, the numbers might need tweaking. It's implemented by providing a function to the list which takes a width and returns a column count.

> - Related to the above, long attribute labels on the data card component need text breaking and hyphenation, and the first column needs to have a fixed width relative to the card to ensure that the label breaks rather than the date. Check the HIV services example in the development protocol to see this.

I've tried to improve this in the UI component, nothing seems ideal, what I'd like is a combination of soft wrap and word breaks..

> - I think we should use the  "grab" cursor for cards, not the pointer (there is no click action)

That should be fixed now, the data card component supports both click and drag, click style takes precedence.

> - Question: how is the performance vs old one? I notice on the Harling protocol that loading felt slower, but it could be subjective

It shouldn't be, not much has changed on that side of things, but I could explore this?

> - There does seem to be a consistent and very noticable performance issue as you cross into a droppable boundary. I think it has always been there, and likely can't be fixed, but it would be good to investigate and at least create an issue for it.

I'm going to see if there are any quick fixes we can do, or whether it'll be an outstanding issue.

> - Huge: The search is still not functioning at all correctly. Load the sample protocol and go to the universities state. Search for "Manch", which should return the university of manchester, manchester metropolitan etc. Searching for these specific universities directly yields no results! Searching for "Oxford" gives one result..."Wofford college". !!!

There were a couple of things going on here. The key bug mentioned above, and keys (i.e searchable fields) were being transposed to their variable ids, which doesn't match the data!

> - The search experience itself feels a little strange at present. I sometimes see a flash of "no results available" when actually the state is "searching".

Can you be more specific with an example, I've seen something similar myself, but unable to reproduce at the moment.

> - Did we ever add an error boundary in case the asset loading fails for some reason? I didn't see it checked off above and just wanted to check

Stages already have an error boundary, but there were some issues with the external data loading which I've resolved, the main one being inconsistent types (could throw an error, or return a rejecting promise). I've put an error screen up in the list view for when data cannot be loaded - as we've already spoken about, that's probably not the right way forward, and the existing error boundary was probably adequate.

rebeccamadsen:
I took a look! This is looking really amazing. I love the feel of it. As far as I can tell, everything already mentioned here has been addressed. I do have a couple of questions of my own.

* In previous small roster, the filter would search any visible field, not just name, but this seems like just name is searched for small roster. Is that correct?
* I'm not sure sort is working. This is from the classnames roster, and it's the same no matter which field I sort by:

image

Are protocols backward compatible or will protocols need updating?

jthrilly:
Just want a flag an additional thing based on the screenshot above - there should be a small amount of spacing on the right to separate the scrollbar from the content.

package.json
```diff
@@ -148,6 +149,7 @@
     "request": "^2.88.0",
     "request-promise-native": "^1.0.5",
     "reselect": "^3.0.0",
+    "resize-observer-polyfill": "~1.5.1",
```
jthrilly, line 152: Can you say more about what this is for?

src/styles/containers/_hyper-list.scss
```diff
@@ -0,0 +1,32 @@
+.hyper-list {
+  @include gpu;
```
jthrilly, line 2: I don't think this mixin is a good idea, generally, anymore.

src/styles/components/_list.scss
```diff
@@ -0,0 +1,32 @@
...
+  align-content: flex-start;
+  min-height: 100%;
+  width: 100%;
+
+  &--hover,
+  &--drag {
+    background-color: var(--node-list-action-bg);
+  }
+
+  // &--transition {
```
jthrilly, line 18: Remove

src/styles/components/_list.scss
```diff
@@ -0,0 +1,32 @@
+$node-inset: 1rem;  // approximate scaled margin + 'outer-trim' of a node
```
jthrilly, line 1: Can we replace this with a CSS variable to reduce our dependence on SASS?

wwqrd: Will remove entirely, it isn't used here and is just borrowed from node-list.scss

jthrilly (approved):

PR #1141, opened 2021-06-14
Feature/prompts refactor

### LINKED ISSUE #1080: New <Prompts> component
Update: see second comment for complete scope of issue.

The prompts on the sociogram interface do not animate when changing. They should match the behviour of the name generator prompts.

Suggest sharing a single prompt component and updating to user framer-motion as part of this.

jthrilly:
Fixes #1080. Companion to reimplementation of promps components in UI: https://github.com/complexdatacollective/Network-Canvas-UI/pull/176

Also includes:

- A wrapper around the UI prompt component allowing the sociogram prompts to be minimized and dragged
- A wrapper around UI prompt component that a development setting to enable text-to-speech 
- A development setting that turns text to speech on
- Standardization of CSS surrounding interface prompt containers
- Removal of existing Interviewer components


rebeccamadsen:
On Android, I get this error whether the speech feature is enabled or not:

image

From slack:
It looks like the API isn’t supported in the android web view (but IS supported in chrome for android, which is using the same engine these days): https://bugs.chromium.org/p/chromium/issues/detail?id=487255
So this is going to need two changes:
1. The feature itself needs to catch this error. Probably as simple as changing the useSpeech hook to have the speech synthesis API calls in a try/catch block, and then using the catch to set the error state in the hook.
2. (Optional) The feature should be disabled in Android in the settings

rebeccamadsen:
Also I noticed that when I started Android for the first time with this branch, the speech feature was enabled. But when I reset app data, it is disabled.

jthrilly:
> Also I noticed that when I started Android for the first time with this branch, the speech feature was enabled. But when I reset app data, it is disabled.

We tracked this down to `undefined` being considered truthy by the Toggle component. The `enableExperimentalTTS` property is nested within the `deviceSettings` store, which is correctly set to initialize as false. However because the property is bested, rehydration of existing user settings will make the value undefined (fresh app install will behave correctly).

We are considering:

1. If the `<Toggle />` semantics should be changed to "false unless explicitly true". Might this impact other things?
2. If we can resolve the initial value issue by controlling the re-hydration depth somehow, using configuration for the middleware.

rebeccamadsen:
With option 1, we could change the Toggle to be false by default, though some future setting may need to be true and Toggle would then swap it to false if undefined -- the reverse issue we have with enableExperimentalTTS.

Option 2 makes me a little nervous with unintended consequences. I'd feel better if it were just deviceSettings we modified.

jthrilly:
I've hopefully addressed existing feedback, so as long as there are no huge issues I think this is good to merge. The speech feature is fine to have some edge cases and roughness, since its just an experiment for now. The prompts themselves should be solid though.

rebeccamadsen:
The "backward" motion on the prompts seems like it's always false, so prompts only swipe one direction regardless of which direction the user clicks on arrows.

Understand about the speech. It generally works on electron, though you have to change stages if you change the setting while currently on a prompt. Android is disabled and doesn't cause issues.

rebeccamadsen (approved):
I think this is ready with the latest in ui. Does this need to update the ui version here, or will it be updated through other work soon enough?

PR #1146, opened 2021-06-25
Export sessions step-by-step

### LINKED ISSUE #1144: Revised session export journey
Session export should be reworked into a full-screen wizard-style flow.

#### Key aims

- Make the export settings more accessible, by including them in the process
- Improve the session selection experience, and make it less cramped.

## Spec

Create a session export flow which includes the following steps:

1. Open export wizard (replace session selector home section with export/manage buttons which open session selector screen)
1. Select sessions sessions, then:
a. Click export to file
b. Click export to server (export to server behaviour will not be changed, this process will stop here)
c. Click delete sessions (same behaviour as previously)
1. (a. cont.) Choose export preferences (formally in: Settings > Data Export Settings)
1. Choose file name

Export preferences should be persisted between exports, e.g. it should default to the previously used settings.

wwqrd:
Replace select section on Start screen with buttons that launch a step-by-step
export process.

Reuses the same components for setting options and selecting sessions, but wraps them
up in a multi-stage process.

Closes #1144

## Todo
- [x] Add more guidance text
- [x] "Exporting..." message (and delay?) on "exporting" screen


jthrilly:
We talked about this today and agreed that I would take a look at the spacing and some of the colors. Feature is ready to review in terms of functionality @rebeccamadsen.

rebeccamadsen:
Functionally this is working well for me. I think some wording could be updated. It is "Manage and export" on the start screen, but the buttons are in the opposite order -- export on the left and manage on the right. When you click into those, it'd be nice for the heading to indicate whether you are exporting or managing instead of just saying the same thing. This time the header is "Export and Manage", though. :)

The wording on the export and manage screen suggests the you can delete or export, and that you can change settings in the settings menu "at the top of this screen", but it's actually the top of the start screen or the next screen in the export dialog:

image

I suggest this blurb only refer to whichever action is applicable and that the settings can be changed on the next screen instead of pointing the user out of the step-by-step export to change settings.

Will the settings in the menu stay put? It's fine, just redundant.

Perhaps on the "Manage" screen, a toggle to select "exported" sessions could be helpful rather than "unexported"?? This is iffy, and I'd love to see use cases. Do we have users trying to remove unexported sessions to clean up so they can upload a new version of the session? Users trying to remove exported sessions to clean up what's been exported already? Both?

We do still have the "success" toast if you cancel an export without finishing (in windows, canceling at the system "save as" dialog), but I'm fairly certain this happened previously. Do we also need a way for users to cancel an export when it's in the middle of exporting, or they just need to close the app or wait for it to finish if they change their mind?

jthrilly:
@rebeccamadsen - I reworked this feature a bit, so a lot of your review might not apply.

Notable changes:

- I created a shared `<Picker>` component that is reused in the export overlay, but is also used in the other start screen interactions that "choose" a protocol or session.
  - Internally it uses `<Overlay>` and `<NewFilterableListWrapper>`.
- I updated `<Overlay>` to use the full screens form preference to set a full screen prop. I also added a toggle on the left to manually override this for a specific instance.
- I integrated the export feedback into the full screen export sprite.

Could you give this another review?

rebeccamadsen:
I tested on electron -- will try out android tomorrow. I like the combined delete/export, and the addition of the "select exported" option.

The override to expand the overlay doesn't respond if "fullscreen forms" is selected in settings. Should the caret be dimmed out, or is the override supposed to function in that scenario?

Cancel works pretty well. I did find that you could click cancel, and then the file choose would pop up before I could confirm cancel. A little strange, but I could still cancel at that point with the file chooser.

We do still have the "success" toast if you cancel an export without finishing (in windows, canceling at the system "save as" dialog), but I'm fairly certain this happened previously. However, I did notice this time that sessions are marked as exported even if the filename chooser is cancelled.

rebeccamadsen:
Cordova works pretty well! Exporting via share on android doesn't mark a file as exported, though it works. The file shared is missing the ".zip" on the filename, so I needed to rename the file on windows to make it usable there, but the contents were all there correctly after adding the ".zip" manually.

When full screen forms is enabled in settings, name generator forms are interesting. There are two "finished" buttons:
image

Toggling the caret doesn't switch to the regular form, which I think is fine, but the full screen mode could use some finessing when made smaller, and doesn't need both finish buttons.

Cancelling is working well! It's nice to not interrupt the cancel operation with a save as dialog, and sessions aren't marked exported falsely.

rebeccamadsen (approved):
Looking great! Zips on android are functioning, and I like the updates to the full screen forms.

rebeccamadsen:
Oh, just noticed one more thing I thought I should mention. The overlay looks great when in normal view:

many

But when I filter the export/delete session overlay, the internal width changes:

one

The other overlays don't change width like this, so thought I'd mention.

jthrilly:
> But when I filter the export/delete session overlay, the internal width changes:

How annoying. This is a flexbox issue. I'll fix it now and then merge. We can catch anything else in the testing for the release branch.

PR #1150, opened 2021-07-27
Fix Searchable List sort controls

### LINKED ISSUE #1148: sort not working correctly
With the roster updates merged in, I noticed the "sort" feature doesn't work on small rosters as it used to. No matter which sort button is selected, it seems to sort using the default sort option instead. This should be sorted by language, but is not. It worked before the updates to roster.

image

It's not noticeable in the development protocol because the names are already in order there, but alphabetical (name) and numerical (age).

wwqrd:
Very small fix which should resolve sort controls not working in SearchableList.

Also added a test to back it up.

Resolves #1148

### DIFF
--- a/src/hooks/__test__/useSort.test.js
+++ b/src/hooks/__test__/useSort.test.js
@@ -1,10 +1,11 @@
+/* eslint-disable @codaco/spellcheck/spell-checker */
 /* eslint-env jest */
 
 import React from 'react';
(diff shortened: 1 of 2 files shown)


jthrilly:
@rebeccamadsen could you also fix the small linting issue when you review this please?

rebeccamadsen (approved):
Perfect!

PR #1151, opened 2021-08-03
Automatic node layout on the Sociogram

wwqrd:
The aim of the feature is to have a toggle which when activated would enable a ‘spring embedder’ physics based layout of the nodes on the sociogram.

This implementation uses this engine: https://github.com/d3/d3-force. It is primarily intended for d3, but is written in such a way that it can be used without it. `cola.js` was also considered but the documentation wasn't up to scratch.

This attempt to build the feature hit a dead-end mainly due to "zoom" controls being a requirement of this feature.

## Requirements
- Force direction should be togglable.
- Nodes should still be draggable, and the layout will run continuously whilst the layout is active.

## Implementation

- Layout engine runs using a web worker.
- `useForceSimulation()` hook provides interface to worker/d3-force, current network state is managed as a React ref to prevent render thrashing. Consumer can read values as needed.
- `useAutoLayout()` hook translates between d3-force format (uses x/y co-ordinates to represent positions) and the Interviewer format (uses 0-1 decimal x/y to represent positions). This data is also managed as a ref.
- Rendering uses requestAnimationFrame() to draw data from `useAutoLayout()`.

## Post mortem

- Found it more perfomant for the worker to send simulation all frames to the client, rather than have the client request them as needed. Any extra frames can simply be ignored at the client end.
- When dragging nodes with auto layout active can experience "fighting the layout". I didn't get to the bottom of why this happened, perhaps an extra rendering loop between sending and calculating the change?
- Not convinced useAutoLayout hook should exist, or at least be in charge of translating positions - instead this could maybe simply be a function: `const transformedNodes = blendLayouts(nodes = [], forceSimulationPositions = [], toggle = true/false);`
- `useAutoLayout()` uses scaling to convert between co-ordinate types. This can change depending on the bounds of the d3-force positions. When dragging a node the bounds may change either smaller or larger, changing the zoom and giving an uncomfortable user experience. My recommendation here would be to use fixed zoom, with user controls.

## Things to solve

- [ ] What is the best way to maintain parity between absolute positions used by d3-force, and relative positions used by Sociogram. One suggestion might be to use static zoom values which are controlled by the user.
- [ ] How to render efficiently without "fighting the layout"? Should position rendering be managed by react or should the auto layout directly manipulate dom elements? Does this need a rethink of the drag and drop libraries?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -64,6 +64,7 @@
         "cross-env": "^5.2.0",
         "css-loader": "^3.4.2",
         "csvtojson": "^2.0.8",
+        "d3-force": "~3.0.0",
         "detect-port": "1.1.0",
         "dotenv": "4.0.0",
         "electron": "9.4.4",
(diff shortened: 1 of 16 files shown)


wwqrd:
Superceded by: https://github.com/complexdatacollective/Interviewer/pull/1156

PR #1152, opened 2021-09-01
cast types for imported csv data

rebeccamadsen:
This adds handling to cast types for imported csv data, and adds a test.

The basic issue is that csv data was all generally being imported as strings, so: ordinal, boolean, categorical, and number values were not being interpreted to match expectations from the codebook. Toggles and bins were not activating as expected, for example.

This solution focuses on first attempting to look up in the codebook the expected data type, and casting that type (e.g. converting numerical data to number); it falls back to string if this fails. If the variable is not in codebook, we attempt to derive the value; if _all_ the column data can be converted to boolean or number, then it is, and otherwise we fallback to string again. I considered attempting to regex for ordinal or categorical data, but reasoned that this may be prone to edge cases and has no meaning in the issue use case. I'm open to reconsidering this.

I debated a few entry points for converting the data, and landed where I did because we still know the data is csv, and have access to the codebook. My first couple of attempts were a little messier because of those two issues.

As for the format for the data...good csv formatted data is expected, of course, but architect will have issues as well if the csv formatting is incorrect. Categorical and layout data need to be in double quotes. Layout needs x and y in double quotes (escaped with double quotes) with values in 0-1 range without quotes (e.g. "{""x"": 0.5,""y"": 0.3}"). Categorical data, the strings could have quotes, unless you want numbers to be interpreted (e.g. "[""two"",0]"). Ditto with ordinal -- quotes to interpret as a string (e.g. "1" vs 1). Date needs to be YYYY-MM-DD or the date time string, double quotes are optional. Numbers are converted to number; if the variable is in the codebook, double quotes are optional, but if it is not in the codebook, the data must not have quotes. Booleans can be either lower or upper case, with double quotes optional. Scalar values range from 0 to 1 without quotes.


jthrilly:
Thanks for this @rebeccamadsen!

One concern I have is that this approach takes us further away from our existing CSV export being usable as roster data, which is somewhere we do need to get to.

In particular, the format the data needs to be in for your approach to work seems to be pretty fundamentally different from what we are currently outputting.

Is it possible to modify this feature to first assume our CSV export format, and then use the codebook to reconstruct the type _followed by_ using the methods you have implemented above?

rebeccamadsen:
Ah, good point! I'll look at adding in an assumption of our formatting.

rebeccamadsen:
Updated to include processing for our csv export data as an import. The main point of dissent was in processing categorical and layout variables, since we handle those in a special way on export (separating x and y, and each option). This makes lookup for codebook uuid a challenge, as well as needing to combine that into one variable again.

Several edge cases: both categorical variables and their options can have underscores in them; options can be numerical or string; categorical variables can have several values as true.

In importing data exported from a previous session to a new protocol, and then bringing those nodes into a new protocol, our custom data will be included: e.g. networkCanvasUUID, networkCanvasEgoUUID, etc.

jthrilly (approved):
Great work on a very complicated feature. There are a few style things I would want to look at in normal times, but I'm happy to merge this now 😄

PR #1158, opened 2021-12-23
make dyad edge check not directional

### LINKED ISSUE #1157: dyad sociogram discrepancy
Creating a link on sociogram only shows up on the dyad census if the link was created in the same direction as the dyad census asks. For example, if a user creates a link between Joseph and Henry on the sociogram, but the Dyad asks about Henry and Joseph, then the Dyad will show a non-answer.

Related, if you have that link existing in sociogram but mark "no" in the Dyad Census, you would expect the link to disappear in the Sociogram, but it does not. Directionality appears to matter. Is this intentional?

Possibly a minor issue if users don't typically use both Dyad and Sociogram.

rebeccamadsen:
Fixes #1157. Also adds a test.

To verify, create two nodes. Check which direction the dyad census offers (`a` and `b`, or `b` and `a`). Create an edge on sociogram in the opposite direction. Return to dyad census, and the edge should show as "Yes".

Dyad and tie-strength could potentially benefit from a refactor as there are multiple overlapping functions, but that would be another issue.


src/containers/Interfaces/DyadCensus/useEdgeState.js
```diff
@@ -2,13 +2,13 @@ import { useState, useEffect } from 'react';
...
 
-const getEdgeInNetwork = (edges, pair, edgeType) => {
+export const getEdgeInNetwork = (edges, pair, edgeType) => {
   if (!pair) { return null; }
   const [a, b] = pair;
 
   const edge = edges.find(({ from, to, type }) => (
     type === edgeType
-    && ((from === a && to === b) || (to === b && from === a))
+    && ((from === a && to === b) || (to === a && from === b))
```
jthrilly, line 11: Ha! So it turns out to have been an implementation bug all along, and was intended to work the way it does not. Great find.

jthrilly (approved):
Looks great! Thank you! 😊

PR #1162, opened 2022-01-26
reinstate old rosters and rename new

rebeccamadsen:
This restores `NameGeneratorAutoComplete` and `NameGeneratorList` and moves the new roster interface to `NameGeneratorRoster`. It also adds a fix in `SearchableList` for handling empty `sortOptions` (or partially empty, such as an empty object or an object with empty arrays). `searchOptions` should already work fine with empty params.

To do:
- [x] fix error with card data populating for name gen auto complete
- [x] test with new roster and old ones
- [x] verify fix for empty `sortOptions` and `searchOptions` parameters
- [x] restore old tests
- [x] verify formatting
- [x] image for new roster


rebeccamadsen:
Updated development-protocol [here](https://github.com/complexdatacollective/development-protocol/pull/20) since schema 6 will need the newer stage name.

PR #1166, opened 2022-02-04
Release for 6.2.0

jthrilly:
## Network Canvas Interviewer 6.2.0

This is a significant feature release, containing numerous changes and bugfixes. It introduces schema version 6, but remains compatible with schema versions 4 and 5. It should be installed by all users currently running version 6.1.0 or above. 

We have prepared a video demonstrating the new features in this software: [here]().

If you are running a version prior to 6.1.0, please consult the release notes for that release before updating: [https://github.com/complexdatacollective/Interviewer/releases/tag/v6.1.0](https://github.com/complexdatacollective/Interviewer/releases/tag/v6.1.0)

### Changelog:

- Include a new interface: The "name generator roster interface" is designed to supersede the previous roster interfaces and simply the experience of using roster data.
- Automatic layout mode on the sociogram, which uses a force-directed physics simulation to automatically position nodes in real time.
- Improved support for using CSV data in rosters. Interviewer will now attempt to correctly match the variable type to any known codebook variables, and can successfully import previous interviews as roster data.
- A new prompt component that handles overflowing text more gracefully, and includes experimental text-to-speech support, allowing prompt text to be read aloud.
- Automatic desaturation of categorical values where the option value is less than zero, or the string value is 'other'. This allows for styling missing or other options with less visual emphasis.
- A new participant 'nudge' on the ego form interface, which reminds the participant to scroll down to answer further questions.
- A new redesigned data management screen for sessions, allowing access to export options and more space to view and select sessions to delete or export.
- Numerous other small bugfixes and improvements.


rebeccamadsen:
We need new updated images for Roster again. Also, I keep getting "Nothing matched your search term" on the roster. Am I missing something?

jthrilly:
I'm not sure what the issue could be with the seach - it works great on the sample protocol v3, but you're right that it isn't working at all on the development protocol. I bet it has to do with default parameters.

jthrilly:
Okay, I believe I have fixed the issue. I think the debounced state stuff steve did (which I tore out) was doing something else behind the scenes. It seemed to only impact the development protocol, but I reverted it and fixed the issue I initially solved a different way, so should be good now.

PR #1176, opened 2022-03-10
Feature/min max namegenerators

jthrilly:
This work is the product of a paid collaboration with Dr Jereed Ivanich at the University of Colorado.

This PR implements the ability for name generator interfaces to specify the minimum and maximum alters to be named.

- Configuration is via the behaviours prop on the stage, with `minNodes`, and `maxNodes`.
- These properties will land in schema version 7, along with changes in #1175 
- Functionality is implemented on `NameGenerator`, `NameGeneratorQuickAdd`, and `NameGeneratorRoster` interfaces.
- Alter count boundaries apply at the **stage level** (not the prompt level). If separate boundaries are required, multiple stages can be created.
- Where alter count is less than `minNodes`, session navigation is disabled in both directions, and a message informs the user of the number of alters that must be created to continue. 
- Where it is equal to or greater than `maxNodes`, a message is shown advising the user to continue to the next stage, and the ability to add new alters is disabled.
- For non-roster interfaces, this required fairly complex logic to handle the side panel functionality.
  - When `maxNodes` is reached, dragging is disabled if the node is from an external data source OR if a node did not originate on this stage.
  - Nodes that were created on this stage can still be dragged to or from panels, since this will not increase the node stage count
- For the roster interface, when `maxNodes` is reached, the search panel will close if open, and any future searches will produce cards that cannot be dragged. Existing cards may still be dragged to remove them.


rebeccamadsen:
This works pretty great! Seems super useful, too.

I did notice you cannot go backward from a stage with this behavior, not even via the stage menu. Unless it is a stage with multiple prompts, and then you can go backward but only via the stage menu. It's a little nuanced, but it's possible to feel stuck. But perhaps this is intentional? For example, on the development protocol, stage 9 (according to the menu's numbering) doesn't let me leave no matter what, but stage 4 will let me leave via the stage menu but only on prompt 2.

Also, I made a protocol with multiple prompts for the roster, and it will not let me past the first prompt without satisfying the minimum requirements. It is meant to be a stage requirement, though, not a prompt, right?

jthrilly:
> Also, I made a protocol with multiple prompts for the roster, and it will not let me past the first prompt without satisfying the minimum requirements. It is meant to be a stage requirement, though, not a prompt, right?

This part sounds like a bug!

> I did notice you cannot go backward from a stage with this behavior, not even via the stage menu.

Yeah, this is tricky. I want to let the stages menu work again but I didn't see how that had been implemented before. I'll look at it again.

jthrilly:
[Min Max Alter test.zip](https://github.com/complexdatacollective/Interviewer/files/8284037/Min.Max.Alter.test.zip)

rebeccamadsen (approved):
Looks great!

PR #1193, opened 2022-08-11
Fix slow alter form rendering

jthrilly:
The per alter form (and other form interfaces using the swiper component)
were rendering extremely slowly. Cause was that the Swiper was rendering
all forms at once, which combined with the terrible performance of
redux-forms to create a mess.

Solution here is to remove the Swiper component in favour of a simple
framer-motion animated transition.


jthrilly:
Additional fix: prevent re-rendering of markdown throughout the app. Significant general performance improvement! :boom:

PR #1211, opened 2022-10-19
Attempt at new sortOrder

jthrilly:
This is a first pass at a new sortOrder that includes functions that are needed by the two mode sociogram feature. These are:

- Ability to sort by node type
- Ability to run across multiple node types, and handle the possibility that some nodes will not have a value for a given attribute

Node type sorting was implemented using an additional property for each sort rule (hierachy) that should only be present on rules targetting the `_nodeType` attribute. This still needs to be added to the protocol schema if we decide it is the way to go.

The API of the function was changed somewhat to simplify its use. This makes it not directly compatible with existing code, but only minor updates should be needed.

Work has stalled because of difficulties handling undefined and null values.

The previous version of this was built on lodash.orderBy, which puts undefined and null values at the start of the list. This is not what we want.

Fixing this is very difficult because:

1. orderBy does not support a custom comparator function in any released version. This functionality was added four years ago, but has not seen the light of day.
2. Without a custom comparator function, we cannot use the attribute type to sort null and undefined values to the end of the list (for example by returning -Infinity/Infinity for a number, and true/false) for a boolean.


jthrilly:
Second attempt at this:

- Replaced `orderBy` with vanilla javascript implementation, using `Array.sort`.
- Implemented the first of possibly many vanilla JS alternatives to other lodash methods, in the form of a `get` implementation using nullish coalescing and optional chaining. This required a new babel plugin because of node12, which can be removed in the future.
- Renamed the sort function `createSorter` since this seemed to be more appropriate.
- New system generates comparitor functions compatible with `Array.sort` based on the existing rule object format. This allows for the desired behaviour with null and undefined values (placing them at the end of sorted lists). 
- An additional 'type` property was added, to provide optional special sorting behaviour. Aside from basic variable types (string, integer, etc.) this also supports a 'hierachy' type, which uses a separately provided array property (`rule.hierachy`) to determine sort order.
- This new property is now used to implement the node type sorting, without the need for a 'special' rule property. Node type sorting can be implemented with `rule.property = 'type'` and `rule.hierarchy = [ 'person', 'place']`.
- The new sorter is not a drop-in replacement for the existing one, so further work is required in order to integrate it.

PR #1212, opened 2022-10-20
Feature/two mode sociogram 2

jthrilly:
I decided to restart this work, abandoning some of the more ambitious
elements of the previous attempt.

This version will continue to use the split Canvas/RealtimeCanvas
components, and will defer merging them to a separate task.


jthrilly:
## Update on this feature

After some consideration, the API changes required to make this work correctly are beyond the scope of what we can achieve in the timeframe we have. We are therefore deferring completion of this feature to a later date, where more significant protocol schema revision can be undertaken.

Specifically:
- Implementation of multiple stage "subjects" revealed the fact that the fundamental "subject" concept is not a good general purpose fit for stages of different types. The name generator creates its subject, but the sociogram merely displays it. Subjects are also defined by both a "type" and a key, which is redundant since no interface uses both node and edge subjects. Future protocol schema versions should replace stage subject with specific verbs that describe what the entity is used for, and simply store the entity key in an array. For example, the name generator could have a `creates` key, and the sociogram could have `displayNodes` and `displayEdges` keys.
- The sociogram is unusual in that much of the configuration is nested within the prompt. This presents significant issues when editing in Architect, since changing the stage "subject" can and will invalidate prompt configuration. This leads to the entire stage needing to be configured each time a one of two node types is changed. There is no way in Architect to mark a complex nested configuration, such as a prompt, as being invalid. This means that we cannot simply invalidate portions of a prompt, and notify the user that their attention is required. Fundamentally, there is a large amount of tension between prompt and stage level configuration that should be resolved. Prompts should configure the interaction behaviour (choosing between edge creation and attribute highlighting), but background, layout, and edge display should be configured at the stage level.
- The fact that variables are scoped to entities in the codebook means that features such as node highlighting and layout require a variable to be specified for each node type. A future revision of the protocol schema could consider making variables independent of entities to resolve this, and the issue above.

We decided to merge this branch "as-is" despite the fact the feature is unfinished because:
- It is backwards compatible with schema 7, and so should not cause any issues. Changes to subject selectors to support subject being a collection rather than an object are conditional. Conditional code paths also exist for storing layout data, based on multiple stage subjects being present, and these should not impact normal sociogram functioning.
- It implements significant fixes, in particular regarding `sortOrder`, which are beneficial in a broader sense. 
- There are other quality of life improvements contained that do not need to be lost, such as new animations for the node bucket.

PR #1225, opened 2023-04-11
Fix ordinal/categorical bin sorting

jthrilly:
This did indeed turn out to be caused by the refactor of the sort functionality. To recap, that refactor addressed bugs related to incorrect sorting results when sorting fields without taking into account their type (number, string, date, etc.).

The `<MultiNodeBucket>` component (among others) passed the protocol sort configuration directly to `createSorter`. This PR updates the `withPrompt` helper so that it processes the sort configuration by adding a `type` property to each object based on the sort attribute's codebook type. See: https://github.com/complexdatacollective/Interviewer/pull/1225/commits/a197e499926650c97d6e1f3e6dc747deb012321d

I then realized there were other instances of incorrect use, including the `<Picker>` component which is used to select a protocol or interview session from the start screen. This PR also updates this component, along with the `<NewFilterableListWrapper>` component to also use the new sort functionality correctly. See https://github.com/complexdatacollective/Interviewer/pull/1225/commits/df49da095e8a9502b3af75fac919db402feda446

Other thoughts:
- I am concerned that there is so much duplicated code covering filtering lists! We have 4 or 5 different implementations of this. A unified hyperlist is badly needed. We must make sure to add this in the next version.
- The `<NodeList>` component seems to also process sort rules, but I have not verified if it is functioning correctly as I can't see a protocol property that would configure it.


buckhalt (approved):
Tested and confirmed this fixes the issue.

- Sort order on categorical and ordinal bin node buckets is configurable and working.
- Verified that the `<Picker>` component is sorting correctly by checking that sort order of interviews and protocols on start screen is working.
- Verified that the `<NodeList>` component is correctly processing sort rules. `<CategoricalItem>` renders a `<NodeList> `when expanded that has configurable sort rules.

PR #1227, opened 2023-04-12
Fix/autocomplete roster

jthrilly:
The name gen autocomplete interface crashed on load because the stage prop was not being passed to node list.


buckhalt (approved):
Confirmed this fixes the issue

PR #1230, opened 2023-04-14
Release branch for 6.4.1

buckhalt:


buckhalt:
This release contains a fix for a bug that caused issues with sort order in categorical/ordinal bin buckets, as well as other components which process sort rules. This bug was caused by the refactor of the sort functionality. 

Other features:
- fixed a bug that caused the autocomplete interface to crash on load because the stage prop was not being passed to node list.

jthrilly (approved):

PR #1238, opened 2023-06-23
Release/6.4.2

jthrilly:


buckhalt:
This PR contains a bugfix within networkQuery:

- Skip logic was not being evaluated properly when a value was an array with a single item (['F']), and the variableValue was that exact single item ('F'). This was returning false as these values are not the same.

- This bug occurred in instances of skip logic configured with 'exactly' rules on categorical variables.

- This now returns true. Values which are arrays containing a single item that exactly matches the variableValue are considered equal.

PR #1241, opened 2023-08-10
Feature/interaction sounds

jthrilly:
This PR implements the interaction sounds as developed for our summer internship by Anika Wilsnack.

This will be an opt-in feature.


src/utils/playSound.js
```diff
@@ -0,0 +1,36 @@
+/* eslint-disable import/prefer-default-export */
+import { debounce } from './lodash-replacements';
+
+// Play a given sound
+export const playSound = ({
+  src,
+  loop = false,
+  volume = 1,
+  debounceInterval = 1000,
```
buckhalt, line 9: The delay between the trigger and the sound feels too long, and sometimes doesn't fire at all, for quicker consecutive sound interactions (e.g., adding nodes from form). Do you think this could be related to this debounce mechanism? If so, maybe there is a way to edit this to still prevent rapid fire while also making the interactions feel responsive.

buckhalt (changes requested):
Overall, this looks fantastic. The interaction sounds add a really nice touch to the user experience.

As described below, there's an issue with quick, consecutive interactions. While this could be related to the debounce mechanism, I think another culprit is the sound file lengths. I noticed that the 'perceived' end of the sound happens much sooner than the true end of the file for most sounds. This may be contributing to delays because the interaction feels completed before the sound is actually done playing. I think the interactions would feel much smoother and responsive if the sound files were trimmed down to remove any excess space past when the sound itself finishes playing.

src/utils/playSound.js
buckhalt: trimming trailing silence in [0340356](https://github.com/complexdatacollective/Interviewer/pull/1241/commits/03403566c125ccb6cc22be2ec551bfbf21e46b98) seems to improve this

buckhalt (approved):

jthrilly:
Agreed to implement true polyphony and multiple instances, and revert the sound trimming.

PR #1242, opened 2023-08-11
Feature/refactor namegenerators

jthrilly:
This PR refactors the name generator interfaces (form and quick add) into a single unified interface. No more duplicated code between them! Just the presence of quickAdd or form properties will determine which node creation method is enabled.

I rewrote the quick add form as part of this, to have a few nicer usability elements (tooltip text, better animation, etc.)

I also rewrote the min/max nodes stage level validation function, as the old one was buggy.

All components were re-written as functional (they used to be class components). 

This functionality is pretty key, so we need to check for bugs. Possible areas include drag/drop, and prompt level additional attributes.


src/containers/Form.js
```diff
@@ -138,6 +138,8 @@ Form.propTypes = {
   submitButton: PropTypes.object,
   initialValues: PropTypes.object,
   validationMeta: PropTypes.object,
+  otherNetworkEntities: PropTypes.object,
+  subject: PropTypes.object.isRequired, // Implicit dependency used by autoInitialisedForm
```
buckhalt, line 142: I'm getting a console warning related to `subject`:

```
Warning: Failed prop type: The prop `subject` is marked as required in `Form`, but its value is `undefined`.
```

src/containers/QuickNodeForm.js
```diff
@@ -1,157 +1,200 @@
...
   };
 
-  handleChange = (e) => {
-    this.setState({
-      nodeLabel: e.target.value,
-      // TODO: is this used?
-      cancelBlur: false, // eslint-disable-line
-    });
-  }
+  // Handle showing/hiding the tooltip based on the nodeLabel
```
buckhalt, line 85: Tooltip text looks great

src/containers/Interfaces/NameGenerator.js
buckhalt, line 1: Prompt level additional attributes aren't being assigned on Quick Add Name Generators. Looks ok for Form Name Generators.

buckhalt (changes requested):
This is looking really great! I especially like the quick add animations and tooltip text.

Besides the couple issues below, things seem to be working correctly. I verified that min/max and drag/drop are working correctly.

src/containers/Interfaces/NameGenerator.js
buckhalt: fixed with [853bc3a](https://github.com/complexdatacollective/Interviewer/pull/1242/commits/853bc3a096e14b909ddd141a36acf9513b367d97)

src/containers/Form.js
buckhalt: fixed with [3da647e](https://github.com/complexdatacollective/Interviewer/pull/1242/commits/3da647ec3fc0f84b08e77f8fe9801c4a5e64d288)

buckhalt (approved):

PR #1267, opened 2024-04-19
Release/6.5.3

buckhalt:
**Release branch for 6.5.3**

Changelog:
- https://github.com/complexdatacollective/Interviewer/pull/1266


jthrilly (approved):
We discussed a more complex version of this fix that didn't 'pollute' the concept of hierarchical sorting, but in the end decided that time was better spent elsewhere. 

If we return to this issue, we can think again about how multiple categorical values per entity can/should be sorted.

